PR #47, opened 2017-03-16
Update README.md

bpbond:


bpbond:
This is weird. I was able to push changes right up from the GitHub client to `master`...and then when I reset to HEAD@{1} and tried to force-push, it rejected me:

```
emote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Cannot force-push to a protected branch
```

And NOW I edit `README.md` online and it makes me create a branch, saying (correctly) that `master` is protected. Wtf? Why was I able to push the demo changes?

PR #69, opened 2017-03-20
Translated from original data system

bpbond:
Note found one mistake - see OLD_DATA_SYSTEM_BEHAVIOR block


skim301:
Good job Ben. Looks great. I have finished my review. You can commit.

bpbond:
Looks good to me.

PR #80, opened 2017-03-21
Translated DEMO_chunk_LA112.U

bpbond:
Note - found an issue with DEMO_OOPS uranium parameter; see
OLD_DATA_SYSTEM_BEHAVIOR block


bpbond:
Looks good Ben. You are a great coder!

Issue #26, opened 2017-01-26
Multiple datasets into same XML


bpbond:
Pralit notes that there are cases in the d.s. where multiple datasets flow into a single XML, and the ordering is important. We'll need to look at this.

pkyle:
Yeah, good point. I think the only case in aglu/energy modules is the bio_hi.xml file in AgLU, which changes yields and carbon contents of bioenergy crops, and those parameters are set in different R code files. But that file is not long for this world (isn't in the integration branch). There is at least one case of this in the emissions module (all_fgas_emissions.xml, created by L241 and L252), but my guess is this and any others could be modularized without much problem.
The issue I've had wrt this issue is that we can get XML files that are identical in the model (every parameter equal) but that will nevertheless appear quite different (comparing the text), if R code files aren't run in the same order.

pralitp:
This should be taken care of it the approach that I took on the branch I just created (I'll open a pull request in the next day or two and reference this issue).

Basically the current level 2 just publish the tables that the current data system would add to a "batch xml file" as output.  And a separate chunk will input them back in and do the XML conversion.  So each "batch xml file" now get's it's own chunk.

PR #81, opened 2017-03-21
Chunk generate xmlbatch

pralitp:
In the "chunk-generator" split out conversion from CSV to XML into their own chunks since it may be the case that they include data from multiple chunks (resolving #26 ). These conversion chunks are named according to the "batch" file name from the old data system that used to tie CSV to XML conversion together.

Note the conversion is set up using R pipeline syntax instead of generating a "batch" file:
```R
create_xml("test.xml") %>%
  add_xml_data(tibble1, "UnlimitRsrc") %>%
  add_xml_data(tibble2, "UnlimitRsrcPrice") %>%
  run_xml_conversion()
```

I've included the CSVToXML.jar in the R package and call the conversion passing the data through STDIN avoiding the need to write any temporary files.


tests/testthat/test_xml_conv_utils.R
```diff
@@ -0,0 +1,95 @@
...
+test_that("converting without adding data causes error", {
+  skip("TODO: need to find a way to get messages from Java")
+  conv_test <- create_xml("test.xml")
+
+  expect_message(run_xml_conversion(conv_test), regex="java.lang.NullPointerException", all=F)
+})
+
+test_that("converting bogus data causes error", {
+  skip("TODO: need to find a way to get messages from Java")
+  `%>%` <- magrittr::`%>%`
```
bpbond: Surprised you needed to do this (`%>% <- magrittr::%>%`). It was throwing an error without?

pralitp: Nope, I guess I was assuming the package wasn't going to be loaded.  Should I back all of those out?

bpbond: No, thanks, no worries. We can tweak later. Everything here looks good to me.

PR #84, opened 2017-03-22
Translated L112.U_DEMO

bpbond:
Note found incorrect behavior in step 3; see fix in code.


bpbond:
Looks good Ben. You are an awesome coder!!!!!

PR #93, opened 2017-03-23
Temporarily disable XML tests

bpbond:
See #81


bpbond:
Hey @pralitp so when I pulled the new master, with your PR #81 , it failed a bunch of tests because I don't have Java installed. Also, it skips a number of tests with printing messages. I absolutely should have checked this _before_ merging--sorry. 

In this commit, I've tweaked some things and disabled all the XML tests until we get a chance to talk further. Basically: could we/is it desirable to have the tests just silently skip if there's no Java available?

PR #107, opened 2017-03-24
Java optional

### LINKED ISSUE #103: Silently skip XML tests if Java not installed?
The XML writing tests all fail if there's no Java installed (and in addition, at least on my Mac, every test cycle now pops up a dialog box asking if I want to go install Java). Is this the behavior we want, i.e. to force folks to install Java? Or would we rather have the tests not execute, or at least not fail, if Java is not installed? Not sure.

pralitp:
Adds a global option gcamdata.use_java (default to FALSE) that controls if `run_xml_conversion` and any of the tests that rely on it are actually run so they can be safely skipped if a user does not have Java installed.

This resolves #103

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ warnings_are_errors: true
 #sudo: required
 cache: packages
 
+env:
+  - GCAMDATA_USE_JAVA=TRUE
+
 r_github_packages:
   - jimhester/covr
   - krlmlr/mockr
--- a/R/onload.R
+++ b/R/onload.R
@@ -1,0 +1,13 @@
+# onload.R
+
+.onLoad <- function(libname, pkgname) {
+  op <- options()
+  op.gcamdata <- list(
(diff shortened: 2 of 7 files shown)


pralitp:
@bpbond, I'm not sure why codeconv hates me.  In particular not sure why it thinks this changes anything about `dstrace`.  I added the commit b84da1d because I thought it was mad that we were no longer testing the XML conversion utilities but now I'm not so sure.  Maybe we can back that commit out.

bpbond:
:)

I think because the branch needs to be updated. Doing that now.

.travis.yml
```diff
@@ -5,6 +5,9 @@ warnings_are_errors: true
 #sudo: required
 cache: packages
 
+env:
+  - GCAMDATA_USE_JAVA=TRUE
```
bpbond: Nice

bpbond:
Thanks again for deal with this @pralitp !

Issue #112, opened 2017-03-27
LB110.For_FAO_R_Y


bpbond:

bpbond:
Hi @chenminbnu @abigailsnyder it looks like this file uses a `translate_to_full_table` function that's also called by LA101.ag_FAO_R_C_Y which @kvcalvin is working on. Between the three of you, please coordinate and make sure only one of you implements this function!

kvcalvin:
@chenminbnu and @abigailsnyder: I'm working on my chunk today and tomorrow. I can try translating this first and then maybe we can cherrypick it into your branch.

abigailsnyder:
@kvcalvin Great, I was not going to start until tomorrow or Wednesday

chenminbnu:
@kvcalvin Thanks Kate.

bpbond:
Hi guys, I just realized that this chunk's inputs are coming in wide-form, but they shouldn't be. Let me make a change to the upstream chunk, and then they should come in long form, saving you a transform step.

abigailsnyder:
@bpbond I'm having trouble running just the chunk for testing, namely doing
driver() -> x is fine but then
x[["module_aglu_LB110.For_FAO_R_Y"]] just returns NULL. 

Also, just to verify, if I want to run just the chunk, I don't have to do
build & reload
driver() -> x
every time?

bpbond:
Should type `x[["name_of_output_data"]]`, not name_of_chunk.

Re second question, right. You should be able to type `module_aglu_LB110.For_FAO_R_Y("MAKE", x)`

kvcalvin:
I don't think we need "translate_to_full_table" anymore. I think we can replicate this using tidyr with expand and possibly left_join. I'll do it for my data and then post a sample code on the wiki.

bpbond:
Great, thanks.

PR #123, opened 2017-03-28
Have LA100.FAO_downscale_ctry product long data

bpbond:
See #112

### DIFF
--- a/R/zchunk_LA100.FAO_downscale_ctry.R
+++ b/R/zchunk_LA100.FAO_downscale_ctry.R
@@ -231,10 +231,13 @@ module_aglu_LA100.FAO_downscale_ctry <- function(command, ...) {
              item.codes = `item codes`) ->
       FAO_data_ALL_5yr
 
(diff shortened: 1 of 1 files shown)


bpbond:
Hi @chenminbnu @abigailsnyder - this change to the upstream chunk from yours makes it produce long data, so you won't need to transform. Just for practice, do you want to check out my changes as approve (presumably) as reviewers?

bpbond:
@abigailsnyder you mean these lines?
```R
FAO_data_ALL_5yr$element <- gsub(pattern = "_[A-Z]*$", "", FAO_data_ALL_5yr$element)
FAO_data_ALL_5yr$element <- gsub(pattern = "^FAO_", "", FAO_data_ALL_5yr$element)
```

The first should strip out any underscore followed by capital letters at the end of `element`; the second remove any "FAO_" occurrence at the beginning.

abigailsnyder:
Awesome, I'm not up on my regular expressions. Thanks

PR #151, opened 2017-03-30
Don't enforce metadata for old d.s. files

bpbond:


bpbond:
Sorry @CLynchy @cahartin - this will fix it.

Issue #156, opened 2017-03-31
Metadata for LDS files incomplete Mueller_yield_levels.csv


bpbond:

pkyle:
Title: Attainable yield levels by country and basin
Units: tonnes per hectare ("count" is number)
Extended description: Min, average, and max refer to the respective yields observed in grid cells within country, basin, and crop; count is the number of grid cells with harvested area for the given crop, country, and basin.
Source: Mueller, N. D., Gerber, J. S., Johnston, M., Ray, D. K., Ramankutty, N. and Foley, J. A.: Closing yield gaps through nutrient and water management, Nature, 490(7419), 254–257, doi:10.1038/nature11420, 2012. Geospatial data processed by Yannick Le Page.

bpbond:
Thanks!

PR #157, opened 2017-03-31
Ben - LA100.0_LDS_preprocessing

bpbond:


bpbond:
@pkyle - I tagged you as a reviewer here because there's a bunch of things I need to fill in about this file but didn't know. What's "LDS" (I assume not Latter- Day Saints)? What's GLU? Also see #156 re incomplete metadata in one of these files.

R/zchunk_LA100.0_LDS_preprocessing.R
```diff
@@ -8,58 +8,118 @@
...
+        sub("ctry_iso", "iso", .) %>%
+        sub("reglr_iso", "GTAP_region", .) %>%
+        sub("glu_code", aglu.GLU, .) %>%
+        sub("land_type", "land_code", .) %>%
+        sub("SAGE_crop", "GTAP_crop", .) %>%
+        sub("mirca_crop", "MIRCA_crop", .) %>%
+        sub("use_sector", "GTAP_use", .) ->
+        names(LDSfiles[[nm]])
+
+      # Replace numerical GLU code with concatenated GLU and the
```
kvcalvin: What does "concatenated GLU" mean?

R/zchunk_LA100.0_LDS_preprocessing.R
```diff
@@ -8,58 +8,118 @@
...
+      # three-digit code (padded with zeroes as necessary)
+      if(aglu.GLU %in% names(LDSfiles[[nm]])){
+        LDSfiles[[nm]][[aglu.GLU]] <- paste(aglu.GLU, sprintf("%03d", LDSfiles[[nm]][[aglu.GLU]]), sep = aglu.GLU_NAME_DELIMITER)
+      }
+
+      # Re-set Taiwan to mainland China, as the current version of AgLU
+      # (and pre-2015 versions of FAOSTAT) doesn't disaggregate Taiwan
+      if("iso" %in% names(LDSfiles[[nm]])) {
+        d <- LDSfiles[[nm]]
+        if(nm != "Pot_veg_carbon_Mg_per_ha") {
```
kvcalvin: Why is this file different?

bpbond: It's operating on a group (12ish) of files and doing same operations to most, and a few specialized operations on a couple. It _is_ an odd duck. But kudos to @pkyle - pretty reasonable way to handle it, actually; I didn't change much of the logic.

R/zchunk_LA100.0_LDS_preprocessing.R
bpbond: I'll clarify that. Just means `23` becomes `GLU023`, for example.

R/zchunk_LA100.0_LDS_preprocessing.R
kvcalvin: I guess I was asking from a scientific standpoint. Do we want Taiwan in that file? Does it not exist to begin with?

bpbond: Oh, sorry. Paging Dr. Kyle, Dr. @pkyle , please report to PR #157.

pkyle: Sorry guys, lots of history here! We don't have Taiwan as an aglu region, because (a) Taiwan was excluded from SAGE/HYDE in the aggregation the Yuyu did, and (b) it was excluded from FAOSTAT when I queried the data four years ago. Alan has addressed (a); it's now included in the land cover data that's the input to the data system. And, FAOSTAT now includes Taiwan in their various databases. However, the FAOSTAT data currently in the data system was queried in about 2012, and Taiwan was added in 2014, so it's not actually in our level0 data files. As such, we're not yet in a position to add Taiwan as an aglu region. We would need to update our FAOSTAT queries, which will come along with more fun because they've certainly changed country names and quantities, if not the available data altogether).
The steps performed here basically re-map the ISO code of Taiwan back to China, and aggregate anything that needs aggregation (e.g., quantity variables like land cover totals, but not characteristic variables like vegetative carbon densities).

pkyle:
Not sure if this is the correct venue; LDS = Land Data System. (Alan was honestly not familiar with Latter Day Saints when he named it; I had assumed he was just being ironic so didn't say anything) The reference for the LDS is:
Di Vittorio, A., P. Kyle, and W. Collins. 2016. What are the effects of Agro-Ecological Zones and land use region boundaries on land resource projection using the Global Change Assessment Model? Environmental Modelling & Software 85, 246-265. http://dx.doi.org/10.1016/j.envsoft.2016.08.016.
GLU = Geographic Land Unit.

bpbond:
@kvcalvin I will merge if things look good to you.

PR #144, opened 2017-03-29
Chen-Snyder lb110.for fao r y

abigailsnyder:
LB110.For_FAO_R_Y first attempt pull request

PR #188, opened 2017-04-01
Restoring missing LB110 code - yikes

bpbond:


bpbond:
I don't know what happened, but this merge restores the code from #144

PR #197, opened 2017-04-03
Add pipeline helpers for standardizing iso codes and protecting numeric column names

rplzzz:
The iso fixes are something that comes up a lot in the early stages of ingesting data, so it's convenient to be able to fix those in a pipeline.

The `protect_integer_cols` and `unprotect_integer_cols` have to do with a [bug in `select_if`](http://stackoverflow.com/questions/41883698/select-if-behaves-erratically-with-non-canonical-column-names) and related functions.  It probably doesn't come up as often as the iso issue, but having a workaround for it makes things a lot clearer when it does arise.


R/pipeline-helpers.R
```diff
@@ -87,3 +85,84 @@ repeat_add_columns <- function(x, y) {
...
+#'
+#' If you're dealing with a split or a merger, then making this change will generally create
+#' duplicate rows, so you should follow it up with an appropriate aggregation.
+#'
+#' @param d The data to be transformed
+#' @param from The old ISO code
+#' @param to The replacement ISO code
+#' @param col The name of the columns with the ISO codes (default = 'iso')
+#' @return Tibble with the old iso code replaced by the new one.
+#' @export
```
bpbond: These are internal functions and probably don't need to be exported.

R/pipeline-helpers.R
```diff
@@ -87,3 +85,84 @@ repeat_add_columns <- function(x, y) {
...
+change_iso_code <- function(d, from, to, col='iso')
+{
+  iso <- d[[col]]
+  d[[col]] <- if_else(iso==from, to, iso)
```
bpbond: For consistency, spaces around `==` please.

R/pipeline-helpers.R
```diff
@@ -87,3 +85,84 @@ repeat_add_columns <- function(x, y) {
...
+#' duplicate rows, so you should follow it up with an appropriate aggregation.
+#'
+#' @param d The data to be transformed
+#' @param from The old ISO code
+#' @param to The replacement ISO code
+#' @param col The name of the columns with the ISO codes (default = 'iso')
+#' @return Tibble with the old iso code replaced by the new one.
+#' @export
+change_iso_code <- function(d, from, to, col='iso')
+{
```
bpbond: To be consistent with rest of code, put this brace at end of previous line.

R/pipeline-helpers.R
```diff
@@ -87,3 +85,84 @@ repeat_add_columns <- function(x, y) {
...
+#' Most of the processing functions expect ISO codes to be in lower case and in
+#' a column called ISO.
+#' @param d The data to be transformed.
+#' @param col The column currently containing the iso codes (default = 'iso')
+#' @return Tibble with the iso codes converted to lower case. If the column with
+#'   the codes was not called 'iso', it will be renamed to 'iso'.
+#' @export
+standardize_iso <- function(d, col='iso')
+{
+  d[['iso']] <- tolower(d[[col]])
```
bpbond: Here and below, worth doing `assertthat::assert_that(is_tibble(d))`?

tests/testthat/test_pipeline-helpers.R
```diff
@@ -26,3 +26,40 @@ test_that("PH_year_value_historical works", {
   expect_true(all(x$year %in% HISTORICAL_YEARS))
   expect_is(x$value, "numeric")
 })
+
```
bpbond: Thanks for test code!

R/pipeline-helpers.R
rplzzz: Good point.  I'll fix that up.

R/pipeline-helpers.R
rplzzz: Couldn't hurt.  I'll add it.

tests/testthat/test_pipeline-helpers.R
rplzzz: Job's not done without it.

R/pipeline-helpers.R
rplzzz: Come to think of it, the same is true of several of the other functions in this file.  Should I remove them from the export list too while I'm at it?

bpbond: Sure, thanks.

rplzzz: Since unexported functions can't be used in examples (including examples for the function itself!) or, apparently, tests, I've left them exported for now.  We can come back to this in a cleanup patch.

rplzzz:
Ok, @bpbond, I think that's hits everything you identified (except as noted above).

bpbond:
Thanks!

PR #236, opened 2017-04-07
Sean l132 water demand manufacturing

swd-turner:


swd-turner:
@bpbond note that the comparison test data was missing here so I just brought it over from the old system and placed it in _/tests/testthat/comparison_data_ before committing... not sure if that's standard procedure.

bpbond:
@swd-turner re test data - sorry! Perfect, that's all that I do.

R/zchunk_L132.water.demand.manufacturing.R
```diff
@@ -37,40 +37,55 @@ module_water_L132.water.demand.manufacturing_DISABLED <- function(command, ...)
...
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
+
+    # Pull out global ratio constants for manufacturing water
+    selfTotalRatio <- manufacturing_water_ratios$`self-to-total-ratio`
+    consWithRatio <- manufacturing_water_ratios$`cons-to-with-ratio`
+
+    # Get total industrial energy use for manufacturing water continent regions for 1995
+    L1322.in_EJ_R_indenergy_F_Yh %>%
+      bind_rows(L1322.in_EJ_R_indfeed_F_Yh) %>%
+      select(GCAM_region_ID, X1995) %>%
```
bpbond: We'd like to make it easy for ourselves later, when the data isn't coming in through `temp-data-inject` and presumably will be in long form, with no X's in front of years. So please: 1) in line 37 above, do `get_data(...) %>% gather(year, value, ...) %>% mutate(year = as.integer(substr(year, 2, 5)))` and 2) here and below change code appropriate. I.e., so when we match in the real data later, we only have to delete two lines at top of file.

bpbond:
Thanks @swd-turner - looks good. This is waiting on review by @mihejazi

Issue #243, opened 2017-04-11
$ Conversions


kvcalvin:
Currently, we use a series of individual $ conversions to go from one year to another (e.g., `CONV_1990_2010_USD`, `CONV_1990_1975_USD`). We should consider a different way. Maybe a function that takes in start & final year?

bpbond:
Agreed--that seems an obvious solution. Where do these conversion numbers come from? @pkyle

pkyle:
It's all from the Bureau of Economic Analysis GDP Deflators. Attached to this comment (note that the attached file was updated in 2014, and as such might have slightly different numbers for all years; not sure). But yeah a function that uses this lookup table could be pretty useful!
[BEA_GDP_Deflator_1929-2014.xlsx](https://github.com/JGCRI/gcamdata/files/914032/BEA_GDP_Deflator_1929-2014.xlsx)

bpbond:
Perfect. Thanks, Page.

PR #247, opened 2017-04-13
Add a `gdp_deflator` function

bpbond:
Generalize the many GDP conversion factors in the old data system code.
Addresses issue #243


bpbond:
For documentation, here's the difference between this function's return values when `OLD_DATA_SYSTEM_BEHAVIOR` is set or not:

```R
> summary(results$NEW_BEHAVIOR-results$OLD_BEHAVIOR)
-1.227e-02  0.000e+00  0.000e+00  6.616e-06  0.000e+00  5.897e-03 
```

diff

R/pipeline-helpers.R
```diff
@@ -167,3 +167,63 @@ unprotect_integer_cols <- function(d) {
...
+
+#' gdp_deflator
+#'
+#' Calculate a gross domestic product (GDP) implicit price deflator between two years.
+#'
+#' @param year Year for which to calculate deflator
+#' @param base_year Base year in deflator calculation
+#' @details The GDP deflator is a measure of price inflation/deflation with respect to a specific base year.
+#' @note This returns slightly different values if \code{OLD_DATA_SYSTEM_BEHAVIOR} is TRUE.
+#' @return Deflator (GDP in \code{year} divided by GDP in \code{base_year}).
```
d3y419: #' @return GDP_deflator_ratio (GDP_deflator in \code{year} divided by GDP_deflator in \code{base_year}).

R/pipeline-helpers.R
```diff
@@ -167,3 +167,63 @@ unprotect_integer_cols <- function(d) {
...
+    if(year == 1975 & base_year == 2006) return(0.3257)  # conv_2006_1975_USD
+    if(year == 1975 & base_year == 2007) return(0.317)   # conv_2007_1975_USD
+    if(year == 1975 & base_year == 2008) return(0.3104)  # conv_2008_1975_USD
+    if(year == 1975 & base_year == 2009) return(0.3104)  # conv_2009_1975_USD
+  }
+
+  # This time series is the BEA "A191RD3A086NBEA" product
+  # Downloaded April 13, 2017 from https://fred.stlouisfed.org/series/A191RD3A086NBEA
+  gdp_years <- 1929:2016
+  gdp <- c(9.896, 9.535, 8.555, 7.553, 7.345, 7.749, 7.908, 8.001, 8.347,
```
d3y419: this may confuse people. this value is actually (nominal_gdp)/(real_gdp)*100. a reader may think this is actual GDP.
i would suggest changing to gdp_deflator, but then that name is already in use. so maybe "implicit_price_index"?

R/zchunk_LB131.LV_R_GLU.R
```diff
@@ -38,7 +38,7 @@ module_aglu_LB131.LV_R_GLU <- function(command, ...) {
     # Calculate the total value of each geographic land unit (GLU)
     L100.GTAP_LV_milUSD %>%
       left_join_error_no_match(iso_GCAM_regID, by = "iso")  %>%                                         # Map in ISO codes
-      mutate(value = value * CONV_2001_1975_USD) %>%                                                    # Convert to 1975$
+      mutate(value = value * gdp_deflator(1975, base_year = 2001)) %>%                                  # Convert to 1975$
```
d3y419: per slack comment, i think this would be better served with:
mutate(value = value / gdp_deflator(2001, base_year = 1975))

Issue #165, opened 2017-03-31
Metadata incomplete for resource_water_share.csv


swd-turner:
Need source for "1995 US data" for this file...

# File: resource_water_share.csv
# Title: Fresh and saline water share and consumption to withdrawal ratios based on **1995 US data**
# Units: Unitless
# Source: ????
# Comment: Assumes 95% of Middle East water use for resource prodution from saline water

skim301:
We need to ask Mohamad or Page what the original source of that data is how 95% was chosen for Middle East.

pkyle:
I had nothing to do with this file, either collecting the data for it or keying it into the csv table...hoping @mihejazi knows?

mihejazi:
I am not sure who generated this file. I believe the first two columns came from the Davies et al 2013 paper with Page and Jae. I recall that the 95% assumption was made for the middle east along with a couple of other GCAM regions in the paper.  the paper should have all the citations (if any). In term of how the 0.425 was calculated, I suspect it was based on the USGS estimate for the US in 1995. The same case for the last two columns. The last year the USGS  produced both consumption and withdrawal estimates for the electricity sector was in 1995.

bpbond:
@swd-turner you'll work with @mihejazi to get the metadata updated and commit an update? Then we can close this issue. Thanks.

swd-turner:
Ok, we'll resolve this shortly.

PR #196, opened 2017-04-03
Sean - L110.water.demand.primary.R

swd-turner:
L110.water.demand.primary.R chunk updated.
Note: Raised issue #179 on repo regarding possible mistake. OLD_DATA_SYSTEM_BEHAVOIR trap used to return old output.

PR #249, opened 2017-04-13
Added source to metadata

swd-turner:
see issue #165


bpbond:
Sean, should this just be pushed into #236, or did you make it stand-alone for a reason? Not a big deal either way.

swd-turner:
Ok... didn't do it this way for any particular reason. Shall I go ahead and push to #236 or will I leave it as is now the pull request as already been opened?

bpbond:
Sure, leave it now. Just wondering. Thanks.

swd-turner:
My thinking was that I should do the pull request because the issue was raised initially when I was working on another chunk (#196), which has already been merged to master.

Issue #255, opened 2017-04-15
L101.nonghg_en_USA_S_T_Y


bpbond:

bpbond:
@pkyle whenever you have time - thanks.

1. The "EPA" files associated with this chunk (`EPA_CO.csv`, `EPA_NH3.csv`, `EPA_NOx.csv`, `EPA_SO2.csv`, `EPA_VOC.csv`, `EPA_tech`) don't have any header info or documentation that I see. Do you have any information on them?

2. `IEA_flow_sector.csv` and `IEA_product_fuel.csv` don't have any source information, and not clear what the unit conversion in `IEA_flow_sector.csv` is.

3. Anything to say about `GCAM_sector_tech.csv`?

bpbond:
Also, near the end of this script, it does this:

```R
#Add extra years
for ( y in X_NH3_extra_years ) {
  temp <- subset( L101.nh3_tgej_USA_en_Sepa_F_Yh.melt, L101.nh3_tgej_USA_en_Sepa_F_Yh.melt$variable == "X1990" )
  temp$variable <- y
  L101.nh3_tgej_USA_en_Sepa_F_Yh.melt <- rbind( L101.nh3_tgej_USA_en_Sepa_F_Yh.melt, temp )
}
```

It looks like it's extending 1990 NH3 data back to 1971. Why?

pkyle:
This module was developed by @kvcalvin ; I'll answer the ones that I know the answers to or that I was otherwise responsible for, and leave the others open.
1) Leaving this open
2) These are mapping files, that map from the categories in the IEA Energy Balances to our intermediate categories for processing. For a source you can use: International Energy Agency, 2012, Energy Balances of OECD Countries 1960-2010 and Energy Balances of Non-OECD Countries 1971-2010. As for the units, the conversion factors convert from the units in IEA Energy Balances (kilotonnes of oil equivalent, terajoules of heat, and gigawatt-hours of electricity) to exajoules, and all energy inputs are indicated as positive numbers. In the IEA Energy Balances, the standard adopted is that for energy production and transformation sectors, energy inputs are negative and energy outputs are positive, whereas for energy consumption sectors, energy inputs are positive. We don't use that convention in GCAM; all energy inputs (and outputs) are indicated as positive numbers.
[...]

PR #256, opened 2017-04-16
Ben - L101.nonghg_en_USA_S_T_Y

bpbond:
Code translated. Note a few comments and input headers need to be filled in/clarified (see #255 ).


bpbond:
@rynacui - do you mind doing a code review on this? Thanks.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
```diff
@@ -0,0 +1,244 @@
...
+      mutate(technology = fuel,
+             fuel = paste(mode, size.class, sep = "_"),
+             sector = UCD_sector) %>%
+      select(-mode, -UCD_sector, -size.class, -UCD_technology, -UCD_fuel) %>%
+
+      # Bind all together
+      bind_rows(L1231.in_EJ_R_elec_F_tech_Yh, temp) ->
+      L101.in_EJ_R_en_Si_F_Yh
+
+    L101.in_EJ_R_en_Si_F_Yh %>%
```
rynacui: Are these temp-data with `Xyear` & wide format going to be replaced by outputs with `year` & long format later? And these lines will be taken out?

bpbond: Maybe, but this is a little tricky. It turns out that there are multiple lines for each combination of year and sector, fuel, region, and technology, which means that once we `gather` it, the dataset can't be re-`spread` (an error is thrown if we try). So we may need to pass these data in wide format, although definitely without the Xyears.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
```diff
@@ -0,0 +1,244 @@
...
+
+    # Compute SO2, CO, NOx, VOC, and NH3 emissions factors by dividing EPA inventory by IEA energy balances
+    # We do this for each gas, so define a function with all the steps
+    EPA_compute_emissions_factors <- function(x, L101.in_EJ_USA_en_Sepa_F_Yh.mlt) {
+      L101.in_EJ_USA_en_Sepa_F_Yh.mlt %>%
+        left_join(x, by = c("EPA_agg_sector" = "sector", "EPA_agg_fuel" = "fuel", "year" = "year")) %>%
+        rename(emissions = value) %>%
+        mutate(emissions = if_else(is.na(emissions), 0, emissions),
+               em_factor = emissions / energy,
+               em_factor = if_else(is.nan(em_factor) | is.infinite(em_factor), 0, em_factor)) ->
```
rynacui: Why `is.nan` here instead of `is.na`? Are they different?

bpbond: Yes, `NaN` means "not a number", for example `sqrt(-1)`. This logic follows the original code, except replacing a string comparison (`em_factor == "NaN"`) with `is.nan`.

rynacui: Got it. Thanks!

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
```diff
@@ -0,0 +1,244 @@
...
+        x_em_factor
+
+      # Compute sector emissions. If these are zero, then reset all fuel em_factors to 1
```
rynacui: Just out of curiosity, when sector emissions equal zero, why em_factors are set to 1 Tg/EJ?

bpbond: Good question; the documentation should make this clear. @pkyle could you weigh in?

bpbond:
Thanks Ryna!

cahartin:
@bpbond and @rynacui The emissions inputs need to be updated with title and source.

bpbond:
@cahartin right, thanks. I have queries out to @pkyle in #255

cahartin:
@bpbond gotcha. thanks.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
pkyle: I looked at the code but also couldn't tell; @kvcalvin  can you comment?

bpbond: @kvcalvin - when you get a chance

kvcalvin: Good question. I don't totally remember, but I think there was a specific case where this was required. Essentially, we will later scale these emissions to match EDGAR totals. If there is a zero here, but positive emissions in EDGAR, then the scaling won't work.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
d3y419: see also: https://github.com/JGCRI/gcamdata/pull/309

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
d3y419: https://github.com/JGCRI/gcamdata/issues/327

PR #268, opened 2017-04-19
Improve screening for use of forbidden functions

rplzzz:
The screening function is now an actual (internal) function in the
package, with tests of its own, which the chunk test code calls on
each module function.  In the process we corrected some broken regular
expressions and added some edge cases.  This should eliminate some
potential false positives in the screening.  Finally, we have upgraded
the info message when this test fails to show the lines that have
forbidden calls in them.

We have also added `merge` to the list of forbidden functions.
Unfortunately, this breaks a formerly working chunk, but at least we
got to test the new info message.

Ironically, the new screening function uses `rbind` and `cbind` (which
are both on the forbidden list) internally to build its output matrix.


rplzzz:
This PR is going to fail its tests on Travis because it adds `merge` to the list of forbidden functions, and there is currently a chunk that uses it.

rplzzz:
It wasn't quite quitting time, so I went in and replaced the `merge` calls with `full_join`.  Tests now pass; however, there is _definitely_ something screwy with the old-new comparison, as a bunch of random chunks are failing on my local copy, despite the fact that I didn't touch them in this branch and the fact that they pass on Travis.  I believe a couple of others have reported similar anomalies, so this is probably something we should track down.

bpbond:
Thanks @rplzzz . Yeah, I haven't yet isolated what's going on with the sporadic-test-failure thing; some people see it, some don't.

Issue #91, opened 2017-03-23
L210.water.demand.primary


bpbond:

yalingliupnnl:
@bpbond There is no zchunk for L210 in the new gcamdata system.

bpbond:
Whoops, sorry, I'll commit.

PR #274, opened 2017-04-21
Translate `get_water_inputs_for_mapping`

bpbond:
Need a better name! For @yalingliupnnl chunk L210 #91


pkyle:
Hey guys, I'm not sure how to go about testing this; tried checking out the water-helper branch but it doesn't have the L210 chunk to the point that it's ready to use the function, and (as far as I can tell) I can't run it on the old data system b/c all of the structures and precursors are different. Sorry if I'm missing something obvious--Ben I can stop by tomorrow if that would help.

PR #278, opened 2017-04-21
Fix test failures caused by change in readr behavior from v. 1.0 to 1.1

rplzzz:
Adds `FLAG_PROTECT_FLOAT` for tables that were having problems with floating point values being mangled on output.

This patch also takes care of a couple of bugs in the test code that were causing tests to be silently skipped sometimes.  One side effect of these fixes is that tests now pass when run with `devtools::test()` and when run under Travis, but they fail when run under `R CMD CHECK`.  This is because CHECK views the outputs directory as unnecessary and deletes it.  Formerly this caused the tests to pass by default; now it is an error because the outputs directory doesn't exist.     `¯\_(ツ)_/¯`

I also tweaked the documentation for `gdp_deflator` to make it a little clearer how it is meant to be used, and I fixed an instance where it was being used in a confusing way.


tests/testthat/test_oldnew.R
```diff
@@ -95,6 +101,5 @@ test_that("matches old data system output", {
         expect_equivalent(round_df(olddata), round_df(newdata), label = paste(basename(newf), "doesn't match"))
```
bpbond: Should be `info =`?

tests/testthat/test_oldnew.R
```diff
@@ -8,15 +8,21 @@ test_that("matches old data system output", {
...
+  ## Look for output data in OUTPUTS_DIR under top level (this code will be run in
+  ## tests/testthat)
+  outputs_dir <- normalizePath(file.path("../..", OUTPUTS_DIR))
+
   if (identical(Sys.getenv("TRAVIS"), "true")) {
-    driver(write_outputs = TRUE)
+    driver(write_outputs = TRUE, outdir = outputs_dir)
   }
+  expect_equivalent(file.access(outputs_dir, mode=4), 0,  # outputs_dir exists and is readable
+                    info = paste('Directory', outputs_dir, "unreadable or does not exist."))
```
bpbond: "Directory" for consistency pls

R/zchunk_L100.GDP_hist.R
```diff
@@ -40,7 +40,8 @@ module_socioeconomics_L100.GDP_hist <- function(command, ...) {
       gather(year, value, -iso) %>%
       mutate(year = as.numeric(year)) %>%
       filter(year %in% HISTORICAL_YEARS, !is.na(value), !is.na(iso)) %>%
-      mutate(value = value * CONV_BIL_MIL / gdp_deflator(2010, base_year = 1990)) %>%
+      mutate(value = value * CONV_BIL_MIL * gdp_deflator(1990, base_year = 2010),
```
bpbond: 👍

R/utils.R
```diff
@@ -197,13 +197,27 @@ find_csv_file <- function(filename, optional, quiet = FALSE) {
...
+#' Write data produced by chunks to csv files.
+#'
+#' Write the data produced by the chunks to their output files.  This is mostly
+#' a wrapper around \code{write_csv} that figures out file names, processes
+#' table flags, writes metadata comments, and so forth.
+#'
+#' One thing to be aware of is that there is a wart in \code{readr} v 1.1 and
+#' later that causes floating point data to be written as integers if they
+#' happen to have integer values.  This can cause problems if there are so many
+#' apparently-integer values before the first obviouisly-float value that
```
bpbond: obviously (fixed)

tests/testthat/test_oldnew.R
```diff
@@ -8,15 +8,21 @@ test_that("matches old data system output", {
...
+
+  ## Look for output data in OUTPUTS_DIR under top level (this code will be run in
+  ## tests/testthat)
+  outputs_dir <- normalizePath(file.path("../..", OUTPUTS_DIR))
+
   if (identical(Sys.getenv("TRAVIS"), "true")) {
-    driver(write_outputs = TRUE)
+    driver(write_outputs = TRUE, outdir = outputs_dir)
   }
+  expect_equivalent(file.access(outputs_dir, mode=4), 0,  # outputs_dir exists and is readable
```
bpbond: Style consistency: `mode = 4`

tests/testthat/test_oldnew.R
bpbond: fixed

bpbond:
Hmm, while all tests pass, the package build process fails. Pushing to GitHub to see if this occurs on Travis.

bpbond:
Ah, I see what's happening. @rplzzz when running "Check package" the R build system strips away `outputs/` (as that directory is listed in `.Rbuildignore`) which means that the new lines 19-21 in `test_oldnew.R`, that make sure `outputs/` exists and is readable, fail. I've moved these two checks into the Travis-only block.

rplzzz:
Yeah, I wasn't quite sure about the best way to handle the R CMD CHECK behavior.  As it stands, I'll bet we get at least one push at some point that fails on travis because it's possible for the local tests not to actually see the new stuff a developer has added.  But, I guess worse things happen at sea (or in C, for that matter).

Issue #194, opened 2017-04-03
Replacement for `match`


bpbond:
>rpl [10:08 AM] 
>If we’re going to use `distinct` and `join`, we might want to roll them into a single utility function.

Yes - currently the wiki says use these two to replicate behavior of `match` but we should provide a function?

rplzzz:
I'm indifferent here, as you might have guessed from the fact that I argued for a utility function here and against it in issue #288.  I guess one advantage to having a utility function is that when we eventually want to retire that functionality, we can easily mark the utility as deprecated and get warnings everywhere it is used.  That's assuming, of course, that we don't think there is ever a case outside of backward compatibility where we would want to have this behavior.

In any case, this seems like a good end-of-day project.  I'll knock it out, and you can see what you think of it.

bpbond:
Ok - thanks.


On Apr 25, 2017, at 4:22 PM, Robert Link <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm indifferent here, as you might have guessed from the fact that I argued for a utility function here and against it in issue #288<https://github.com/JGCRI/gcamdata/issues/288>. I guess one advantage to having a utility function is that when we eventually want to retire that functionality, we can easily mark the utility as deprecated and get warnings everywhere it is used. That's assuming, of course, that we don't think there is ever a case outside of backward compatibility where we would want to have this behavior.

In any case, this seems like a good end-of-day project. I'll knock it out, and you can see what you think of it.

rplzzz:
Here's what I wrote when we discussed it in #287 

> I'm a little leery of having several different variations on wrappers around left_join. Probably what we need are some option arguments. For example:
>
> * .ONE_TO_ONE: Error if multiple matches on the RHS.
> * .FIRST_ONLY: Include the first match only
> * .NO_NA: Assert no NA values in the new columns.
>After excluding the combinations that don't make sense, that subsumes six possibilities. Of course, then we have to come up with test data to test all of the possible success and failure modes for all of those possibilities, which will most likely take longer than developing the function itself (which is probably why people are so tempted to skimp on tests.)

I started to work on this yesterday, and the code is actually pretty easy to do.  Writing documentation to explain it all so that someone unfamiliar with what we were doing wouldn't be utterly confused was actually taking longer.  Writing tests will be a bit tedious too, as we'll have a lot of cases to cover.  Maybe I'll push the code up as a PR, and we can get someone else to polish up the docs and tests.

[...]

PR #287, opened 2017-04-24
Robert - module_socioeconomics_L102.GDP

rplzzz:
I confess, enough time has passed that I don't remember how we decided to assign reviewers.  Can someone remind me?


bpbond:
Hi @rplzzz - we've asked people to pick their team leader (in this case @d3y419 ) who can either review or assign to someone with particular domain expertise, as appropriate.

d3y419:
i'll review the substance. @bpbond will review the code?

R/constants.R
```diff
@@ -19,6 +19,7 @@ FLAG_PROTECT_FLOAT <- "FLAG_PROTECT_FLOAT"
 # Time constants
 HISTORICAL_YEARS <- 1971:2010
 FUTURE_YEARS <- seq( 2015, 2100, 5 )
+IMF_FUTURE_YEARS <- seq( 2010, 2020, 1 )
```
bpbond: No GCAM-style spaces with parentheses and brackets please.

R/constants.R
```diff
@@ -64,9 +70,10 @@ aglu.CCONTENT_CELLULOSE <- 0.45
 energy.CDIAC_CO2_HISTORICAL_YEARS <- HISTORICAL_YEARS[HISTORICAL_YEARS < 2010]
 
 
-# ======================================================================
-# Conversion constants
+## ======================================================================
+## Conversion constants.  The naming convention is CONV_(FROM-UNIT)_(TO-UNIT).
```
bpbond: Thanks

R/zchunk_L102.GDP.R
```diff
@@ -1,39 +1,118 @@
...
+
+    ## Find the base year
+    base.year <- max(dplyr::intersect(past$year, future$year))
+    ## Base year gdp from the future dataset
+    baseyear.future.gdp <- filter(future, year==base.year) %>%
+        rename(base.gdp = gdp) %>%
+        select(-year)
+
+    gdp.future.ratio <- filter(future, year > base.year) %>%
+        left_join_error_no_match(baseyear.future.gdp, by=c('scenario',grouping)) %>%
```
bpbond: Generally put a space after commas pls

R/zchunk_L102.GDP.R
```diff
@@ -1,39 +1,118 @@
...
+        select(-year)
+
+    rslt <- left_join(baseyear.past.gdp, gdp.future.ratio,
+                      by=c('scenario',grouping)) %>%
+        mutate(gdp = base.gdp * gdp.ratio) %>%
+        select(one_of(c('scenario', grouping, 'year', 'gdp'))) %>%
+        bind_rows(gdp.past, .)
+
+    if(drop.scenario)
+        select(rslt, -scenario)
```
bpbond: Add explicit braces around this and `else` clause

R/zchunk_L102.GDP.R
```diff
@@ -42,121 +121,204 @@ module_socioeconomics_L102.GDP_DISABLED <- function(command, ...) {
     # Load required inputs
     iso_GCAM_regID <- get_data(all_data, "common/iso_GCAM_regID")
     SSP_database_v9 <- get_data(all_data, "socioeconomics/SSP_database_v9")
-    GCAM3_GDP <- get_data(all_data, "socioeconomics/GCAM3_GDP")
     IMF_GDP_growth <- get_data(all_data, "socioeconomics/IMF_GDP_growth")
     L100.gdp_mil90usd_ctry_Yh <- get_data(all_data, "L100.gdp_mil90usd_ctry_Yh")
     # Temporary data injection from old data system
```
bpbond: Remove

R/zchunk_L102.GDP.R
```diff
@@ -42,121 +121,204 @@ module_socioeconomics_L102.GDP_DISABLED <- function(command, ...) {
...
+
+    gdp_bilusd_rgn_Yfut[['scenario']] <- substr(gdp_bilusd_rgn_Yfut[['scenario']], 1, 4)
+    gdp.mil90usd.SSP.rgn.yr <- join.gdp.ts(gdp_mil90usd_rgn, gdp_bilusd_rgn_Yfut, 'GCAM_region_ID')
+
+    ## Get the IMF GDP growth rates.  Some countries are missing, so we have to
+    ## add them in with an assumed zero growth rate.
+    imfgdp.growth <-
+        select(IMF_GDP_growth, one_of(c('ISO', IMF_FUTURE_YEARS))) %>%
+          standardize_iso('ISO') %>% change_iso_code('rou', 'rom') %>%
+          tidyr::gather(year, gdp.rate, -iso) %>%
```
bpbond: No need for `tidyr::` prefix

R/zchunk_L102.GDP.R
```diff
@@ -42,121 +121,204 @@ module_socioeconomics_L102.GDP_DISABLED <- function(command, ...) {
...
+    pop.thous.fut <-
+        rename(L101.Pop_thous_Scen_R_Yfut, population = value) %>%
+          filter(year %in% FUTURE_YEARS)
+    pop.thous.hist <-
+        rename(L101.Pop_thous_R_Yh, population = value) %>%
+          filter(year %in% HISTORICAL_YEARS) %>%
+          tidyr::crossing(scenario = unique(pop.thous.fut[['scenario']]))
+    pop.thous.scen.rgn.yr <-
+        bind_rows(pop.thous.hist, pop.thous.fut) %>%
+          mutate(year = as.integer(year), population=as.numeric(population)) %>%
```
bpbond: Be consistent: spaces around `=`

R/zchunk_L102.GDP.R
```diff
@@ -42,121 +121,204 @@ module_socioeconomics_L102.GDP_DISABLED <- function(command, ...) {
...
+
+    ## calculate per-capita GDP.  This is another final output
+    pcgdp.thous90usd.scen.rgn.yr <-
+        left_join(gdp.mil90usd.scen.rgn.yr, pop.thous.scen.rgn.yr,
+                  by = c('scenario', 'GCAM_region_ID', 'year')) %>%
+          mutate(pcgdp = gdp / population) %>%
+          select(scenario, GCAM_region_ID, year, pcgdp)
+
+    ## Calculate the PPP-MER conversion factor for each region.  As far as I can
+    ## tell, we do this only in 2010 and assume it to be constant for all time
```
bpbond: Worth entering as an issue to follow up on?

tests/testthat/test_gdp_helpers.R
```diff
@@ -0,0 +1,39 @@
+## Test the gdp join function
+
+context('gdp-join')
+library('tibble', warn.conflicts=FALSE)
+library('magrittr', warn.conflicts=FALSE)
+
```
bpbond: Don't need these two `library` calls.

tests/testthat/test_gdp_helpers.R
```diff
@@ -0,0 +1,39 @@
...
+## set up test data
+past <- tibble(year=c(2000,2005,2010)) %>%
+  tidyr::crossing(iso=c('xxl', 'xxp'))  %>%
+  dplyr::mutate(gdp=if_else(iso=='xxl', 10*year, 5*year))
+
+future <- tibble(year=rep(c(2005, 2010, 2015), 2), iso=c(rep('xxl',3), rep('xxp',3)),
+                 gdp=c(cumprod(rep(1.05,3)), cumprod(rep(1.02,3)) ) ) %>%
+  tidyr::crossing(scenario=c('gud','bad')) %>%
+  dplyr::mutate(gdp=if_else(scenario == 'gud', gdp, gdp*0.9^((year-2005)/5)))
```
bpbond: Don't need `dplyr::` here

tests/testthat/test_gdp_helpers.R
```diff
@@ -0,0 +1,39 @@
...
+
+test_that('Joining time series with future scenarios works', {
+  rslt <- gcamdata:::join.gdp.ts(past, future, 'iso')
```
bpbond: Nor `gcamdata:::` here and below

R/zchunk_L102.GDP.R
```diff
@@ -42,121 +121,204 @@ module_socioeconomics_L102.GDP_DISABLED <- function(command, ...) {
...
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    ppp.mer.rgn %>%
+      add_title("Purchasing Power Parity (PPP) to Market Exchange Rate (MER) GDP conversions, by region") %>%
+      add_units("unitless") %>%
+      add_comments("Calculated as GDP(PPP) / GDP(MER) in the base year (2010) for each region.  The") %>%
+      add_comments("table also contains PPP-GDP and MER-GDP in billions of 2005 USD for the base year, ") %>%
+      add_comments("because they were included in the original table, but I'm not sure they are used for, ") %>%
```
d3y419: i think these two comments can be in the code, but no need for it in the table itself.

R/zchunk_L102.GDP.R
```diff
@@ -1,39 +1,118 @@
...
+#' joined.  The units of the output time series will be the same as the units of
+#' \code{past}.
+#'
+#' @param past Tibble with the past time series (year, gdp, and grouping).
+#' @param future Tibble with the future data (year, gdp, scenario, and
+#' grouping).
+#' @param grouping Name of the grouping column (generally either 'iso' or
+#' 'GCAM_region_ID', but could be anything
+#' @return Time series with the past and future joined as described in details.
+join.gdp.ts <- function(past, future, grouping) {
```
d3y419: this function is great!

R/zchunk_L102.GDP.R
```diff
@@ -1,39 +1,118 @@
...
+#' GDP outputs are in millions of 1990 USD, Market Exchange Rate (measured in 2010) is used for foreign currency. 
+#' Per-capita values are in thousands of 1990 USD.
+#'
 #'
 #' @param command API command to execute
 #' @param ... other optional parameters, depending on command
 #' @return Depends on \code{command}: either a vector of required inputs,
 #' a vector of output names, or (if \code{command} is "MAKE") all
-#' the generated outputs: \code{L102.gdp_mil90usd_Scen_R_Y}, \code{L102.pcgdp_thous90USD_Scen_R_Y}, \code{L102.gdp_mil90usd_GCAM3_R_Y}, \code{L102.gdp_mil90usd_GCAM3_ctry_Y}, \code{L102.pcgdp_thous90USD_GCAM3_R_Y}, \code{L102.pcgdp_thous90USD_GCAM3_ctry_Y}, \code{L102.pcgdp_thous90USD_GCAM3_ctry_Y}, \code{L102.PPP_MER_R}. The corresponding file in the
+#' the generated outputs: \code{L102.gdp_mil90usd_Scen_R_Y}, \code{L102.pcgdp_thous90USD_Scen_R_Y}, \code{L102.gdp_mil90usd_GCAM3_R_Y}, \code{L102.gdp_mil90usd_GCAM3_ctry_Y}, \code{L102.pcgdp_thous90USD_GCAM3_R_Y}, \code{L102.pcgdp_thous90USD_GCAM3_ctry_Y}, \code{L102.PPP_MER_R}. The corresponding file in the
```
d3y419: i'm not entirely sure what happened to GCAM3 outputs.
while i fully support moving from GCAM3 legacy to GCAM4 gdp, if we're making that change, we should probably get consensus from a wider group. e.g. @pkyle @leon @sha

rplzzz: I deleted them.  They're a huge hassle to produce (they take as many lines of code as the rest of the module put together), and nobody is using them for anything.  Carrying dead code is bad enough, but expending hours to modernize and update dead code is a huge waste of resources.

R/zchunk_L102.GDP.R
rplzzz: Are you sure?  When I first wrote this function it choked on `gather` calls without the namespace qualifier.  Have we added tidyr as an imported namespace since then?

bpbond: Works for me without, anyway.

R/zchunk_L102.GDP.R
rplzzz: I'm not sure exactly what parts of the comments you're referring to, but the table comments should document the contents of the table.  I found it a little nonintuitive that a table of PPP-MER conversion factors should also contain the numbers that were divided to produce the factors; hence the explanation.

R/zchunk_L102.GDP.R
rplzzz: Thanks.

R/zchunk_L102.GDP.R
rplzzz: To that I'll add, if someone thinks they might need to use these some day, they are welcome to add them back.  I would recommend doing them in a separate chunk, using the outputs of this chunk inputs to the new chunk where required.

I would also recommend that the work to reinstate the GCAM3 scenarios be charged to the project that actually requires them because everybody I've talked to seems to think we have no use for them in the SFA.

R/zchunk_L102.GDP.R
```diff
@@ -87,9 +87,9 @@ join.gdp.ts <- function(past, future, grouping) {
 #' by GCAM region.
 #'
 #' The scenarios generated include the SSPs, the gSSPs (SSPs modified by
-#' near-term IMF projections), and, for some reason, something derived from
-#' GCAM3.  GDP outputs are in millions of 1990 USD, which presumably means they
-#' are MER values.  Per-capita values are in thousands of 1990 USD.
```
rplzzz: Sorry about this.  These were part of notes to myself about things to follow up on.  I cleaned out most of them, but missed these.

R/constants.R
rplzzz: ✅

R/zchunk_L102.GDP.R
pkyle: I also support the removal of GCAM3 socioeconomics--no sense in re-developing unused materials.

R/constants.R
rplzzz: Strange that GitHub doesn't allow some variant of ✅ and ❌ as reactions to comments.  They would seem a natural way of indicating status on reviews.

R/zchunk_L102.GDP.R
rplzzz: ✅

R/zchunk_L102.GDP.R
rplzzz: ✅

R/zchunk_L102.GDP.R
rplzzz: ✅

R/constants.R
bpbond: There's an "Add your reaction" pop-up menu available in the upper-right corner of every comment.

R/zchunk_L102.GDP.R
bpbond: Agreed; we definitely want to capture at least a brief description of how each table was generated attached to each object.

R/constants.R
rplzzz: Yes, but it only allows certain ones (specifically the ones I've added above).  The check mark and cross-mark would be useful to have as options.

R/zchunk_L102.GDP.R
rplzzz: Yeah, we have an importFrom for tidyr up there.  Must have gotten added somewhere along the line.  Since I rebased a couple of times, it's not so obvious what my starting point was anymore.

rplzzz: ✅

R/zchunk_L102.GDP.R
rplzzz: ✅

R/zchunk_L102.GDP.R
rplzzz: No, I think this is working as designed, and it's not obvious (to me at least) what you would do instead.  It just confused me a bit when I was first working through it.  Haewon updated the comment a bit to clarify that these assumptions are by design.

inst/extdata/socioeconomics/GCAM3_GDP.csv
```diff
@@ -1 +1,16 @@
...
Korea,71339,230019,538839,607069,656741,707030,756512,810748,870628,941268,1017410,1099640,1187740,1282620,1378260,1475150,1570990,1668740,1766150,1880310,2024860,2151700
Latin America,814694,1187430,1919720,2234190,2553590,2949630,3368380,3834340,4299980,4873900,5476220,6236520,7014730,8019950,9193280,10738400,12553700,14719600,17144300,19760200,22336300,25366900
Middle East,298287,354600,649161,799322,974253,1174110,1392710,1635480,1889650,2189190,2499060,2852530,3190200,3578500,3967850,4432780,4896340,5423610,5952670,6525370,7119600,7769320
Southeast Asia,222332,593705,1230630,1460320,1961080,2601350,3372790,4305090,5403760,6689470,8210530,9959920,12061300,14448000,17201800,20092200,23375600,26794100,30544100,34509200,38874800,43534900
USA,3473910,5728530,9104140,9600970,10756900,11914400,13175100,14458700,15829500,17465700,19164900,21251000,23198700,25517400,27872600,30669200,33583600,36744000,39970500,43838800,48155500,52760200
Western Europe,3802400,5646240,7801260,7930550,8546440,9184410,9843660,10470000,11052000,11624200,12189200,12795600,13407800,14089100,14736400,15486200,16180100,16919800,17638600,18551700,19738200,20254600
```
rplzzz: It's possible we could remove this file entirely, since we're not using it anymore in this chunk, and I don't think it was used anywhere else.  I will investigate.

tests/testthat/test_gdp_helpers.R
rplzzz: ✅

tests/testthat/test_gdp_helpers.R
rplzzz: ✅

tests/testthat/test_gdp_helpers.R
rplzzz: @bpbond I'm stumped by this one.  Because `join.gdp.ts` is not exported, it shouldn't be visible to the test code, even though the `gcamdata` namespace is imported.  What's more, I'm pretty certain that it failed for precisely that reason when I first wrote the code.  However, I just tested it, and sure enough, we don't need the namespace qualifier now.  How are we getting around the fact that that symbol is not exported?

bpbond: @rplzzz I think this is because `testthat` uses `devtools::load_all`. From the latter's help page:

>Normally when loading a package, the objects listed as exports in the NAMESPACE file are copied from the namespace to the package environment. However, load_all by default will copy all objects (not just the ones listed as exports) to the package environment. This is useful during development because it makes all objects easy to access.

rplzzz: Weird.  I could have sworn that this one failed for me when I first tried it without the qualifier.  Maybe I forgot to reload the package after adding the function and when I changed the test I reloaded then.

Anyway: ✅

R/constants.R
pkyle: I just happened to notice this from the code snippet above--I don't like the "IMF_FUTURE_YEARS" object 2010:2020, because within that dataset, the "future" years differ by nation, and for most the 2010:2014 time period is actual historical data, with 2015:2020 being the future years. What do you guys think about just calling it IMF_YEARS?

R/zchunk_L102.GDP.R
pkyle: This is correct--we only need PPP:MER for future years in the model, as some of our functions (non-CO2 control functions for sure, maybe others?) use the PPP GDP. The C++ code for evolving the PPP:MER ratio over time is in gdp.cpp, but right now that is de-activated in the socioeconomics_*.xml input files by setting the constRatio attribute. For example:
<PPPConvert constRatio="1">0.9534</PPPConvert>
Note that this is necessary for using the SSPs as the basis for the GDP projections, as these are in PPP. If we wanted to evolve the PPP:MER over time in any regions as per the code in gdp.cpp, we would need to back out the MER GDPs from the PPP assumptions, using those functional forms.

R/zchunk_L102.GDP.R
d3y419: specifically: "but I'm not sure they are used for"

bpbond:
This is waiting on approval from @pkyle and/or @d3y419

bpbond:
@rplzzz may need to rebuild NAMESPACE for on this. Sorry.

bpbond:
Test code is throwing a `left_join_error_no_match` error.

rplzzz:
@bpbond Is that the same error I reported in #295?  If so, then it's present in the base branch.

The fundamental problem here is that `left_join_error_no_match` uses the presence of `NA` values as a proxy for failed matches.  The problem is that there may very well have been `NA` values in the right-side table to start with, and it's not clear that this function should be failing on those.  The point here is to assert that every entry in the left-side table has a match, not to assert anything in particular about what is in those matches on the right side.

bpbond:
@rplzzz No - or at least, that wasn't my intention on writing `left_join_error_no_match`. NA isn't intended to be a proxy for failed matches as such, but rather by calling this function we're saying that no NA should _ever_ appear as a result. In other words, if you have a join you want to do that might result in NA, don't call it.

Perhaps ideally we'd have `left_join_error_no_match` (no NAs allowed); `left_join_1to1` (no change in dataset size allowed); and `left_join_error_no_match_121` (both).

rplzzz:
I see.  We had a miscommunication on what the intended use of the function was.  I'm a little leery of having several different variations on wrappers around `left_join`.  Probably what we need are some option arguments.  For example:
* `.ONE_TO_ONE`: Error if multiple matches on the RHS.
* `.FIRST_ONLY`:  Include the first match only
* `.NO_NA`: Assert no `NA` values in the new columns.

After excluding the combinations that don't make sense, that subsumes six possibilities.  Of course, then we have to come up with test data to test all of the possible success and failure modes for all of those possibilities, which will most likely take longer than developing the function itself (which is probably why people are so tempted to skimp on tests.)

bpbond:
Something that like that makes sense. Let's continue discussion in #194 as necessary.

bpbond:
@d3y419 is this OK to merge from your point of view? Or are we still waiting on consensus re GCAM3?

pkyle:
Hey guys, I'm late to the game...just reviewing it now (tried yesterday!), and there are a few things I think ought to be looked at...I'll just post one at a time. First is that the read_csv function seems to be doing that thing where it just scans not quite enough of the data frame to figure out what is what, and in the object gdp_bilusd_rgn_Yfut, before the as.numeric(gdp) command is executed, the data associated with the 2000 and 2005 years are considered characters. The reason is that the first model and variable in the SSP database didn't report these years. Is this as.numeric() method how we're treating this? I thought we had re-configured read_csv to scan the whole table.

pkyle:
I'd remove ", it turns out," from the comment "These, it turns out, are PPP values in 2005 dollars" (line 145)

pkyle:
Looks like line 168 can be removed (redundant with line 159)

pkyle:
That join.gdp.ts() function is great! Looks like it might be usable for other variables where the past is known, the future is exogenous and scenario-dependent, and we want to put together a single continuous time series of both past and future. Should we move it to module-helpers.R so it's accessible to others?

R/constants.R
rplzzz: Makes no nevermind to me.

pkyle:
In IMF_GDP_growth, I don't like that the data columns 2010 and 2011 are being interpreted as character variables because the string "n/a" is not read in as a value to na.strings. @ben is the na.strings argument user-configurable and can we add "n/a" to the list? It looks like get_data() is not the place as the data have already been read in. But there aren't any datasets where "n/a" would mean anything other than a missing value. [Unrelated: note that the string "NA" is a product code for natural gasoline in GCAM-USA]

pkyle:
Sorry still going...can you please delete the line 249 comment: "Assume this conversion to be constant beyond 2010." The assumption of constant PPP:MER is made later on (in chunk L201.), and future users/developers may want to re-enable GCAM's endogenous PPP:MER evolution.

pkyle:
The comment around line 263--regions can and will change over time, so can you replace "region 21" with "the region containing the Palestinian Territories (pse)"?

rplzzz:
@pkyle When you have comments that pertain to specific lines, could you comment on that line in the diff (go to the diff view, mouse over the line, click the '+' next to the line number)?  That makes it easy to find the line you're talking about while still looking at your comment.

R/zchunk_L102.GDP.R
d3y419: RESOLVED. GCAM3 legacy socioeconomics will be deprecated.

bpbond:
@d3y419 - thanks, but in the future please let me do merges. (I have a couple final checks I perform.)

Issue #272, opened 2017-04-20
Intermittent test failures


bpbond:
We're seeing intermittent test failures--intermittent from machine to machine, not in time--that we think may have to do with different versions of the `readr` package (I'm on 1.0.0, while people like @rplzzz and @kvcalvin are on 1.1.0).

bpbond:
Seems like we're getting bitten by tidyverse/readr#645

rplzzz:
I have a fix that I'll push up as soon as this telecon is over.  It involves scanning the entire file on read-in, which is obviously not ideal.  I haven't done detailed timing, but it seems pretty slow.  I'm wondering if a better solution would be to see if `write.csv` (instead of `write_csv`) produces valid output.  If so, then by using that to do our output we could continue to use `read_csv` with its default scan length.

bpbond:
OK, thanks. I bet it's slow. Does your fix involve scanning _every_ file? As we discussed, one thing that might help would be a `FLAG_SCAN_ENTIRE_FILE`.

bpbond:
We could also force people to use `readr` 1.0.

rplzzz:
Yes it does, so that's a possibility, but I still think writing data with base R might work for us.  Are there any specific warts in `write.csv` we need to watch out for (other than row numbers)?

Forcing people to use 1.0 isn't a great idea, IMO.  Forced downgrades are always a pain..

bpbond:
No, that would be a last resort, I agree.

AFAIK the biggest issue with `write.csv` will be speed; it's significantly slower than `write_csv`.

rplzzz:
Maybe a `protect_float` argument to our write wrapper that causes it to use base R?

bpbond:
Or our write routine could check for integer-ish numeric columns and change them to `int` before writing?

bpbond:
>Maybe a protect_float argument to our write wrapper that causes it to use base R?

Note this would also depend on a flag mechanism (`FLAG_WRITE_BASE_R`).

rplzzz:
Right, I figured that out when I went to implement the other solution.  We don't write each chunk individually; we call a function that writes them in batch.

bpbond:
Fixed in #278

PR #290, opened 2017-04-25
Aaron - LA101.EIA_SEDS.R

aarony2j:
1. No errors
2. Tried left_join_error_no_match, and it didn't work (at least for the first left_join)
3. Not sure whether A_fuel_conv.csv and EIA_use_all_Bbtu.csv should be placed in gcam-usa folder or temp-data-inject. I put it in gcam-usa.

#140


R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -1,20 +1,22 @@
...
 #' @importFrom tidyr gather spread
-#' @author YourInitials CurrentMonthName 2017
-#' @export
-module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
+#' @author AS April 2017
+#'
+module_gcam.usa_LA101.EIA_SEDS <- function(command, ...) {
   if(command == driver.DECLARE_INPUTS) {
     return(c(FILE = "gcam-usa/EIA_SEDS_fuels",
              FILE = "gcam-usa/EIA_SEDS_sectors",
```
bpbond: Right. Only put in `temp-data-inject/` if you're temporarily using old system data that will eventually be provided by another chunk.

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
-    #
-    # NOTE: there are `merge` and/or 'match' calls in this code. Be careful!
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Merge-and-Match
+
+    # Define vector to be used for filtering
+    GCAM_year <- 1971:2010
+
+    # Prep for output tables - add columns for GCAM sector and fuel names, using the substrings of the MSN code, and filter out U.S.
+    EIA_use_all_Bbtu %>%
+      gather(year, value, 4:55) %>%
```
bpbond: Please don't use hard-coded column numbers like this. Do `gather(year, value, -col_name1, -col_name2, -col_name3)` or something similar.

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
+      mutate(year = as.integer(year)) %>%
+      mutate(EIA_fuel = substr(MSN, 1, 2 )) %>%
+      mutate(EIA_sector = substr(MSN, 3, 4)) %>%
```
bpbond: Above two lines: explain what's happening

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
+      mutate(state = State, fuel = GCAM_fuel, sector = GCAM_sector) ->
+      x
+
+    # Create 1 of the 2 ouput tables: narrow years from 1971-2010, convert billion BTU to EJ (fuel specific), remove rows that have no defined sector or fuel name
+    x %>%
+      select(state, sector, fuel, year, value) %>%
+      filter(year %in% GCAM_year) %>%
+      filter(fuel != "NA", sector != "NA") %>%
+      left_join(A_fuel_conv, by="fuel") %>%
+      mutate(value = value*conv_Bbtu_EJ) %>%
```
bpbond: In general please leave space around mathematical operators `*`, `=`, etc.

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
+      spread(year, value) %>%
+      mutate(`1979` =if_else(!is.na(`1979`),`1979`,`1980`)) %>%
+      mutate(`1978` =if_else(!is.na(`1978`),`1978`,`1979`)) %>%
+      mutate(`1977` =if_else(!is.na(`1977`),`1977`,`1978`)) %>%
+      mutate(`1976` =if_else(!is.na(`1976`),`1976`,`1977`)) %>%
+      mutate(`1975` =if_else(!is.na(`1975`),`1975`,`1976`)) %>%
+      mutate(`1974` =if_else(!is.na(`1974`),`1974`,`1975`)) %>%
+      mutate(`1973` =if_else(!is.na(`1973`),`1973`,`1974`)) %>%
+      mutate(`1972` =if_else(!is.na(`1972`),`1972`,`1973`)) %>%
+      mutate(`1971` =if_else(!is.na(`1971`),`1971`,`1972`)) %>%
```
bpbond: Probably would be better to use `tidyr::fill` for this.

inst/extdata/gcam-usa/A_fuel_conv.csv
```diff
@@ -0,0 +1,18 @@
+# File: A_fuel_conv.csv,
+# Title: Fuel specific conversions from Billion BTU to EJ,
+# Units: EJ divided by Billion BTU,
```
bpbond: Source of data?

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
-    # } else {
-    #   ... new code with a fix
-    # }
-    #
-    #
-    # NOTE: there are `merge` and/or 'match' calls in this code. Be careful!
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Merge-and-Match
+
+    # Define vector to be used for filtering
+    GCAM_year <- 1971:2010
```
rplzzz: Is this meant to be the same as `HISTORICAL_YEARS`?  If so, then you should use that constant instead.  If not, then you should explain where this range comes from (i.e., is it determined by the data covered in one or more of the input files or something of that sort?)

R/zchunk_LA101.EIA_SEDS.R
```diff
@@ -34,48 +36,68 @@ module_gcam.usa_LA101.EIA_SEDS_DISABLED <- function(command, ...) {
...
+      mutate(`1979` =if_else(!is.na(`1979`),`1979`,`1980`)) %>%
+      mutate(`1978` =if_else(!is.na(`1978`),`1978`,`1979`)) %>%
+      mutate(`1977` =if_else(!is.na(`1977`),`1977`,`1978`)) %>%
+      mutate(`1976` =if_else(!is.na(`1976`),`1976`,`1977`)) %>%
+      mutate(`1975` =if_else(!is.na(`1975`),`1975`,`1976`)) %>%
+      mutate(`1974` =if_else(!is.na(`1974`),`1974`,`1975`)) %>%
+      mutate(`1973` =if_else(!is.na(`1973`),`1973`,`1974`)) %>%
+      mutate(`1972` =if_else(!is.na(`1972`),`1972`,`1973`)) %>%
+      mutate(`1971` =if_else(!is.na(`1971`),`1971`,`1972`)) %>%
+      gather(year, value, 8:59) %>%
```
rplzzz: Here is another set of hard-wired column numbers.

bpbond: Thanks @rplzzz

bpbond:
This is throwing an error now, and it seemed to be the old #272 - the test code was throwing a read error related to int-vs-double. But when I added `FLAG_PROTECT_FLOAT` to the relevant data frame, it doesn't seem like it's correctly protecting the floats when I look at the output frame (?). @rplzzz if you get a chance could you look at this please? Thanks.

rplzzz:
@bpbond, Which data frame seemed to be the problem?

rplzzz:
The problem appears to be in `add_flags`, which saves the flags as a single string:
```R
add_flags <- function(x, ...) {
  flags <- paste(..., collapse = " ")
  attr(x, "flags") <- c(attr(x, "flags"), flags)
  x
}
```
However, `save_chunkdata` tests flags as if they were character vectors.  It's not just the float protect flag where we make this mistake; here's an example from another section:
```R
      if(FLAG_NO_OUTPUT %in% flags ||
         FLAG_INPUT_DATA %in% flags && !write_inputs) {
        next
      }
```
These tests work when there is only one flag on a table (as was the case for the FAO tables -- though now I'm not sure why they didn't have more), but they won't work with multiple flags.

So, either we need to change the flag attribute to a character vector of flags, or we need to make sure that all tests are written like the ones in the test code:
```R
    if(grepl(FLAG_NO_TEST, new_firstline)) {
      next
    }

    flag_long_year_form <- grepl(FLAG_LONG_YEAR_FORM, new_firstline)
    flag_no_xyear_form <- grepl(FLAG_NO_XYEAR, new_firstline)
    flag_sum_test <- grepl(FLAG_SUM_TEST, new_firstline)
```
My recommendation would be to use a character vector rather than a flat string, though I suppose that requires a split on read-in.  Note that whichever way we decide to go, some of the flag tests will have to be adjusted, as `grepl` does not provide the single logical value we are looking for.

bpbond:
Couldn't we just simplify the code above to this? The problem is happening because the flags are getting pasted together, right?

```R
add_flags <- function(x, ...) {
  attr(x, "flags") <- c(attr(x, "flags"), ...)
  x
}
```

aarony2j:
@bpbond noting that we confirmed that the chunk is producing matching results, but there is a problem with the writing process

bpbond:
Thanks @aarony2j . Right, the chunk's results are correct, but are being read back in by the testing system with numerical discrepancies. I'll track down what's happening, just not for a few days.

bpbond:
@rplzzz I'll get to this later this week, but if you're interested and have time - it looks like some large integers are not getting written correctly here? I.e. Aaron's chunk is producing 100% correct results, but failing checks because (it seems) when the test code reads in the outputs, the largest integers become slightly wrong.

rplzzz:
Thats weird.  They have integer type?  Even 32-bit integers should be good up to about 2 billion or so.

bpbond:
@rplzzz yeah it can't be that exactly. Not sure. I will look tomorrow.

bpbond:
This is holding (and has been holding for quite a while, hint, hint) on a review or delegation from @skim301 or @mwisepnnl

bpbond:
@aarony2j - I took a careful look at this and no, the data really don't match: there are different number of NA values in the `EIA_fuel` column, and that's why the test is failing. So back to you.

Note that I have pushed up some changes: moved files into correct folders (they shouldn't have been in `temp-data-inject/`) and in R you test for NA using `is.na`, not comparing to the string "NA". 

Anyway, take a look. If you can't get to this today, let's just close the PR and re-open it when things are fixed. Thanks, B

bpbond:
Finally @skim301 @mwisepnnl this PR has been sitting here for 10 days with no review. It's a very short piece of code; could one of you _please_ get to it?

aarony2j:
@bpbond Hi Ben. I can't identify anything amiss other than the issue with writing the tables.

You said there are a different number of NA values for EIA_fuel, but I counted both in R and both were 2652. (Note that this is actually a real code for EIA_fuel...not "not available").

Again, I summed both value columns in R and they were in agreement, but not when written. When checking in excel, every mismatch occurred in numbers greater than 10 million, but there were a couple above that that matched.

I don't know how to troubleshoot any further, and I'm afraid I need your or someone's help at this point.

rplzzz:
Can you give some examples of values that were not written out correctly (i.e., what was the value as seen in the code, and what was the value written to the file.  BTW, when you do this, _don't_ pass either value through some intermediary program like Excel.  Excel does its own number mangling, and as a result you can't be sure whether or not the values really matched before Excel got its mitts on them.

bpbond:
>when you do this, don't pass either value through some intermediary program like Excel. Excel does its own number mangling, and as a result you can't be sure whether or not the values really matched before Excel got its mitts on them.

What @rplzzz said. Please don't use Excel for this kind of thing.

bpbond:
OK, I think I see what the problem is. There's a `EIA_fuel` value "NA" (the string NA), no idea what it's supposed to stand for, but it comes from the MSN code "NAICB".

When the testing system reads the old data frame in for comparison, however, I bet this is getting transformed into `NA`--R's missing value. There are 2652 "NA" strings in the new dataset, and 2652 `NA` values in the old.

bpbond:
Let's look at one particular row of your failing data object:
```R
> x[["L101.EIA_use_all_Bbtu"]][429212,]
# A tibble: 1 × 9
  Data_Status state   MSN  year    value EIA_fuel EIA_sector sector  fuel
        <chr> <chr> <chr> <int>    <dbl>    <chr>      <chr>  <chr> <chr>
1       2011F    TX FFTCB  2008 10734755       FF         TC   <NA>  <NA>
```

Now read in the disk version:
```R
> newdata <- readr::read_csv("outputs/L101.EIA_use_all_Bbtu.csv", comment="#", skip=1)
> newdata[429212,]
# A tibble: 1 × 9
  Data_Status state   MSN  year    value EIA_fuel EIA_sector sector  fuel
        <chr> <chr> <chr> <int>    <dbl>    <chr>      <chr>  <chr> <chr>
1       2011F    TX FFTCB  2008 10734760       FF         TC   <NA>  <NA>
```

Aaron's question is - **why is the value different?** In-memory, it matches the old data system value, but after being written and re-read from disk, it differs.

rplzzz:
10734760 is what you get if you round 10734755 to 7 significant figures.

bpbond:
Yeah, realized that.

L101.EIA_use_all_Bbtu needs `FLAG_PROTECT_FLOAT` because otherwise we get parsing errors (thanks so much readr team for your [poor decision](https://github.com/tidyverse/readr/issues/645)). So this line is getting written to disk as
>2011F,TX,FFTCB,2008,1.073476e+07,FF,TC,NA,NA

bpbond:
@gokuliyer it would be great to get this PR merged soon. You were assigned as a reviewer on this; anything to add, any questions, any concerns?

bpbond:
I'm going to assume that after two weeks you would have reviewed/commented if you'd wanted to, @gokuliyer - feel free to weigh in and @aarony2j can submit a patching PR if necessary. Re this one, merging and closing.

gokuliyer:
I had two comments. First, please use more meaningful variable names other than x and y. Please include more comments to justify the filtering of 1971:2011.

bpbond:
@aarony2j at your convenience please open a PR to address Gokul's comments above.

Hi @gokuliyer normally you can just enter comments on the PR webpage. Sorry if that wasn't clear before. I'm happy to demonstrate if useful.

aarony2j:
I'll create a new branch (and pull request) and make these changes.

gokuliyer:
@bpbond : I am confused - Is my review on this required on the LA101.EIA_SEDS file any more?

bpbond:
HI @gokuliyer - no, we're good--it's been merged. Thanks!

gokuliyer:
@bpbond : Thanks!

PR #296, opened 2017-04-26
Add function left_join_keep_first_only()

rplzzz:
This function replicates the behavior of the (deprecated) vecpaste-and-match idiom from the old data system.  Note, by the way, that the suggested replacement currently in our wiki doesn't work as intended.


R/pipeline-helpers.R
```diff
@@ -35,6 +35,36 @@ left_join_error_no_match <- function(d, ...) {
...
+#' @param by Vector of id columns.  Unlike in other join variants, these must be
+#' supplied explicitly.
+#' @return Joined table.  In case of multiple matches, only the first will be
+#' included.
+left_join_keep_first_only <- function(x, y, by) {
+    ## Our strategy is to use "distinct" to filter y to a single element for
+    ## each match category, then join that to x.
+    do.call(distinct_, c(list(y), as.list(by), list(.keep_all = TRUE))) %>%
+      left_join(x, ., by = by)
+}
```
bpbond: That's slick. Nice. Should we provide a version that embodies `left_join_error_no_match`, i.e. guarantees no NAs and no changes in data set size?

rplzzz: Yeah, I was thinking the same thing.  I think it's pretty easy to do as options to one or both of the wrapper functions we've got.  Let's kick around some ideas and decide how we want to do it.

d3y419:
i just tested this and it WORKS!

R/pipeline-helpers.R
rplzzz: @bpbond Should we just go with this as-is for now, and consolidate later?  It looks like Haewon needs to use this function in the chunk he's working on.

bpbond: Yeah, that's fine for now.

By the way, why did you say the wiki example doesn't work? It seems to in my testing, although admittedly I haven't done so exhaustively.

rplzzz: It's because `y %>% right_join(x)` is not exactly a synonym for `y %>% left_join(x, .)`.  The latter is what we want.

PR #302, opened 2017-04-28
Ensure no line breaks in object comments

bpbond:
Addresses #289 . This commit is dedicated to @cahartin


cahartin:
thanks @bpbond !

Issue #275, opened 2017-04-21
LB122.LC_R_Cropland_Yh_GLU


bpbond:
Hi Abigail, this is long, but looks straightforward, and all data are available. Thank you! 

cc @kvcalvin

Issue #117, opened 2017-03-27
LA102.ag_LDS_R_C_GLU


bpbond:

bpbond:
Files are available. Looks easy. Thanks!

abigailsnyder:
@kvcalvin @bpbond 
I'm failing tests due to rounding causing mismatches (I think). I investigated by 
1) reading in the DSR output
2) reading in the old gcam output
3) using round_df from the test_oldnew script with digits=3
4)  investigating where New$value-Old$value != 0

What I'm finding is very strange: 
Old DS rounds 10.8685 to 10.868
New DS rounds 10.8685 to 10.869

All my mismatches are exactly the same - the value has is x.xxx5 and for the Old DS, round_df rounds down and for the New DS, it rounds up. 

When I compare the full, unrounded values, the largest magnitude difference is 10^-14.

I pushed my branch so you can see. I included a script "RoundingDifferences.R" that walks through the above. It's completely standalone as long as your have the old and new outputs.

bpbond:
Hi @abigailsnyder - I've run into this too, and it's super frustrating; I don't understand why it's failing either. What I do have is a workaround. If you've verified that the outputs are all the same (without reasonable precision) and it's rounding weirdness that tripping things up, try adding `FLAG_SUM_TEST` to the failing output(s). This bypasses the normal test and verifies simply that all numeric columns sum to the same total, so it's less stringent.

abigailsnyder:
Hi @bpbond - I tried `FLAG_SUM_TEST` and it gave the error:
1. Error: matches old data system output (@test_oldnew.R#90) -------------------
only defined on a data frame with all numeric variables
1: expect_equivalent(sum(olddata[numeric_columns_old]), sum(newdata[numeric_columns_new]), 
       label = paste(basename(newf), "doesn't match (sum test)")) at C:\Users\snyd535\Documents\GitHub\gcamdata/tests/testthat/test_oldnew.R:90
2: compare(object, expected, check.attributes = FALSE)
3: compare.numeric(object, expected, check.attributes = FALSE)
4: all.equal(x, y, tolerance = tolerance, ...)
5: all.equal.numeric(x, y, tolerance = tolerance, ...)
6: data.class(current)
7: Summary.data.frame(structure(list(), .Names = character(0), row.names = c(NA, -3605L
   ), class = c("tbl_df", "tbl", "data.frame")), na.rm = FALSE)
8: lapply(args, function(x) {
[...]

PR #306, opened 2017-04-28
Snyder lb122 lc r cropland yh glu

abigailsnyder:
PR so Ben can see L122.LC_bm2_R_HarvCropLand_Yh_GLU numeric-matrix mismatch following test. 

#275


abigailsnyder:
[L122.LC_bm2_R_ExtraCropLand_Yh_GLU.pdf](https://github.com/JGCRI/gcamdata/files/964906/L122.LC_bm2_R_ExtraCropLand_Yh_GLU.pdf)
[L122.LC_bm2_R_HarvCropLand_C_Yh_GLU.pdf](https://github.com/JGCRI/gcamdata/files/964907/L122.LC_bm2_R_HarvCropLand_C_Yh_GLU.pdf)
[L122.LC_bm2_R_HarvCropLand_Yh_GLU.pdf](https://github.com/JGCRI/gcamdata/files/964909/L122.LC_bm2_R_HarvCropLand_Yh_GLU.pdf)
[L122.LC_bm2_R_OtherArableLand_Yh_GLU.pdf](https://github.com/JGCRI/gcamdata/files/964908/L122.LC_bm2_R_OtherArableLand_Yh_GLU.pdf)

bpbond:
@abigailsnyder you're getting burned by a problem we're discussing in #290 - an old friend. Hang on, let me try to test and commit a fix that will help.

abigailsnyder:
@bpbond  - an interesting point: giving L122.LC_bm2_R_HarvCropLand_Yh_GLU FLAG_SUM_TEST fixes the numeric-matrix mismatch.

Now I just have to get it to accept FLAG_SUM_TEST flags for  L122.LC_bm2_R_HarvCropLand_C_Yh_GLU and  L122.LC_bm2_R_ExtraCropLand_Yh_GLU. When I give it to them, the test complains that FLAG_SUM_TEST is only for numeric data frames.

bpbond:
@abigailsnyder - OK, everything passes except for one output, which fails the sum test by 1e-5 or something. Investigate, but we may end up using FLAG_NO_TEST if absolutely necessary.

I've added FLAG_PROTECT_FLOAT to several of your outputs, because they had >1000 'integer-ish' values before a double, and this was screwing up `readr::read_csv`. For extensive discussion of this, if you're interested, see here: https://github.com/tidyverse/readr/issues/645

abigailsnyder:
Thank you!

bpbond:
Your code is beautiful and clean overall. If you don't mind, since I already have it local, I'll make a few minor stylistic changes just to save you the trouble?

abigailsnyder:
@bpbond no complaints from me

R/zchunk_LB122.LC_R_Cropland_Yh_GLU.R
```diff
@@ -43,91 +43,609 @@ module_aglu_LB122.LC_R_Cropland_Yh_GLU_DISABLED <- function(command, ...) {
...
+      # and sum:
+      summarise(value = sum(value)) ->
+      L122.ag_HA_bm2_R_Y_GLU
+
+
+    # Lines 51-64 in original file
+    # old comment: Calculate the average cropland, fallow land, and land in temporary crops from FAO RESOURCESTAT
+    # old comment: The time series is unreliable, so only using the last available year (and applying to all historical years)
+    # based on above old comment, get the last available year and set as fallowland_year:
+    fallowland_year <- AGLU_HISTORICAL_YEARS[length(AGLU_HISTORICAL_YEARS)]
```
bpbond: `fallowland <- dplyr::last(AGLU_HISTORICAL_YEARS)` might be clearer

abigailsnyder:
@bpbond - I figured out the failing sum test. It's a result of the data system truncating the outputs to 6 decimal places when they are written to csv. 
 I forced that data frame to do at least 12 digits and the test is now failing by 1e-11 instead of 1e-5. Enough to pass the tests.

That may or may not end up being relevant to other chunks down the line. I'm relatively sure that forcing extra digits to be printed will NOT prevent us from having to use SUM_TEST in the event of the weird rounding up and down we've seen (a la https://github.com/JGCRI/gcamdata/issues/117) since that occurs with bad rounding to the 3rd decimal place.

abigailsnyder:
@bpbond also, thank you for doing the downstream stitching before I got to it!

bpbond:
>I forced that data frame to do at least 12 digits and the test is now failing by 1e-11 instead of 1e-5. Enough to pass the tests.

Cool! How did you force it?

Sorry about the `temp-data-inject` screwup on my end.

abigailsnyder:
`mutate(value = format(value, digits = 12))`

abigailsnyder:
@bpbond , @kvcalvin , @pkyle  I _think_ it's ready to be properly reviewed, with particular attention paid to accuracy and thoroughness of comment information. It's more than a little lengthy.

pkyle:
@abigailsnyder I'm modifying comments; do you want for me to just commit/push them into this pull request, or should I email you my modifications so you can review them first? I'm fine either way.

bpbond:
This is holding on review by @kvcalvin

abigailsnyder:
@pkyle  sorry for the delay in response - it's fine for you to just commit the comments.

R/zchunk_LB122.LC_R_Cropland_Yh_GLU.R
```diff
@@ -43,91 +47,603 @@ module_aglu_LB122.LC_R_Cropland_Yh_GLU_DISABLED <- function(command, ...) {
...
+      select(-Land_Type) ->
+      L122.LC_bm2_R_HarvCropLand_C_Yh_GLU
+
+
+    # Force-Convert years and GCAM_region_IDs in some problematic outputs to integers:
+    L122.LC_bm2_R_ExtraCropLand_Yh_GLU %>%
+      ungroup() %>%
+      mutate(year = as.integer(year),
+             GCAM_region_ID = as.integer(GCAM_region_ID),
+             value = format(value, digits = 12)) ->
```
rplzzz: This statement causes `value` to be converted into a character, which is going to cause problems downstream.  Any subsequent processing that tries to use this data is going to have to convert it back to numeric.

The right thing to do here is add `FLAG_PROTECT_FLOAT` to this table, which will cause this character conversion to be done just before write-out.  If the current number of digits in that character conversion isn't enough; we can always increase it.

abigailsnyder: @rplzzz which test script does `FLAG_PROTECT_FLOAT` set digits in? I can adjust it and push the changes.

bpbond: That's weird because the default number of digits is already quite high--6 digits I think. That should be plenty to pass the tests.

abigailsnyder: @bpbond - printing 6 digits leads to column sum differences of 1e-5 (fails `SUM_TEST`); printing 10 digits leads to column sum differences of 1e-9. printing 12 digits leads to column sum differences of 1e-11 (passes `SUM_TEST`).  So we should set the digits on `FLAG_PROTECT_FLOAT` to be whatever the minimum tolerance needed to pass the `SUM_TEST`

R/zchunk_LB131.LV_R_GLU.R
```diff
@@ -28,11 +28,11 @@ module_aglu_LB131.LV_R_GLU <- function(command, ...) {
...
-    get_data(all_data, "temp-data-inject/L122.LC_bm2_R_HarvCropLand_Yh_GLU") %>%
-      # Two temporary lines for data injection
-      gather(year, value, -GCAM_region_ID, -Land_Type, -GLU) %>%
-      mutate(year = as.integer(substr(year, 2, 5))) ->
-      L122.LC_bm2_R_HarvCropLand_Yh_GLU
+    L122.LC_bm2_R_HarvCropLand_Yh_GLU <- get_data(all_data, "L122.LC_bm2_R_HarvCropLand_Yh_GLU") # %>%
+      # # Two temporary lines for data injection
+      # gather(year, value, -GCAM_region_ID, -Land_Type, -GLU) %>%
+      # mutate(year = as.integer(substr(year, 2, 5))) ->
+      # L122.LC_bm2_R_HarvCropLand_Yh_GLU
```
kvcalvin: Do we need to keep these? Or can we delete them?

abigailsnyder: I think they can be deleted but I didn't want to just assume and do it since I was just adjusting someone else's downstream chunk.

R/zchunk_LB122.LC_R_Cropland_Yh_GLU.R
abigailsnyder: @bpbond @rplzzz weirdly, I just removed the `value=format(value...)` line and somehow it magically passes Test Package and Check Package. 
When I tried Test Package previously (before adding the line), this output failed the `SUM_TEST` (as we discuss above; also scatter plot of old vs new showing lie in straight line above).  
I'm going to wait on Page to push his comments and then I'll remove the line and double check again.

abigailsnyder:
@rplzzz  @bpbond: I've done some testing now that Page's updates to the comments are in. 
Removing the problematic line highlighted by RPL finds the following:
1) if I build & reload, driver, test package:  the test fails because, as BBL noted above, the sum test is failed by 1e-5 - this is what the line was initially included to fix. Between the plot of old vs new points and the way the difference in colsums decreases as I force more printing, I'm quite confident it's just this truncation when the new DS writes outputs.
2) if I build & reload, driver, check package: everything is fine. So that's weird. 

Regardless, we have a few options:
1) leave the line in. As RPL noted, this will cause problems downstream since this table now prints with values as characters and not numeric. Except,  I believe, this table (ExtraCropLand) never gets used downstream, correct @pkyle?
2) remove the offending line and adjust the default number of digits for `FLAG_PROTECT_FLOAT` to be  at least the tolerance needed by the `SUM_TEST`. This might be helpful down the road for other chunks.
3)  remove the offending line and flag this output with `FLAG_NO_TEST`.  I don't like this, but I believe the output is actually correct and if this output ultimately is never used downstream, it may be a non-issue.


Let me know what you think the best way to proceed is, and what you would like me to do.

pkyle:
@abigailsnyder I misspoke when I said I thought that the ExtraCropLand wasn't used later on--it is used, in LB124.LC_R_UnMgd_Yh_GLU.R.

abigailsnyder:
Ok, jk, we will have to actually do something about this then. thanks @pkyle

bpbond:
>remove the offending line and adjust the default number of digits for FLAG_PROTECT_FLOAT to be at least the tolerance needed by the SUM_TEST. This might be helpful down the road for other chunks.

Let's try this first, before going to `FLAG_NO_TEST`?

Relevant code to tweak is in `utils.R`, the `protect_float` function.

abigailsnyder:
@rplzzz @bpbond fixed

rplzzz:
Looks good.  Maybe sometime we should revisit the implementation of protect_float.  Right now it works, but it's kind of derpful (e.g, why did I try to reproduce %g-style formatting instead of just doing everything in %e?).

bpbond:
Kudos and thanks @abigailsnyder !

PR #308, opened 2017-04-28
Page - LA116.geo

pkyle:
geothermal supply curves (energy module LA116)

#174


R/zchunk_LA116.geo.R
```diff
@@ -1,23 +1,27 @@
...
 #' @importFrom tidyr gather spread
-#' @author YourInitials CurrentMonthName 2017
-#' @export
-module_energy_LA116.geo_DISABLED <- function(command, ...) {
+#' @author GPK April 2017
+module_energy_LA116.geo <- function(command, ...) {
   if(command == driver.DECLARE_INPUTS) {
     return(c(FILE = "common/iso_GCAM_regID",
-             "L100.Land_type_area_ha",
+             FILE = "aglu/LDS/Land_type_area_ha",
```
bpbond: Just to double-check, this is _not_ an output of some L100 script, but is instead an external file?

R/zchunk_LA116.geo.R
```diff
@@ -29,57 +33,100 @@ module_energy_LA116.geo_DISABLED <- function(command, ...) {
...
-    # There's also a `same_precursors_as(x)` you can use
-    # If no precursors (very rare) don't call `add_precursor` at all
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    if( nrow(unique(A16.geo_curves[ c("grade", "extractioncost")])) > length( unique( A16.geo_curves$grade ) ) ){
+      stop( "The geothermal (hydrothermal) supply curves have regionally differentiated price points" )
+    }
```
bpbond: Remove GCAM-style spacing around parentheses and brackets please

R/zchunk_LA116.geo.R
```diff
@@ -29,57 +33,100 @@ module_energy_LA116.geo_DISABLED <- function(command, ...) {
...
+    if( nrow(unique(A16.EGS_curves[ c("grade", "extractioncost")])) > length( unique( A16.EGS_curves$grade ) ) ){
```
bpbond: FYI there's also `dplyr::n_distinct` which is faster and clearer than `length(unique(`

R/zchunk_LA116.geo.R
```diff
@@ -29,57 +33,100 @@ module_energy_LA116.geo_DISABLED <- function(command, ...) {
...
+
+    # Aggregate country-level hydrothermal geothermal resource supply curves by GCAM region
+    L116.geothermal_ctry %>%
+      group_by_(GCAM_REGION_ID, "resource", "subresource", "grade", "extractioncost") %>%
+      summarise(available = sum(available)) ->
+      L116.geothermal_rgn
+
+    # Specify the names of the resource table that will be written out
+    L116.table.names <- c( GCAM_REGION_ID, "resource", "subresource", "grade", "available", "extractioncost" )
+    as_tibble( L116.geothermal_rgn[ L116.table.names ]) %>%
```
bpbond: Why not just `L116_geothermal_rgn %>% select(GCAM_REGION_ID, resource, sub resource, grade, available, extractioncost) %>%` ...seems clearer?

R/zchunk_LA116.geo.R
```diff
@@ -29,57 +33,100 @@ module_energy_LA116.geo_DISABLED <- function(command, ...) {
...
+      mutate(available = share * available) ->
+      L116.EGS_ctry
+
+    # Aggregate country-level EGS geothermal resource supply curves by GCAM region
+    L116.EGS_ctry %>%
+      group_by_(GCAM_REGION_ID, "resource", "subresource", "grade", "extractioncost") %>%
+      summarise(available = sum(available)) ->
+      L116.EGS_rgn
+
+    as_tibble( L116.EGS_rgn[ L116.table.names ]) %>%
```
bpbond: Same as comment above

R/zchunk_LA116.geo.R
pkyle: That's correct. It's the output of Alan's Land Data System but it's just a versioned file treated as data from an external source (the LDS weighs in at something like 40 GB so it's not going to be part of the data system)

pkyle: BTW normally we would use the output of the L100 script, because the data are cleaned a bit, but there are several reasons why that is not done here. For one, in the old data system, I didn't want for the LA116 energy file to be dependent on a level1 aglu file, because that would make for more back-and-forth in the hard-wired dependencies set in the makefile (i.e., would need to run LA100 ag before running LA116 energy, which currently isn't the file run order). Two, one of the steps in the LA100 aglu code chunk that pre-processes the LDS data is to map Taiwan back to China, as Taiwan is not an aglu region. However, it is a region in the energy system, so it's actually preferable to keep it separate, as it is when written out by the LDS. And third, this is the way to replicate the data in the old data system.

bpbond: Maybe add that (above) as a comment in this code? Might be useful to know in the future.

Issue #315, opened 2017-04-28
LB165.ag_water_R_C_Y_GLU_irr


bpbond:

bpbond:
@pkyle :

* Title and units for data in `Mekonnen_Hoekstra_Rep47_A2.csv`?
* Title and units for `Rohwer_2007_IrrigationEff.csv`?

Thanks.

pkyle:
Mekonnen
Title: Water footprint per ton of crop or derived crop product at national and sub-national level (m3/ton) (1996-2005)
Units: Cubic meters of water per tonne of crop. [Note that some might consider this a unitless IO coef.]
Rowher
Title: Annex B Irrigation Functional Types, assignment and parameters
Units: Unitless efficiency; large.scale.frac in percentage

PR #329, opened 2017-05-02
Ben - LB165.ag_water_R_C_Y_GLU_irr

bpbond:
#315


R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+      L165.ag_Water_ctry_MHcrop_GLU
+
+    # Work through calculation sequence to determine blue and green water coefficients for irrigated
+    # and rainfed production for gridded crops (original lines 108-122). The sequence has 5 steps:
+    #	(1) Calculate the total biophysical (green + blue) water quantities and coefficients for each country, GLU, and crop
+    #	(2) Match in the irrigated production and calculate the blue water coef for irrigated as min(blue water quantity / irr prod, total biophysical coef)
+    #	(3) Calculate the green water coef for irrigated crops as the total biophysical coef minus the blue water coef
+    #	(4) Calculate the total volume of green water on rainfed crops, as the total green water minus green water used by irrigated crops
+    #	(5) Calculate the green water coef for rainfed crops as rainfed green water volume divided by rainfed production
+    L165.ag_Water_ctry_MHcrop_GLU %>%
```
abigailsnyder: It would be helpful to know which of the 5 sequence steps is being carried out at each point. It's easy to trace through with the comments, but it would probably save an external user time to have them explicitly labeled.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+      # Set the coefs to zero wherever irrigated production is zero (these couldn't have any water consumption in GCAM anyway)
+      mutate(BlueIrr_m3kg = if_else(irrProd_t == 0, 0, BlueIrr_m3kg),
+             # Cap blue water coefficients at the total biophysical quantity
+             BlueIrr_m3kg = pmin(BlueIrr_m3kg, total_m3kg),
+             # Green water coefs of irrigated crops = total biophysical coef - calculated blue water coef
+             GreenIrr_m3kg = total_m3kg - BlueIrr_m3kg,
+             # Green water quantities for rainfed crops = total green water - green water on irrigated
+             GreenIrr_thousm3 = GreenIrr_m3kg * irrProd_t,
+             GreenRfd_thousm3 = green_thousm3 - GreenIrr_thousm3) %>%
+      left_join_error_no_match(L151.ag_rfdProd_t_ctry_crop, by = c("iso", "GTAP_crop", "GLU")) %>%
```
abigailsnyder: Comment explicitly stating that are joining in rfd crop production values to calculate the green water coefficient for rfd crops would probably be helpful to external user.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+      L165.ag_Water_ctry_crop_GLU
+
+    # Original lines 243-265
+    # NOTE: Fodder crops are not in the M+H inventories. They instead have been assigned a
+    # corresponding crop that is in the M+H inventories and at this point are multiplied by
+    # an exogenous fodder:crop adjustment factor. This factor is designed to reflect that
+    # fodder crop masses include the whole plant, often harvested green (~70% water). As such,
+    # their water coefficients will tend to be a good deal lower than the corresponding crop
+    # to which they are assigned. The specific number is based on analysis of a virtual water
+    # content dataset (Chapagain 2004) that included fodder crops.
```
abigailsnyder: not really a review comment so much as a thought: does it make sense for us to essentially write a reference list for the data system? For example, Chapagain 2004 and the context is probably enough to find that source, but maybe we should compile everything in one place.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+      left_join_error_no_match(select(FAO_ag_items_PRODSTAT, MH_crop, GTAP_crop), by = "MH_crop") ->
+      L165.Water_footprint_m3
+
+    # Re-cast nation-level data so that blue and green are represented as data columns (original lines 59-64)
+    Mekonnen_Hoekstra_Rep47_A2 %>%
+      gather(iso, coef_m3t, -FAO_crop, -water_type) %>%
+      mutate(coef_m3kg = coef_m3t / CONV_T_KG) %>%
+      select(iso, FAO_crop, water_type, coef_m3kg) %>%
+      # Set all missing values to zero -- this may be re-visited at another point
+      # We may want to substitute default values from the MH2011 table
```
kvcalvin: This pair of comments is a little confusing. Do we revisit at another point? Or do you think we should (in which case we should either open an issue or use OLD_DATA_SYSTEM_BEHAVIOR)?

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+      mutate(Prod_t = if_else(is.na(Prod_t), 0L, Prod_t)) %>%
+      # For the 18 M+H gridded crops, calculate the coefs by country, GLU, and crop (original lines 79-82)
+      # NOTE: dropping all country / GLU / crops with zero production; these won't have any water anyway
+      filter(Prod_t > 0) %>%
+      mutate(coef_m3kg = value / (Prod_t * CONV_T_KG)) ->
+      L165.ag_Water_ctry_MHcrop_GLU
+
+    # Clip extremely high values which are due to statistical discrepancies (original lines 84-106)
+    # The assumed maximum values are crop specific and based on analysis of the 2011 inventory data
+    # In this step, we're adding 2 m3/kg to every maximum observed value across all national averages.
```
kvcalvin: Can we rephrase this slightly? When I first read it I thought you were adding 2 to the data, not setting the cap at max + 2.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
bpbond: This is probably a question for @pkyle - comment is verbatim from old data system.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
bpbond: Done.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
bpbond: Opened a new issue #345 about this

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
bpbond: Done.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
kvcalvin: Sounds like a good idea, but I'm not sure the best way to do it.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
```diff
@@ -0,0 +1,407 @@
...
+    # NOTE: Fodder crops are not in the M+H inventories. They instead have been assigned a
+    # corresponding crop that is in the M+H inventories and at this point are multiplied by
+    # an exogenous fodder:crop adjustment factor. This factor is designed to reflect that
+    # fodder crop masses include the whole plant, often harvested green (~70% water). As such,
+    # their water coefficients will tend to be a good deal lower than the corresponding crop
+    # to which they are assigned. The specific number is based on analysis of a virtual water
+    # content dataset (Chapagain 2004) that included fodder crops.
+    # The "corresponding" crops are assigned in the mapping file (FAO_ag_items_PRODSTAT),
+    # based on species (e.g., maize) or family (e.g., alfalfa)
+    CONV_CROP_FODDER <- 0.2
```
kvcalvin: Is this only used here? If so, can we use `WATER` somewhere in its name so it is clearer?

R/zchunk_data_Maddison_population.R
```diff
@@ -35,7 +35,7 @@ module_data_Maddison_population <- function(command, ...) {
       add_units("people") %>%
       add_comments(paste("Read from", gsub("^.*extdata", "extdata", fqfn))) %>%
       add_legacy_name("Maddison_population") %>%
-      add_flags(FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR, FLAG_NO_TEST) ->
+      add_flags(FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR, FLAG_NO_OUTPUT) ->
       Maddison_population
```
kvcalvin: ?

bpbond: I realized that Maddison_population not only doesn't need to be tested (it's not a data system output), but doesn't even need to be written to disk. This saves a few microseconds.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
bpbond: `FODDER_CROP_WATER_ADJUSTMENT`?

kvcalvin:
That works.

R/zchunk_LB165.ag_water_R_C_Y_GLU_irr.R
pkyle: I'd recommend just leaving it as a comment for now, and providing some more information. Hard to see this really rising to a priority level but here's the info in case:
The source data table has 29,725 observations (country x crop), of which 24,063 are missing values, and of these missing values, 321 did actually have crop production in Monfreda's inventory, 13 of which had > 1 Mt of production. We are using this dataset to supplement a global gridded inventory with 18 major crops. Filtering those out, there are still 4 observations with between 1 and 3 Mt of production, all in Nigeria: CitrusFrtNES, CowPeasDry, Plantains, and FrtFrshNES.
If we want to replace these with defaults, the blue water coef (i.e., the important one) will still be zero, because it's in Nigeria, where ~100% of crop production is rainfed (according to the inventories). So, if we do implement a different method, we'll get a slightly higher estimate of biophysical water consumption for MiscCrop in Africa_Western.

bpbond: Thanks Page. I'll add to comments. @kvcalvin otherwise ready for merge?

kvcalvin: Wow, @pkyle: much more information than I expected. I just wasn't sure if the "revisit later" meant later in the file.

@bpbond: approving now. merge away.

pkyle: Who knows what I meant when I wrote that!? Just from the info I left, it could have been an important omission, and I doubt I looked into it at the time.

PR #333, opened 2017-05-02
Aaron - LA126.distribution.R

aarony2j:
No errors or warnings during check. However, test had a few failures:
1. @test_oldnew.R#103, round_df(olddata) not equivalent to round_df(newdata) on L123.eff_R_elec_F_Yh.csv (not part of my chunk)
2. @test_oldnew.R#70, length(oldf) == 1 isn't true for every one of my nine output tables


R/zchunk_LA126.distribution.R
```diff
@@ -12,15 +16,14 @@
...
+module_energy_LA126.distribution <- function(command, ...) {
   if(command == driver.DECLARE_INPUTS) {
-    return(c(FILE = "energy/calibrated_techs",
-             FILE = "temp-data-inject/L1011.en_bal_EJ_R_Si_Fi_Yh",
+    return(c(FILE = "temp-data-inject/L1011.en_bal_EJ_R_Si_Fi_Yh",
              FILE = "temp-data-inject/L122.out_EJ_R_gasproc_F_Yh",
-             "L123.out_EJ_R_elec_F_Yh",
-             "L123.out_EJ_R_indchp_F_Yh"))
+             FILE = "L123.out_EJ_R_elec_F_Yh",
+             FILE = "L123.out_EJ_R_indchp_F_Yh"))
```
bpbond: Why did you add `FILE =` to these lines? I don't understand why this even runs.

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +39,166 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
+    # ELECTRICITY OWNUSE, i.e., electricity consumed onsite prior to any transmission and distribution losses
+      # Summing industrial CHP electricity generation and electricity generation by GCAM region ID and year
+    L123.out_EJ_R_indchp_F_Yh %>%
+      group_by(GCAM_region_ID, year) %>%
+      summarise(value_indchp_F_Yh = sum(value)) %>%
+      left_join_error_no_match( # Joining L123.out_EJ_R_elec_F_Yh, but first summing by GCAM region ID and year
+        summarise(
+          group_by(L123.out_EJ_R_elec_F_Yh, GCAM_region_ID, year),
+          value_elec_F_Yh = sum(value)),
+        by = c("GCAM_region_ID", "year")) %>%
```
bpbond: This is some nested gnarliness. Consider breaking out into a separate d.f.

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +39,166 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
+      mutate(value_b = value_elec_F_Yh + value_indchp_F_Yh) %>% # Summing electricity generation and industrial CHP electricity generation. This creates values for table, L126.in_EJ_R_elecownuse_F_Yh (i.e., input).
+      select(-value_elec_F_Yh, -value_indchp_F_Yh) ->
+      a
+
+      # Filtering energy balance by "net_electricity ownuse" sector and joining with previous table
+    L1011.en_bal_EJ_R_Si_Fi_Yh %>%
+      gather(year, value, -GCAM_region_ID, -sector, -fuel) %>%
+      mutate(year = as.integer(substr(year, 2, 5))) %>%
```
bpbond: These two lines should be up above (i.e. easy to find), right after the `get_data` call, as they'll become unnecessary once we stop using the temp data.

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +39,166 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
+      filter(sector == "net_gas pipeline") %>%
+      mutate(sector = substr(sector, 5, 30)) %>% # Renaming sector as "gas pipeline"
+      group_by(GCAM_region_ID, sector, fuel, year) %>%
+      summarise(value_a = sum(value)) %>%
+      left_join( # Joining L122.out_EJ_R_gasproc_F_Yh, but first converting to long form, stripping years of X's, and summing by GCAM region ID and year
+        summarise(
+          group_by(
+            mutate(
+              gather(L122.out_EJ_R_gasproc_F_Yh, year, value, -GCAM_region_ID, -sector, -fuel),
+              year = as.integer(substr(year, 2, 5))),
```
bpbond: Hahaha! No, sorry, I'm putting my foot down with this one. Too much nesting.

R/zchunk_LA126.distribution.R
aarony2j: Just thought I needed to. No good reason.

R/zchunk_LA126.distribution.R
aarony2j: Broke it out

R/zchunk_LA126.distribution.R
aarony2j: done

R/zchunk_LA126.distribution.R
aarony2j: You do not want to see my latest code then! My goal was to write one long stream without a single break. I thought that was the most efficient because it has the fewest lines.

aarony2j:
You mentioned talking about the test data files because my chunk had failed the test process. Thanks. I appreciate it.

bpbond:
@aarony2j - you had accidentally stomped on the master version of `L1011.en_bal_EJ_R_Si_Fi_Yh.csv.zip`, which was triggering the error in `L123.eff_R_elec_F_Yh.csv`. I have fixed this, and tweaked a bit of formatting. All tests now pass. 👍

bpbond:
This is holding (and has been holding for quite a while, hint, hint) on a review or delegation from @skim301 or @mwisepnnl

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +40,180 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
-    #
-    # ===================================================
+    # ELECTRICITY OWNUSE, i.e., electricity consumed onsite prior to any transmission and distribution losses
+    # Summing industrial CHP electricity generation and electricity generation by GCAM region ID and year
+    # Summing electricity generation by GCAM region ID and year
+    L123.out_EJ_R_elec_F_Yh %>%
+      group_by(GCAM_region_ID, year) %>%
+      summarise(value_elec_F_Yh = sum(value)) ->
+      a
+
```
skim301: Please use descriptive names for a, b, c, d, e, and f so that we can better understand the process. For example, "a" could be name electricity.generation and "b" could be named electricity.industry and so on. Also, rather than value_elec_F_Yh, you could use elec_total.

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +40,180 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
+    # Summing industrial CHP electricity generation and then joining it with electricity generation
+    L123.out_EJ_R_indchp_F_Yh %>%
+      group_by(GCAM_region_ID, year) %>%
+      summarise(value_indchp_F_Yh = sum(value)) %>%
+      left_join_error_no_match(a, by = c("GCAM_region_ID", "year")) %>%
+      mutate(value_b = value_elec_F_Yh + value_indchp_F_Yh) %>% # Summing electricity generation and industrial CHP electricity generation. This creates values for table, L126.in_EJ_R_elecownuse_F_Yh (i.e., input).
+      select(-value_elec_F_Yh, -value_indchp_F_Yh) ->
```
skim301: value_b should be a descriptive name such as total_gross_elec or something like that.

R/zchunk_LA126.distribution.R
```diff
@@ -36,123 +40,180 @@ module_energy_LA126.distribution_DISABLED <- function(command, ...) {
...
+      b
+
+    # Filtering energy balance by "net_electricity ownuse" sector and joining with previous table
+    L1011.en_bal_EJ_R_Si_Fi_Yh %>%
+      filter(sector == "net_electricity ownuse") %>%
+      mutate(sector = substr(sector, 5, 30)) %>% # Renaming sector as "electricity ownuse"
+      group_by(GCAM_region_ID, sector, fuel, year) %>%
```
skim301: It's clearer to remove "net_", rather than substr(sector, 5, 30) and also prevents error if length (30) is incorrect.

aarony2j:
Thank you for the comments. I'll remember that for next time.

R/zchunk_LA126.distribution.R
aarony2j: please note that I didn't know how to do the reverse, so I used mutate/replace:
mutate(sector = replace(sector, sector == "net_gas pipeline", "gas pipeline"))

skim301: You could use sub("net_", "", sector), which substitutes the pattern "net_" for a blank but the way you have it is clearer and more explicit.

PR #334, opened 2017-05-02
Stephanie - L100.Population_downscale_ctry

swaldhoff:


R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -55,23 +165,23 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
     # There's also a `same_precursors_as(x)` you can use
     # If no precursors (very rare) don't call `add_precursor` at all
     tibble() %>%
```
d3y419: this just creates an empty tibble on mine.

R/constants.R
```diff
@@ -130,9 +130,11 @@ ELECTRICITY_INPUT_FUELS<- c( "biomass", "coal", "gas", "refined liquids" )
 # and should contain all historical years used by other modules
 socioeconomics.MADDISON_HISTORICAL_YEARS <- seq(1700, 1900, 50) # Years for which to use Maddison data
 socioeconomics.UN_HISTORICAL_YEARS <- c(1950, 1971:2010) # Years for which to use UN data
-
+socioeconomics.AGG_BASE_YEAR <- 1950 # Year used to generate ratios for region -> country downscaling
+socioeconomics.FINAL_HIST_YEAR <- 2010 # Final historical year
 socioeconomics.BASE_POP_SCEN <- "SSP2"
```
d3y419: better to define here:
socioeconomics.pop_model <- "IIASA-WiC POP"

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -34,19 +34,129 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
+    ## (2) SSP population projections by country
+
+    # Final historical population from UN
+    pop_final_hist <- filter(L100.Pop_thous_ctry_Yh, year == socioeconomics.FINAL_HIST_YEAR) %>%
+      rename(pop_final_hist = value) %>%
+      select(-year)
+
+    # First, generate ratios of future population to base year (2010) for all SSPs. The ratios will be applied to the historical year populations so there are no jumps/inconsistencies.
+    L100.Pop_thous_SSP_ctry_Yfut <- as_tibble(SSP_database_v9) %>% # Note units in SSP database are millions, but convert to thousands when we multiply by historic year
+      filter(MODEL == "IIASA-WiC POP" & VARIABLE == "Population") %>%  # IIASA-WiC is the official SSP population data set
```
d3y419: Model can be defined in constants

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -34,19 +34,129 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
+      select(-pop) %>%
+      # Second, project country population values using SSP ratios and final historical year populations.
+      # Not all countries in the UN data are in SSP data. Create complete tibble with all UN countries & SSP years.
+      ungroup() %>%
+      complete(scenario = unique(scenario),
+               year = unique(year),
+               iso = unique(L100.Pop_thous_ctry_Yh$iso)) %>%
+      # For these countries, the ratio will be set to 1 (per the old data system).
+      mutate(ratio_iso_ssp = if_else(is.na(ratio_iso_ssp), 1, ratio_iso_ssp)) %>%
+      ## Note: In the old data system, Taiwan is in this category and has constant population. Issue has been opened to deal with this later. ##
```
d3y419: recommendation:
```
    if(OLD_DATA_SYSTEM_BEHAVIOR) {
### original code here
    } else {
      # First, clean up SSP data
      # taiwan is missing from the main db
      # get taiwan from OECD Env model
      temp <- SSP_database_v9 %>%
        filter(MODEL == "OECD Env-Growth" & VARIABLE == "Population" & REGION == "TWN")
      
      # all other countries from the standard model
      SSP_database_clean <- as_tibble(SSP_database_v9) %>% # Note units in SSP database are millions, but convert to thousands when we multiply by historic year
        filter(MODEL == socioeconomics.pop_model & VARIABLE == "Population") %>%
        bind_rows(temp) %>%
        mutate(REGION = gsub("ROU", "ROM", REGION))   # SSP uses "ROM" for the iso for Romania; replace with "rom" for consistency with other data sources
      
      # generate ratios of future population to base year (2010) for all SSPs. The ratios will be applied to the historical year populations so there are no jumps/inconsistencies.
      L100.Pop_thous_SSP_ctry_Yfut <- SSP_database_clean %>% # Note units in SSP database are millions, but convert to thousands when we multiply by historic year
        mutate(iso = tolower(REGION), scenario = substr(SCENARIO, 1,4)) %>%
        select(-MODEL, -VARIABLE, -UNIT, -REGION, -SCENARIO) %>%
        gather(year, pop, -iso, -scenario) %>%  # Long format
        mutate(year = as.integer(year), pop = as.numeric(pop)) %>%  # Clean year variable
        filter(year %in% c(socioeconomics.FINAL_HIST_YEAR, modeltime.FUTURE_YEARS)) %>% # Retain only years needed for GCAM
        group_by(scenario, iso) %>%
        mutate(ratio_iso_ssp = pop / pop[year == socioeconomics.FINAL_HIST_YEAR]) %>%  # Calculate population ratios to final historical year (2010), no units
        select(-pop) %>%
        # Second, project country population values using SSP ratios and final historical year populations.
        # Not all countries in the UN data are in SSP data. Create complete tibble with all UN countries & SSP years.
        ungroup() %>%
        complete(scenario = unique(scenario),
                 year = unique(year),
                 iso = unique(L100.Pop_thous_ctry_Yh$iso)) %>%
        # For these countries, the ratio will be set to 1 (per the old data system).
        mutate(ratio_iso_ssp = if_else(is.na(ratio_iso_ssp), 1, ratio_iso_ssp)) %>%
        left_join(pop_final_hist, by = "iso") %>% # Join with final historic period population
        mutate(value = pop_final_hist * ratio_iso_ssp) %>%  # Units are 1000 persons (UN 2010 value is in thousands)
        select(-pop_final_hist, -ratio_iso_ssp)
    }  
```

d3y419: note that i also separated the pure clean up of original data vs. ratio processing.
this way, when we get the next-gen SSPs, we only need to change the data clean up part. ratio processing need not change.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -55,23 +165,23 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
+      add_comments("Maddison population data cleaned to develop complete data for all years, (dis)aggregated to modern country boundaries") %>%
+      # add_comments("can be multiple lines") %>%
       add_legacy_name("L100.Pop_thous_ctry_Yh") %>%
-      add_precursors("precursor1", "precursor2", "etc") %>%
+      add_precursors("socioeconomics/socioeconomics_ctry", "Maddison_population") %>%
       # typical flags, but there are others--see `constants.R`
       add_flags(FLAG_NO_TEST, FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR) ->
       L100.Pop_thous_ctry_Yh
 
     tibble() %>%
```
d3y419: change to:
`L100.Pop_thous_SSP_ctry_Yfut`

R/zchunk_L100.Population_downscale_ctry.R
rplzzz: Disagree.  This isn't really a constant; it's extracting a single table from a data structure that contains a whole bunch of tables smooshed together.  Unless there's a reason why we might want to make a coordinated change across several chunks in which table is selected here, there is no reason to clutter `constants.R` with this.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -34,19 +34,129 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
-    # } else {
-    #   ... new code with a fix
-    # }
-    #
+
+    ## (1) Historical population by country
+
+    # First clean up Maddison raw data -- NOTE: Maddison data are used to develop population ratios relative to 1950 to combine with UN data from 1950 onward
+    pop_thous_ctry_reg <- Maddison_population %>%
+      filter(!is.na(Country)) %>%  # Remove values with missing country name (all year values are also missing for these--probably just blank rows in the original data)
```
rplzzz: This really should have been taken care of in the chunk that generated the `Maddison_populaiton` table.  If this is the only place that data is used (i.e., downstream modules use it only indirectly through the outputs of this module), then I'd say let it be.  However, if any other modules are using that table, then we should fix it at its source.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -34,19 +34,129 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
+      rename(Maddison_ctry = Country,
+             pop = value) %>%  # Change name to match socioeconomics_ctry mapping file
+      left_join(socioeconomics_ctry, by = "Maddison_ctry")  # Join with iso codes
+
+    # Second, estimate population values prior to 1950 for countries in aggregate regions. This is what we want: pop_country_t = (pop_aggregate_t / pop_aggregate_1950) * pop_country_1950
+    # Generate a scalar for population in each aggregate region in 1950 (to generate the population ratios)
+    agg_ratio <- pop_thous_ctry_reg %>%
+      select(Maddison_ctry, year, pop) %>%
+      filter(year <= socioeconomics.AGG_BASE_YEAR & (Maddison_ctry == "Total Former USSR" | Maddison_ctry == "Czechoslovakia" | Maddison_ctry == "Yugoslavia")) %>%  # Only want years prior to 1951 for the three regions
```
rplzzz: FYI, you can give filter multiple arguments, and they are implicitly joined with `&`.  So, for example, this would be equivalent and perhaps a bit easier to read:
```
filter(year <= socioeconomics.AGG_BASE_YEAR,
        Maddison_ctry %in% c("Total Former USSR", "Czechoslovakia", "Yugoslavia"))
```
No need to change it here, but it's a useful tool to have in your kit when you're writing these kinds of expressions.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -34,19 +34,129 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
+      group_by(Maddison_ctry) %>%  # Group to perform action on each aggregate region individually
+      mutate(ratio = pop/pop[year == socioeconomics.AGG_BASE_YEAR]) %>%  # Create ratio of population in prior years to population in 1950
```
rplzzz: I'm a little surprised that this works, since `pop` and `pop[year==...]` have different lengths.  I suppose it's because within each grouping the year criterion selects exactly row, and you're allowed to divide a vector by a scalar.

rplzzz:
@swaldhoff This PR is showing conflicts with `inst/extdata/socioeconomics/SSP_database_v9.csv.gz`.  What did you change in this file?

R/zchunk_L100.Population_downscale_ctry.R
swaldhoff: Thanks. Since there are a few other changes, I'll make this change, too. (It is easier to read.)

R/constants.R
swaldhoff: I agree with Robert.

R/zchunk_L100.Population_downscale_ctry.R
swaldhoff: I think Robert is correct--the selection of the model is only used in this chunk.

R/constants.R
```diff
@@ -130,9 +130,11 @@ ELECTRICITY_INPUT_FUELS<- c( "biomass", "coal", "gas", "refined liquids" )
 # and should contain all historical years used by other modules
 socioeconomics.MADDISON_HISTORICAL_YEARS <- seq(1700, 1900, 50) # Years for which to use Maddison data
 socioeconomics.UN_HISTORICAL_YEARS <- c(1950, 1971:2010) # Years for which to use UN data
-
+socioeconomics.AGG_BASE_YEAR <- 1950 # Year used to generate ratios for region -> country downscaling
+socioeconomics.FINAL_HIST_YEAR <- 2010 # Final historical year
```
swaldhoff: Okay. Will fix that.

swaldhoff:
@rplzzz I added the file, title, units, etc., info to the SSP_database file.

rplzzz:
Ok, you're up to date with the master branch, and conflicts are resolved.

R/constants.R
```diff
@@ -130,9 +130,11 @@ ELECTRICITY_INPUT_FUELS<- c( "biomass", "coal", "gas", "refined liquids" )
 # and should contain all historical years used by other modules
 socioeconomics.MADDISON_HISTORICAL_YEARS <- seq(1700, 1900, 50) # Years for which to use Maddison data
 socioeconomics.UN_HISTORICAL_YEARS <- c(1950, 1971:2010) # Years for which to use UN data
```
swaldhoff: I'm not sure how that would work in this case. The SSP values begin in 2010 and, since we don't want a jump in the population values when we move from the UN to the SSP data, the GCAM SSP population values are not exactly the population values in the SSP database, but the ratio of the future SSP values to the 2010 SSP value times the UN 2010 population. If the "future" years started in 2001, how would we estimate the SSP population projections?
I don't know that we can't, I'm just not sure how we would. Also, possibly we wouldn't want to. Since population is always exogenous, the pre-2015 values are all always going to be the same (we're not projecting 2020 based on 2015, that's determined by the scenario, and the historic years are all fixed--they're not something that would change depending on the definition of "historical" and "future" years in the model.
Not sure if I answered your question, though.

inst/extdata/socioeconomics/socioeconomics_ctry.csv
```diff
@@ -1 +1 @@
...
ven,Venezuela,Venezuela,
vnm,Viet Nam,Vietnam,
vgb,"Virgin Islands, British",British Virgin Islands,
vir,"Virgin Islands, U.S.",Virgin Islands,
wlf,Wallis and Futuna,Wallis and Futuna,
esh,Western Sahara,Western Sahara,
yem,Yemen,Yemen,
yug,"Yugoslavia, Federal Republic of",,
zmb,Zambia,Zambia,
zwe,Zimbabwe,Zimbabwe,
```
swaldhoff: I opened the csv file in text edit and R and don't see the "^M" at the end of the lines. How do I fix this?

inst/extdata/socioeconomics/Maddison_population.csv
```diff
@@ -1 +1 @@
...
Uruguay,,,,55,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,132,,,,,,,,,,,,,,,,,,,,343,354,364,376,387,399,411,424,437,450,464,482,502,522,543,564,587,610,635,660,686,706,727,748,770,792,815,839,864,889,915,930,945,961,977,993,1009,1026,1043,1062,1081,1112,1144,1177,1223,1246,1269,1292,1316,1341,1371,1402,1433,1465,1498,1534,1571,1608,1646,1685,1713,1741,1770,1799,1829,1859,1889,1921,1952,1944,1965,1987,2010,2032,2055,2081,2107,2134,2160,2188,2194.275,2223.06,2252.965,2283.92,2316.543,2353.136,2389.29,2424.959,2460.411,2495.022,2530.969,2563.794,2598.466,2632.111,2663.61,2693.133,2721.207,2748.579,2776.607,2801.859,2823.659,2826.042,2829.526,2834.115,2838.213,2841.818,2857.453,2873.52,2889.495,2905.131,2919.851,2935.649,2953.997,2972.57,2990.097,3008.27,3026.971,3045.153,3063.668,3084.065,3105.708,3127.574,3149.262,3170.958,3193.413,3215.423,3238.28,3262.838,3284.853,3304.234,3323.876,3343.797,3363.085,3381.606,3399.237,3415.92,3431.932,3447.496,3462.531,,3721.919
Venezuela,,,,718,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1324,,,,,,,,,,,,,,,,,,,,1653,1675,1699,1725,1753,1784,1816,1849
```
swaldhoff: The L100.Population_downscale chunk now reads in the Maddison population data after it has been processed by zchunk_data_Maddison_population.R. I didn't commit any changes to the csv file.

swaldhoff: (Or at least I don't think I did. I added the file, title, etc., lines on my local repository, but since it turned out that this file was being processed elsewhere, I didn't think I committed the changes to the csv file. If I did, I didn't mean to!)

R/constants.R
swaldhoff: Actually, I think we need to leave it as 2010 (see comment above) because we need to be able to match SSP and UN population data and the SSP data start in 2010. Unless there is ever a new SSP database (in which case this would likely need to be rewritten anyway), 2010 is always going to be this "historical" year for the SSPs (while 2015 is the first future year in the SSPs).

inst/extdata/socioeconomics/socioeconomics_ctry.csv
d3y419: i'm sure Page would love to take credit for this, but we should make this more accessible.

# source:  ISO 3166 standard published by the International Organization for Standardization (ISO)

R/constants.R
rplzzz: I thought something like that might be the case.  All of this makes me wonder if we are really quite certain what is supposed to happen when we tinker with the historical/future years.

R/constants.R
rplzzz: In that case it should be `max(UN_HISTORICAL_YEARS)`.  Either way, a quantity that is semantically "the last item in this (ordered) set" should be computed from the set, not defined separately.

inst/extdata/socioeconomics/socioeconomics_ctry.csv
swaldhoff: It's the matching between the ISO codes and the Maddison country names is from Page, not the ISO codes.

R/constants.R
swaldhoff: It's not because it's the maximum of the UN historical years, but the minimum of the SSP years. So there isn't really another value from which to define it (there isn't a constant with the SSP future years, but we could make one I guess, though there isn't really anywhere to use it).

inst/extdata/socioeconomics/socioeconomics_ctry.csv
d3y419: I'm pretty sure @Page got this mapping from ISO or its derivative.

rplzzz: The point is, "source" is supposed to be the location where you can download the data, if we got it from somewhere else.  If the data was generated internally, then the source tag should indicate as much.  The source tag isn't for describing why we constructed an original data set a certain way.  You can put that in comments, if it's useful to do so.

(No need, however, to put the names of individual developers in there.)

d3y419: I'm fine with:
Source: internally generated based on ISO. 

So long as it doesn't mention individual developers name unless absolutely necessary.

swaldhoff: Changed source to "Internally generated mapping file."

@d3y419 The ISO codes are not internally generated, but this mapping file--between the Maddison country names (which are not standard ISO country names) to the ISO codes--was internally generated and cannot be accessed anywhere else other than here.

R/zchunk_L100.Population_downscale_ctry.R
bpbond: Haha, that ("I'm surprised that this works") is exactly what I said when @swaldhoff showed me this.

R/zchunk_L100.Population_downscale_ctry.R
bpbond: Done. @swaldhoff you can remove line 42.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -55,23 +165,23 @@ module_socioeconomics_L100.Population_downscale_ctry_DISABLED <- function(comman
...
-      add_comments("can be multiple lines") %>%
+      add_title("Population by country, 1700-2010") %>%
+      add_units("thousand") %>%
+      add_comments("Maddison population data cleaned to develop complete data for all years, (dis)aggregated to modern country boundaries") %>%
+      # add_comments("can be multiple lines") %>%
       add_legacy_name("L100.Pop_thous_ctry_Yh") %>%
-      add_precursors("precursor1", "precursor2", "etc") %>%
+      add_precursors("socioeconomics/socioeconomics_ctry", "Maddison_population") %>%
       # typical flags, but there are others--see `constants.R`
       add_flags(FLAG_NO_TEST, FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR) ->
```
bpbond: (This has since been removed from the skeleton-generating script. Not a good idea on my part.)

inst/extdata/socioeconomics/socioeconomics_ctry.csv
bpbond: I've done this.

rplzzz:
What happened to this PR?  Merged or no?

bpbond:
I'm working with @swaldhoff on this right now. Closed this b/c she had to completely redo her branch or something.

Issue #128, opened 2017-03-28
L131.nonco2_proc_R_S_T_Y.R


bpbond:

bpbond:
Check with me before starting--may need data inject.

bpbond:
@cahartin this is already in repo.

cahartin:
@bpbond thanks!

PR #339, opened 2017-05-03
Corinne - L131_nonco2_proc_r_s_t_y

cahartin:
#128


cahartin:
There are 2 files that I'm failing on that aren't mine. 
`1. Failure: matches old data system output (@test_oldnew.R#103) ----------------
Rows in x but not y: 21, 20, 7, 15, 4, 9, 25, 3, 14, 24, 22[...]. Rows in y but not x: 24, 21, 4, 9, 3, 14, 13, 7, 15, 22, 20[...]. 
L102.ghg_tgej_USA_en_Sepa_F_2005.csv doesn't match`

`2. Failure: matches old data system output (@test_oldnew.R#103) ----------------
Rows in x but not y: 1, 5, 2, 3. Rows in y but not x: 5, 1, 2, 3. 
L103.ghg_tgmt_USA_an_Sepa_F_2005.csv doesn't match`

cahartin:
I had to use FLAG_SUM_TEST in order to pass the test.  Here are my results compared to the old system. 
`old<- readr::read_csv("C:/Users/hart428/Documents/GitHub/gcamdata/tests/testthat/comparison_data/emissions/L131.nonco2_tg_R_prc_S_S_Yh.csv", comment = "#")`
`old_long<-gather(old, year, value, -GCAM_region_ID, -supplysector, -subsector,-stub.technology, -Non.CO2)`

`summary(old_long)
 value         
 Min.   : 0.00000  
 1st Qu.: 0.00009  
 Median : 0.00141  
 Mean   : 0.13864  
 3rd Qu.: 0.01831  
 Max.   :34.56765`

`new<- readr::read_csv("C:/Users/hart428/Documents/GitHub/gcamdata/outputs/L131.nonco2_tg_R_prc_S_S_Yh.csv", comment = "#")`

`summary(new)
     value         
 Min.   : 0.00000  
 1st Qu.: 0.00009  
 Median : 0.00141  
 Mean   : 0.13864  
 3rd Qu.: 0.01831  
 Max.   :34.56765 ` 

`combined<-left_join(old_long, new, by = c("GCAM_region_ID", "Non.CO2", "supplysector", "subsector", "stub.technology", "year"))`

`qplot(combined$value.x, combined$value.y) + geom_abline()`

dsr

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+       rename(tech_emissions = tech_emissions.x) %>%
+       mutate(tech_share = tech_emissions/sector_emissions) ->
+       L131.nonco2_pct_R_prc_S_S_2005
+
+     # Fourth: Disaggregate EDGAR emissions to subsectors
+     # First aggregate all EDGAR data and subset for processing sectors
+
+     #filter out years 2009 and 2010 in EDGAR_VOC
+     EDGAR_VOC %>%
+       filter(year!= "2009" & year != "2010") ->
```
bpbond: `year` should be integer, not character. Also WHY are we filtering out these years?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+                                "metals","foams","solvents", "semiconductors")) %>%
+       select(year,value, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+       na.omit()  %>%  # delete rows with NA's
+       group_by(year, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+       summarise(EDGAR_emissions = sum(value)) %>%
+       mutate(EDGAR_emissions = EDGAR_emissions * CONV_GG_TG) -> # convert from gg to tg
+       L131.EDGAR.mlt
+
+     L131.EDGAR.mlt$GCAM_region_ID <- as.integer(L131.EDGAR.mlt$GCAM_region_ID)
+     L131.EDGAR.mlt$year <- as.integer(L131.EDGAR.mlt$year)
```
bpbond: Ah ha! Why not include 151-152 in pipeline above?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+       na.omit() %>%  # delete rows with NAs
+       group_by(sector, subsector, Non.CO2) %>% # group by sector, subsector and Non.CO2
+       summarise(x = sum(value)) -> # sum the values
+       L131.EPA_nonco2_indproc.mlt
+
+     # Second: Compute share of sectoral emissions in each subsector using EPA data
+     GCAM_sector_tech %>%
+       select(supplysector, subsector, stub.technology, EPA_agg_sector, EPA_agg_fuel_ghg, EDGAR_agg_sector) %>%
+       filter(EDGAR_agg_sector %in% c("industry_processes", "chemicals", "landfills", "wastewater","aerosols",
+                                      "metals","foams","solvents", "semiconductors")) %>%
```
bpbond: This string vector (`industry processes...semiconductors`) gets referenced several times, it looks like. Probably define in a constant (in this file) for convenience.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+
+     L131.nonco2_pct_R_prc_S_S_2005 %>%
+       group_by(EPA_agg_sector, Non.CO2) %>% # group by sector, and Non CO2
+       summarise(tech_emissions = sum(tech_emissions)) ->
+       L131.nonco2_gg_R_prc_S_2005
+
+
+     L131.nonco2_pct_R_prc_S_S_2005 %>%
+       group_by(EPA_agg_sector, Non.CO2) %>%
+       left_join(L131.nonco2_gg_R_prc_S_2005, by = c("EPA_agg_sector", "Non.CO2")) %>%
```
bpbond: Possible to use a more restrictive join here?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+     #filter out years 2009 and 2010 in EDGAR_VOC
+     EDGAR_VOC %>%
+       filter(year!= "2009" & year != "2010") ->
+       EDGAR_VOC
+
+     # combine all of the nonco2s together in one file
+       bind_rows(EDGAR_CH4, EDGAR_N2O, EDGAR_VOC, EDGAR_NOx, EDGAR_SO2, EDGAR_CO, EDGAR_NH3) %>%
+       left_join(EDGAR_sector, by = c("IPCC_description", "IPCC")) %>%
+       rename(EDGAR_agg_sector = agg_sector) %>%
+       left_join(distinct(EDGAR_nation), by = "ISO_A3") %>%
```
bpbond: Note we now have `left_join_keep_first_only`

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+       left_join(L131.EDGAR.mlt, by = c("GCAM_region_ID","EDGAR_agg_sector","Non.CO2", "year")) %>%
+       na.omit() %>%  # delete rows with NA's
+       mutate(input.emissions = EDGAR_emissions * tech_share) %>%
+       select(-sector_emissions, -tech_emissions) %>%
+      group_by(GCAM_region_ID, supplysector, subsector, stub.technology, Non.CO2, year) %>%
+       summarise(value = sum(input.emissions)) ->
+       L131.nonco2_tg_R_prc_S_S_Yh
+
+     # in order to match old data we have to turn the data in wide format
+     # which introduces NA's that are converted to 0. Then we can convert back to long format.
```
bpbond: OK, I guess, but it would be better to use `tidyr::complete`

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+
+     EPA_Ind %>%
+       left_join(EPA_tech, by = "Source_Category") %>%
+       rename(subsector = fuel) %>%
+       select(-EPA_Source_Category_Raw, -Source_Category) %>%   # Remove unnecessary columns
+       unique(.) %>%
+       gather(Non.CO2, value, -sector, -subsector) %>%  # make a new column called Non.co2
+       na.omit() %>%  # delete rows with NAs
+       group_by(sector, subsector, Non.CO2) %>% # group by sector, subsector and Non.CO2
+       summarise(x = sum(value)) -> # sum the values
```
bpbond: You rename this to `tech_emissions` in line 110 below; why not just do it here?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+       na.omit()  %>%  # delete rows with NA's
+       group_by(year, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+       summarise(EDGAR_emissions = sum(value)) %>%
+       mutate(EDGAR_emissions = EDGAR_emissions * CONV_GG_TG) -> # convert from gg to tg
+       L131.EDGAR.mlt
+
+     L131.EDGAR.mlt$GCAM_region_ID <- as.integer(L131.EDGAR.mlt$GCAM_region_ID)
+     L131.EDGAR.mlt$year <- as.integer(L131.EDGAR.mlt$year)
+
+     # Fifth: map in all data and compute emissions (EDGAR_emissions * tech_share)
```
bpbond: This whole section (below) needs more comments to help reader understand what's going on.

bpbond:
Thanks for documentation on `FLAG_SUM_TEST`...we may want to investigate this some more, too.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
 
-    # Load required inputs
+    # Load required inputs, set to long format and change column names where needed
+
     GCAM_region_names <- get_data(all_data, "common/GCAM_region_names")
     iso_GCAM_regID <- get_data(all_data, "common/iso_GCAM_regID")
     EDGAR_sector <- get_data(all_data, "emissions/EDGAR_sector")
-    EPA_ghg_tech <- get_data(all_data, "emissions/EPA_ghg_tech")
     EDGAR_nation <- get_data(all_data, "emissions/EDGAR_nation")
+    EPA_tech <- get_data(all_data, "emissions/EPA_ghg_tech")
```
kvcalvin: This seems like an unnecessary name change.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+      mutate(Non.CO2 = "NOx")->
+      EDGAR_NOx
+    get_data(all_data, "emissions/EDGAR_SO2") %>%
+      gather(year, value, -`IPCC-Annex`, -`World Region`, -ISO_A3, -Name, -IPCC, -IPCC_description) %>%
+      mutate(Non.CO2 = "SO2")->
+      EDGAR_SO2
+    get_data(all_data, "emissions/EDGAR_CO") %>%
+      gather(year, value, -`IPCC-Annex`, -`World Region`, -ISO_A3, -Name, -IPCC, -IPCC_description) %>%
+      mutate(Non.CO2 = "CO")->
+      EDGAR_CO
```
kvcalvin: @bpbond: Should these be taken care of upstream? Or is this going to have to happen in each emissions file?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+
+    ### Perform computations  ###
+
+    # First: Compute subsector share of sectoral emissions using EPA data
+
```
kvcalvin: Is this for all sectors? Or just industry/industrial processing? 

Also -- maybe note in the comment that these are US emissions? and US shares?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+      summarise(x = sum(value)) -> # sum the values
+      L131.EPA_nonco2_indproc.mlt
+
+    # Second: Compute share of sectoral emissions in each subsector using EPA data
+    GCAM_sector_tech %>%
+      select(supplysector, subsector, stub.technology, EPA_agg_sector, EPA_agg_fuel_ghg, EDGAR_agg_sector) %>%
+      filter(EDGAR_agg_sector %in% c("industry_processes", "chemicals", "landfills", "wastewater", "aerosols",
+                                     "metals", "foams", "solvents", "semiconductors")) %>%
+      repeat_add_columns(tibble(Non.CO2 = c("CH4", "N2O", "NMVOC", "NOx", "SO2", "CO", "VOC"))) ->
+      L131.nonco2_pct_R_prc_S_S_2005
```
kvcalvin: This data frame doesn't have any values in it yet, right?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+
+    # Third: Group by and then join by sector, subsector, and Non.CO2.  However, to join the columns they have to have
+    # the same name. Rename column to tech_emissions. Tech_emissions is the sum and tech_share is the tech_emissions/sector_emissions
+    L131.nonco2_pct_R_prc_S_S_2005 %>%
+      group_by(EPA_agg_sector, EPA_agg_fuel_ghg, Non.CO2) %>%
+      left_join(L131.EPA_nonco2_indproc.mlt, by = c(EPA_agg_sector = "sector", EPA_agg_fuel_ghg = "subsector", "Non.CO2")) %>%
+      rename(tech_emissions = x) %>%
+      mutate(tech_emissions = if_else(is.na(tech_emissions), 1, tech_emissions)) ->
```
kvcalvin: Add a comment here that says we're resetting missing values to 1 and that the emissions coef will be overwritten later.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+      summarise(tech_emissions = sum(tech_emissions)) ->
+      L131.nonco2_gg_R_prc_S_2005
+
+    L131.nonco2_pct_R_prc_S_S_2005 %>%
+      group_by(EPA_agg_sector, Non.CO2) %>%
+      left_join(L131.nonco2_gg_R_prc_S_2005, by = c("EPA_agg_sector", "Non.CO2")) %>%
+      rename(sector_emissions = tech_emissions.y) %>% # sector emisisons from thr join
+      rename(tech_emissions = tech_emissions.x) %>%
+      mutate(tech_share = tech_emissions/sector_emissions) ->
+      L131.nonco2_pct_R_prc_S_S_2005
```
kvcalvin: I think you can merge the last two pipelines together doing something like:

`tech_share = tech_emissions / sum( tech_emissions )`.

I'll let Ben say whether this is needed though.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+
+    # Fourth: Disaggregate EDGAR emissions to subsectors
+    # First aggregate all EDGAR data and subset for processing sectors
+
+    # filter out years 2009 and 2010 in EDGAR_VOC
+    EDGAR_VOC %>%
+      filter(year!= "2009" & year != "2010") ->
```
kvcalvin: Do we want to make this more generic in case we ever get more EDGAR data?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -36,55 +39,168 @@ module_emissions_L131.nonco2_proc_R_S_T_Y_DISABLED <- function(command, ...) {
...
+                                     "metals", "foams", "solvents", "semiconductors")) %>%
+      select(year,value, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+      na.omit()  %>%  # delete rows with NA's
+      group_by(year, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+      summarise(EDGAR_emissions = sum(value)) %>%
+      mutate(EDGAR_emissions = EDGAR_emissions * CONV_GG_TG) -> # convert from gg to tg
+      L131.EDGAR.mlt
+
+    L131.EDGAR.mlt$GCAM_region_ID <- as.integer(L131.EDGAR.mlt$GCAM_region_ID)
+    L131.EDGAR.mlt$year <- as.integer(L131.EDGAR.mlt$year)
```
kvcalvin: Maybe add these two lines to the pipeline above using a `mutate`?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: i agree but that's what the old data system has so i kept it.  thoughts?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: i was confused by this.  most of the other data is coming in up to 2008, and VOC has an extra 2 years of data.  This doesn't allow us at all to add in data in the future.  Maybe i'll open an issue?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: good point.  Ben also brought this up, I should double check it's even needed.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
bpbond: `EPA_ghg_tech` is clearer and more consistent...I would change it.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
bpbond: Good question! If multiple files are going to use this EDGAR data, and they're all going to need to make these transformations, we can definitely have a data chunk handle it once, upstream.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
bpbond: You could definitely do that; but this way is perhaps clearer? I guess I'd change to
```R
rename(sector_emissions = tech_emissions.y) %>% # sector emisisons from thr join
      mutate(tech_share = tech_emissions.x/sector_emissions)
```

Also @cahartin note "sector emisisons from thr join" which is a colorful comment :)

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
bpbond: Yeah, it would be good to get clarification.

inst/extdata/emissions/EPA_FCCC_IndProc_2005.csv
```diff
@@ -0,0 +1,66 @@
+# File: EPA_FCCC_IndProc_2005.csv,,,,,,,,,,,
+# Title: Sectoral Report for Energy,,,,,,,,,,,
+# Units: NA,,,,,,,,,,,
+# Comments:  I NEED TO FIND OUT WHAT THIS IS,,,,,,,,,,,
```
bpbond: Yes.

bpbond:
I do NOT understand why this chunk, which has differences NO LARGER THAN 10^-14, fails the old-new comparison. Grr. Anyway, totally fine to use the sum test.

bpbond:
@cahartin I have split the EDGAR-reading code into a separate data chunk, per @rplzzz comments above, and have changed your chunk so that it uses those outputs.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: for now I'll leave it. should i open an issue?

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: 👍

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: thanks!

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: 👍

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: ah, good catch. it looks like my comments are out of place.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: @kvcalvin  why do we do this?

kvcalvin: @cahartin: We have problems in some region/sector/tech/gas combinations where the EPA USA inventory has zero emissions but the EDGAR inventory has positive emissions. If we leave the this as an NA or as a 0, then we can never get positive emissions when we scale to EDGAR. There is an issue open about this because it does have some unintended consequences when one tech is NA and another tech in the same sector has a value.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: @bpbond thanks! i tried really hard on that one!

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
cahartin: thanks!

bpbond:
@cahartin It would be great to get this merged today if you have time. Let me know and I can provide support if useful.

bpbond:
Looks great. I made a few tweaks and changed once instance of `left_join(distinct(...))` to `left_join_keep_first_only`. This is now holding on @kvcalvin approval.

R/zchunk_L131.nonco2_proc_R_S_T_Y.R
```diff
@@ -1,90 +1,155 @@
...
+      filter(EDGAR_agg_sector %in% c("industry_processes", "chemicals", "landfills", "wastewater", "aerosols",
+                                     "metals", "foams", "solvents", "semiconductors")) %>%
+      select(year,value, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+      na.omit()  %>%  # delete rows with NA's
+      group_by(year, Non.CO2, EDGAR_agg_sector, GCAM_region_ID) %>%
+      summarise(EDGAR_emissions = sum(value)) %>%
+      mutate(EDGAR_emissions = EDGAR_emissions * CONV_GG_TG) -> # convert from gg to tg
+      L131.EDGAR.mlt
+
+    #L131.EDGAR.mlt$GCAM_region_ID <- as.integer(L131.EDGAR.mlt$GCAM_region_ID)
```
kvcalvin: Delete this?

Issue #346, opened 2017-05-04
LB152.ag_GTAP_R_C_GLU_irr


kvcalvin:
@bpbond: next chunk for Ryna. Can you set up?

PR #351, opened 2017-05-04
Ryna - LB152.ag_GTAP_R_C_GLU_irr

rynacui:
#346


rynacui:
I got one test error regarding the output file `L152.ag_rfdHA_bm2_R_C_GLU.csv`, basically saying there is one mismatched observation. 

But I've checked the ratio between old and new for all data and all equal to 1, and the mismatched row both have the value of `0.02250`.

bpbond:
Looking at the one failure, here are the differences I see:
```R
> summary(combined$rfdHA- combined$rfdHA.old)
-4.547e-13 -6.420e-17  0.000e+00  1.800e-17  4.860e-17  1.208e-13 
```
I.e., differences are < 10^-12. WHY IS THIS FAILING?!?

Here's the plot of old versus new (note log scale):

compare

rynacui:
Thanks @bpbond ! I'm very happy of getting a simple chunk!

bpbond:
This is holding on review by @kvcalvin

Issue #338, opened 2017-05-03
Refactor base year scaling operation


rplzzz:
We do a lot of scaling stuff by base year ratios.  Specifically, we compute the ratio of a time series to its base year value, and then we find the base year value of some other quantity and multiply the ratio series by that base year value.  This seems like a good candidate for refactoring into a utility function, since we waste a lot of lines of code rediscovering this idiom everywhere it is used.

_However_, it may be that as we work forward in the data pipeline we don't do this so much anymore (i.e., we're working off of time series from previous stages where this was already done).  The question is, is there enough value in whatever instances of this operation are still ahead to justify doing the refactoring?

bpbond:
I definitely agree that this may be a good candidate. Tagging @pkyle for his thoughts on how much _more_ for this is out there.

pkyle:
Hmm...not sure how common it is, and the method here is pretty similar to others that might be captured under a similar function, or even the same one. Off the top of my head, it's used in socioeconomics for merging historical and future data, in the energy sector for computing industrial sector output and income elasticities (L232), in the ag sector for merging historical and projected yield, in land for estimating land cover prior to 1971.
WRT the broader applications, there many of instances of scaling and/or expanding one year's portional allocations to other years. However I don't know that a function would be easier or more transparent than just implementing the individual steps?

rplzzz:
Abstracting an operation like that is always better than implementing the individual steps, both for productivity and reliability.  The question here is, have we already done this as many times as we're going to do it, or are there still gains to be had?

pkyle:
I don't know; if I had to guess I'd say 60% of the instances of this are already done, and 40% remain. Totally made up tho.

PR #353, opened 2017-05-05
Jill - LA124.heat

jhoring:
#177


R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      rename(fuel=heat) %>%
+      gather(key=year, value=value, -fuel, -sector, -GCAM_region_ID) %>%
+      group_by(fuel, sector, GCAM_region_ID, year) %>%
+      summarise(value=sum(value)) %>%
+      mutate(year = sub("X", "", year)) %>%
+      mutate(year = as.integer(year)) -> L124.in_EJ_R_heat_F_Yh
+
+    # Heat production from district heat sector
+    # Create empty table of historical years to fill with interpolation
+    A24.globaltech_coef[,as.character(HISTORICAL_YEARS[-1])] <- NA
```
bpbond: This is...weird. Is the `A24.globaltech_coef` frame wide? Why are you dropping the first column?

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+
+    # Interpolate to fill in missing globaltech_coef historical years
+    A24.globaltech_coef %>%
+      gather(year, value, -supplysector, -subsector, -technology, -minicam.energy.input) %>%
+      mutate(year = as.integer(year))  %>%
```
bpbond: Same thing as above -- I'd suggest putting these two lines (66-67) right up after the `getdata` call above. They'll be deleted once this chunk starts using 'real' data.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      summarise(value=sum(value)) %>%
+      filter(fuel != is.na(fuel)) %>%
+      mutate(year = sub("X", "", year)) %>%
+      mutate(year = as.integer(year)) -> L124.out_EJ_R_heatfromelec_F_Yh
+
+
+    #Secondary output coefficients on heat produced by main activity CHP plants
+    L1231.out_EJ_R_elec_F_tech_Yh %>%
+      filter(GCAM_region_ID %in% heat_regionIDs) %>%
+      filter(technology %in% calibrated_techs$technology[calibrated_techs$secondary.output == "heat"]) %>%
```
bpbond: What's going on in this `filter`? Complicated syntax--needs a comment explaining at least.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      mutate(value=if_else(is.na(value)|is.infinite(value), 0, value))-> L124.heatoutratio_R_elec_F_tech_Yh
+
+    #Drop all rows where all years are 0
+    L124.heatoutratio_R_elec_F_tech_Yh %>%
+      group_by(GCAM_region_ID, technology, sector, fuel) %>%
+      summarise(sum=sum(value)) %>%
+      filter(sum==0) -> all_years_0
+
+    L124.heatoutratio_R_elec_F_tech_Yh %>%
+      left_join(all_years_0) %>%
```
bpbond: Here and elsewhere, is it possible to use a more restrictive join, e.g. `left_join_error_no_match`?

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      mutate(sum=if_else(is.na(sum), 2, sum)) %>% #2 was a random, nonzero choice
+      filter(sum!=0)  %>%
+      select(-sum) -> L124.heatoutratio_R_elec_F_tech_Yh
+
+    # This is complicated. Some historical regions / years have 0 district heat output (all fuels). This is fine unless there is heat output from
+    # IEA's MAINCHP flow, which we model in the electric sector with a secondary output of heat to the district heat sector. In GCAM this will
+    # cause solution failure as the district heat supply will be 0. In order to avoid this outcome we need a nominal amount of heat produced
+    # in the zero years. The code below sets an exogenous year as the one to use to extract fuel shares for each region. In all 0 region / years,
+    # output is equal to the output in the fuel-share year times 1e-3 to avoid throwing off the energy balances.
+    fuel_share_year <- "2010"
```
bpbond: Should be integer - again, deal with the years-and-X's-as character thing once, at the beginning.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
-    # If no precursors (very rare) don't call `add_precursor` at all
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L124.in_EJ_R_heat_F_Yh %>%
+      add_title("Inputs to heat by GCAM region / fuel / historical year") %>%
+      add_units("EJ") %>%
+      add_comments("Data on input heat is read in, aggregated") %>%
```
bpbond: Not a great comment; improve. Aggregated by what, for example? Why?

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
       L124.in_EJ_R_heat_F_Yh
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L124.out_EJ_R_heat_F_Yh %>%
+      add_title("Output of district heat sector by GCAM region / fuel / historical year") %>%
+      add_units("EJ") %>%
+      add_comments("Output heat calculated based on input divided by a technology coefficient") %>%
```
bpbond: 👍

inst/extdata/energy/A24.globaltech_coef.csv
```diff
@@ -0,0 +1,8 @@
+# File: A24.globaltech_coef.csv,,,,,
```
bpbond: Here and in files below, please remove these stupid Excel commas at end of header lines

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
-    #   ... new code with a fix
-    # }
-    #
-    #
-    # NOTE: there are 'match' calls in this code. You probably want to use left_join_error_no_match
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
-    # NOTE: This code uses vecpaste
-    # This function can be removed; see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
+    # Create list of heat regions
+    heat_regionIDs <- A_regions$GCAM_region_ID[ A_regions$heat == 1]
```
rplzzz: This is _not_ selecting the USA region.  It is selecting all those regions for which `A_regions$heat` is 1.  This could be clarified a bit by filtering the table based on the `heat` column and then extracting the region column.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+
+    # Fuel inputs to district heat
+    # Process fuel inputs in all regions; some will have the energy assigned to bld/ind, and others have a district heat sector
+    L1011.en_bal_EJ_R_Si_Fi_Yh  %>%
+      filter(sector == "in_heat") %>%
+      mutate(sector = sub("in_", "", sector)) %>%
+      left_join(enduse_fuel_aggregation, by="fuel") %>%
+      select(-fuel, -electricity, -bld, -industry, -trn) %>%
+      rename(fuel=heat) %>%
```
rplzzz: It's not at all clear what's happening in this pipeline.  Comments should give us an idea of what we're starting with and what we will end up with.  Some explanation of why we are doing this would also not be amiss.

Be especially cognizant of this when using the negative form of `select`.  When you select a (positive) list of columns, then I at least know what will be in the resulting table.  When you tell me what columns you are dropping (i.e., by using the negative form), it may not be obvious what is left.

R/zchunk_LA124.heat.R
rplzzz: Actually, this code is _adding_ columns, not dropping them.  It is setting all historical year columns (except 1971) to NA for later interpolation.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+
+    # Interpolate to fill in missing globaltech_coef historical years
+    A24.globaltech_coef %>%
+      gather(year, value, -supplysector, -subsector, -technology, -minicam.energy.input) %>%
+      mutate(year = as.integer(year))  %>%
+      arrange(year) %>%
+      group_by(technology, subsector, supplysector, minicam.energy.input)  %>%
+      mutate(value = approx_fun(year, value)) %>%
+      left_join(distinct(calibrated_techs)) %>%
+      select(-calibration, -secondary.output) -> L124.globaltech_coef
```
rplzzz: Again, after an operation like this, drop a comment in to let us know what the structure of the resulting table is.  Don't make us mentally simulate each of the steps in the pipeline to figure it out ourselves.  (Alternately, just use the positive form of `select`.  You might have to list more columns, but at least it's self-documenting).

R/zchunk_LA124.heat.R
bpbond: @rplzzz whoops, thanks

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+    # IEA's MAINCHP flow, which we model in the electric sector with a secondary output of heat to the district heat sector. In GCAM this will
+    # cause solution failure as the district heat supply will be 0. In order to avoid this outcome we need a nominal amount of heat produced
+    # in the zero years. The code below sets an exogenous year as the one to use to extract fuel shares for each region. In all 0 region / years,
+    # output is equal to the output in the fuel-share year times 1e-3 to avoid throwing off the energy balances.
+    fuel_share_year <- "2010"
+    L124.in_EJ_R_heat_F_Yh %>%
+      group_by(sector, GCAM_region_ID, year) %>%
+      summarise(value=sum(value)) %>%
+      mutate(value=if_else(value==0, 1e-3, value)) %>%
+      mutate(value=if_else(value!=1e-3, 0, value)) -> L124.mult_R_heat_Yh
```
rplzzz: These last two lines are very confusing.  As best I can tell, the idea is to zero out everything that is nonzero, and set everything that was initially zero to 1e-3.  That seems a very strange thing to do.  If that really is what we want to do, then we should probably make it explicit:  `mutate(value = if_else(value == 0.0, 1e-3, 0.0))`.

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+
+    L124.out_EJ_R_heatfromelec_F_Yh %>%
+      group_by(sector, GCAM_region_ID, year) %>%
+      summarise(elec_heat=sum(value)) %>%
+      mutate(elec_heat=if_else(elec_heat!=0, 1, elec_heat))-> L124.mult_R_heatfromelec_Yh
+
+    L124.mult_R_heat_Yh %>%
+      left_join(L124.mult_R_heatfromelec_Yh %>%
+                  rename(temp=sector),
```
rplzzz: Why are we doing this rename and then dropping the renamed column later on (without using it)?  

Also, I gather that all of this is meant to be implementing the steps described in the comments starting on line 127.  It's good that we explained the overall plan, but we should really be explaining how each of these steps ties into that plan.  What, for example, are we doing in this step?

R/zchunk_LA124.heat.R
```diff
@@ -40,70 +40,175 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      filter(year==fuel_share_year) %>%
+      select(-year) %>%
+      rename(fsy_val=value)-> fsy_in
+
+    L124.in_EJ_R_heat_F_Yh %>%
+      left_join(fsy_in, by= c("fuel", "sector", "GCAM_region_ID")) %>%
+      left_join(L124.mult_R_heat_Yh %>%
+                  rename(heat_val=value), by=c("sector", "GCAM_region_ID", "year")) %>%
+      mutate(value=value+(fsy_val*heat_val)) %>%
+      select(-fsy_val, -heat_val) -> L124.in_EJ_R_heat_F_Yh
```
rplzzz: From line 152 to here looks like another instance of the base year scaling described in #338.

inst/extdata/energy/A24.globaltech_coef.csv
skim301: Are these commas removed manually from the original csv? That file is used in multiple chuncks.

bpbond:
@skim301 this is waiting for revisions for @jhoring . I agree, makes sense for you to do your review after her revisions come in.

jhoring:
@bpbond I believe I have updated the PR to address these comments. Let me know if it doesn't look different!

R/zchunk_LA124.heat.R
```diff
@@ -36,29 +37,36 @@ module_energy_LA124.heat <- function(command, ...) {
...
-      select(-fuel, -electricity, -bld, -industry, -trn) %>%
-      rename(fuel=heat) %>%
-      gather(key=year, value=value, -fuel, -sector, -GCAM_region_ID) %>%
-      group_by(fuel, sector, GCAM_region_ID, year) %>%
-      summarise(value=sum(value)) %>%
-      mutate(year = sub("X", "", year)) %>%
-      mutate(year = as.integer(year)) -> L124.in_EJ_R_heat_F_Yh
+      select(GCAM_region_ID, sector, year, value, heat) %>%
+      rename(fuel = heat) %>%
+      group_by(fuel, sector, GCAM_region_ID, year) -> temp %>%
```
skim301: is the order reversed on rename( heat = fuel)? Are you renaming heat to fuel?

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
 
     # Heat output: fuel inputs to heat divided by exogenous input-output coefficients
     L124.in_EJ_R_heat_F_Yh %>%
       left_join(L124.globaltech_coef %>%
-                  rename(temp=value),
-                      by=c("sector", "fuel", "year")) %>%
-      mutate(value=value/temp)%>%
+                  rename(temp = value),
+                      by = c("sector", "fuel", "year")) %>%
+      mutate(value = value / temp) %>%
```
skim301: Bit confusing here with temp and value with the nested structure. You could rename temp to temp_IOCoef for clarity.

R/zchunk_LA124.heat.R
skim301: could break this up by finding calibrated techs with heat output prior to starting this section of code.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
-      mutate(year = sub("X", "", year)) %>%
-      mutate(year = as.integer(year)) %>%
       left_join(L124.out_EJ_R_heatfromelec_F_Yh %>%
-                  rename(value_elec=value) %>%
-                  rename(temp=sector), by=c("GCAM_region_ID", "fuel", "year")) %>%
-      mutate(value=value_elec/value) %>%
-      select(-value_elec, -temp)-> L124.heatoutratio_R_elec_F_tech_Yh
+                  rename(value_elec = value) %>%
+                  rename(temp = sector), by=c("GCAM_region_ID", "fuel", "year")) %>%
+      mutate(value = value_elec / value) %>%
```
skim301: It'll be good to have a comment to figure out what's being divided by what.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
+      select(GCAM_region_ID, sector, fuel, technology, year, value) %>%
+      # Reset missing and infinite values (applicable for CC in the base years) to 0
+      mutate(value = if_else(is.na(value)|is.infinite(value), 0, value)) -> L124.heatoutratio_R_elec_F_tech_Yh
 
-    #Reset missing and infinite values (applicable for CC in the base years) to 0
-    L124.heatoutratio_R_elec_F_tech_Yh %>%
-      mutate(value=if_else(is.na(value)|is.infinite(value), 0, value))-> L124.heatoutratio_R_elec_F_tech_Yh
 
-    #Drop all rows where all years are 0
+    # Drop all rows where all years are 0
```
skim301: please revise comment to "Drop all rows where value = 0 for all years".

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
+    # Create a table of all the years of all 0
     L124.heatoutratio_R_elec_F_tech_Yh %>%
       group_by(GCAM_region_ID, technology, sector, fuel) %>%
-      summarise(sum=sum(value)) %>%
-      filter(sum==0) -> all_years_0
+      summarise(sum = sum(value)) %>%
+      filter(sum == 0) -> all_years_0
```
skim301: maybe something more descriptive than all_years_0 would be helpful. Such as tech_null_outputs or something like that. I'm not even sure whether these rows are.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
 
+    # Filter out the years in all_years_0 from heatoutratio
     L124.heatoutratio_R_elec_F_tech_Yh %>%
       left_join(all_years_0) %>%
-      mutate(sum=if_else(is.na(sum), 2, sum)) %>% #2 was a random, nonzero choice
-      filter(sum!=0)  %>%
+      #2 is a random, non-zero option for temporary use
+      mutate(sum = if_else(is.na(sum), 2, sum)) %>%
```
skim301: maybe 1 would be better than 2, since 0 or 1 would represent false or true.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
+
+    fuel_share_year <- 2010
     L124.in_EJ_R_heat_F_Yh %>%
       group_by(sector, GCAM_region_ID, year) %>%
-      summarise(value=sum(value)) %>%
-      mutate(value=if_else(value==0, 1e-3, value)) %>%
-      mutate(value=if_else(value!=1e-3, 0, value)) -> L124.mult_R_heat_Yh
+      summarise(value = sum(value)) %>%
+      # Creating an intermediary table - for years with 0 heat output, set value to 1e-3, for all others, set value to 0
+      mutate(value = if_else(value == 0, 1e-3, 0)) -> L124.mult_R_heat_Yh
```
skim301: if I understand this correctly, for those with heat output we are going to use a minimal scaler (1e-3) to represent a non-zero secondary output. Otherwise, the scaler is 0. So I would define output_scaler = 1e-3 above this section of code, with a comment, and then use it here.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
 
     L124.out_EJ_R_heatfromelec_F_Yh %>%
       group_by(sector, GCAM_region_ID, year) %>%
-      summarise(elec_heat=sum(value)) %>%
-      mutate(elec_heat=if_else(elec_heat!=0, 1, elec_heat))-> L124.mult_R_heatfromelec_Yh
+      summarise(elec_heat = sum(value)) %>%
+      # Creating an intermediary table - for years with nonzero heat from CHP, set value to 1, for 0 values, leave as 0
+      mutate(elec_heat = if_else(elec_heat != 0, 1, elec_heat))-> L124.mult_R_heatfromelec_Yh
```
skim301: Very confusing original code. I think 1 here is so that value * elect_heat below results in 1e-3. Isn't it clearer to set value = output_scaler (defined above) when elec_heat = 0 after line 157 instead.

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
       select(-sum) -> L124.heatoutratio_R_elec_F_tech_Yh
 
     # This is complicated. Some historical regions / years have 0 district heat output (all fuels). This is fine unless there is heat output from
     # IEA's MAINCHP flow, which we model in the electric sector with a secondary output of heat to the district heat sector. In GCAM this will
     # cause solution failure as the district heat supply will be 0. In order to avoid this outcome we need a nominal amount of heat produced
     # in the zero years. The code below sets an exogenous year as the one to use to extract fuel shares for each region. In all 0 region / years,
     # output is equal to the output in the fuel-share year times 1e-3 to avoid throwing off the energy balances.
-    fuel_share_year <- "2010"
+
+    fuel_share_year <- 2010
```
skim301: fuel_share_year and later fsy gets very confusing. I would prefer historical_year and hist_value or norm_year or norm_value (normalized).

R/zchunk_LA124.heat.R
```diff
@@ -69,117 +77,121 @@ module_energy_LA124.heat <- function(command, ...) {
...
       select(-temp, -elec_heat) %>%
-      mutate(value=if_else(is.na(value), 0, value))-> L124.mult_R_heat_Yh
+      mutate(value = if_else(is.na(value), 0, value))-> L124.mult_R_heat_Yh
 
     L124.in_EJ_R_heat_F_Yh %>%
-      filter(year==fuel_share_year) %>%
+      filter(year == fuel_share_year) %>%
       select(-year) %>%
-      rename(fsy_val=value)-> fsy_in
+      rename(fsy_val = value)-> fsy_in
```
skim301: related comment above at line 138. fsy_val could be hist_value or norm_value, fsy_in could be hist_in or norm_in.

R/zchunk_LA124.heat.R
skim301: I'm not even sure what category these rows are

jhoring:
@skim301 Thanks for your comments! I'll address them now.

R/zchunk_LA124.heat.R
jhoring: I'm not sure... I think putting it in two steps like that was supposed to make it more clear (?) but happy to move it.

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,189 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+
+    # Fuel inputs to district heat
+    # Process fuel inputs in all regions; some will have the energy assigned to bld/ind, and others have a district heat sector
+    L1011.en_bal_EJ_R_Si_Fi_Yh  %>%
+      filter(sector == "in_heat") %>%
+      mutate(sector = sub("in_", "", sector)) %>%
+      left_join(enduse_fuel_aggregation, by="fuel") %>%
+      select(GCAM_region_ID, sector, year, value, heat) %>%
+      rename(fuel = heat) %>%
+      group_by(fuel, sector, GCAM_region_ID, year) -> temp %>%
```
bpbond: This line is syntactically incorrect, which means you didn't run the driver, much less the tests, before committing this and pushing it to GitHub. 15 demerits @jhoring !  😠   ...I have some work to do on this PR and will fix.

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+    A24.globaltech_coef[,as.character(HISTORICAL_YEARS[-1])] <- NA
+
+    # Interpolate to fill in missing globaltech_coef historical years
+    A24.globaltech_coef %>%
+      gather(year, value, -supplysector, -subsector, -technology, -minicam.energy.input) %>%
+      mutate(year = as.integer(year))  %>%
+      arrange(year) %>%
+      group_by(technology, subsector, supplysector, minicam.energy.input)  %>%
+      mutate(value = approx_fun(year, value)) %>%
+      left_join(distinct(calibrated_techs)) %>%
```
bpbond: Need `by = ...`

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+    A_regions %>%
+      filter(heat == 1) %>%
+      select(GCAM_region_ID) -> heat_regionIDs
+
+    # Fuel inputs to district heat
+    # Process fuel inputs in all regions; some will have the energy assigned to bld/ind, and others have a district heat sector
+    L1011.en_bal_EJ_R_Si_Fi_Yh  %>%
+      filter(sector == "in_heat") %>%
+      mutate(sector = sub("in_", "", sector)) %>%
+      left_join(enduse_fuel_aggregation, by="fuel") %>%
```
bpbond: Spaces around `=` and all other math operators please

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+    # Drop all rows where value = 0 for all years
+    # Create a table of all the years of all value = 0
+    L124.heatoutratio_R_elec_F_tech_Yh %>%
+      group_by(GCAM_region_ID, technology, sector, fuel) %>%
+      summarise(sum = sum(value)) %>%
+      filter(sum == 0) -> years_heatout_0
+
+    # Filter out the years in years_heatout_0 from heatoutratio
+    L124.heatoutratio_R_elec_F_tech_Yh %>%
+      left_join(years_heatout_0) %>%
```
bpbond: Need `by = ....`

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      #Using 1 for all rows where heatout is not 0 for all years
+      mutate(sum = if_else(is.na(sum), 1, sum)) %>%
+      filter(sum != 0)  %>%
+      select(-sum) -> L124.heatoutratio_R_elec_F_tech_Yh
+
+    # This is complicated. Some historical regions / years have 0 district heat output (all fuels). This is fine unless there is heat output from
```
bpbond: >This is complicated

No kidding!

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
+      filter(year == norm_year) %>%
+      select(-year) %>%
+      rename(norm_val = value)-> norm_in
+
+    # If heat output is 0 and heat from CHP is nonzero, value for heat input will be equal to
+    # the heat input plus the fuel share year input times 1e-3
+    L124.in_EJ_R_heat_F_Yh %>%
+      left_join(norm_in, by= c("fuel", "sector", "GCAM_region_ID")) %>%
+      left_join(L124.mult_R_heat_Yh %>%
+                  rename(mult_val = value), by=c("sector", "GCAM_region_ID", "year")) %>%
```
bpbond: `left_join(rename(L124.mult_R_heat_Yh, mult_val = value), by=c("sector", "GCAM_region_ID", "year"))` probably clearer?

bpbond:
There's one another issue - this chunk is failing the timeshift test I think. More on this in a bit.

R/zchunk_LA124.heat.R
```diff
@@ -36,74 +37,186 @@ module_energy_LA124.heat_DISABLED <- function(command, ...) {
...
-    #
-    #
-    # NOTE: there are 'match' calls in this code. You probably want to use left_join_error_no_match
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
-    # NOTE: This code uses vecpaste
-    # This function can be removed; see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
-    # ===================================================
+    # Create list of heat regions
+    A_regions %>%
+      filter(heat == 1) %>%
```
bpbond: Please add a comment explaining what this filter is doing. What does `heat == 1` signify?

R/zchunk_LA124.heat.R
bpbond: Has this `by` been added?

R/zchunk_LA124.heat.R
jhoring: I'm actually not sure. It seems like some regions model heat (==1) and some do not (==0). I don't know why that is though

bpbond: OK, thanks. @pkyle @skim301 can someone weigh in here? What does this mean?

jhoring:
@bpbond @skim301 An updated version has been committed!

R/zchunk_LA124.heat.R
skim301: It looks like heat is a logical variable to determine which regions have district heat. Not sure how A_regions.csv, which contains several other information, is created in regards to identifying regions with district heat. I think it's done manually, but I wonder if it can be done using the IEA energy balance data which the determination must be based on. I would recommend, replacing "heat" in A_regions.csv to "has_district_heat", and then use filter( has_district_heat == 1 ) above.

skim301: hang in there Jill.

pkyle: Yes that's exactly right--this is just a boolean indicating whether a region has district heat that we model. We can't just infer the condition from the IEA energy balances because many regions have countries with non-zero values in some years, but aren't modeled as having district heat as an energy commodity in GCAM. Re-naming the "heat" column to "has_district_heat" sounds good to me.

jhoring: Thanks! I'll make that change and add comment "create list of regions with district heat modeled"

jhoring:
@bpbond @skim301 please check this latest commit!

R/zchunk_LA124.heat.R
skim301: # Calibrated technologies with secondary heat output
calibrated_techs_heat <- filter(calibrated_techs, secondary.output == "heat")

L1231.out_EJ_R_elec_F_tech_Yh %>%
       filter(GCAM_region_ID %in% heat_regionIDs$GCAM_region_ID) %>%
       # Select only technologies that have heat output in calibrated techs mapping
       filter(technology %in% calibrated_techs_heat$technology) %>%

skim301: This might be easier to read.
calibrated_techs_heat <- filter(calibrated_techs, secondary.output == "heat")

L1231.out_EJ_R_elec_F_tech_Yh %>%
       filter(GCAM_region_ID %in% heat_regionIDs$GCAM_region_ID) %>%
       # Select only technologies that have heat output in calibrated techs mapping
       filter(technology %in% calibrated_techs_heat$technology) %>%

R/zchunk_LA124.heat.R
```diff
@@ -90,19 +90,18 @@ module_energy_LA124.heat <- function(command, ...) {
...
 
-
     # Secondary output coefficients on heat produced by main activity CHP plants
     L1231.out_EJ_R_elec_F_tech_Yh %>%
       filter(GCAM_region_ID %in% heat_regionIDs$GCAM_region_ID) %>%
       # Select only technologies that have heat output in calibrated techs mapping
       filter(technology %in% calibrated_techs$technology[calibrated_techs$secondary.output == "heat"]) %>%
-      left_join(L124.out_EJ_R_heatfromelec_F_Yh %>%
+      left_join_error_no_match(L124.out_EJ_R_heatfromelec_F_Yh %>%
                   rename(value_heatfromelec = value) %>%
```
skim301: Per Ben's earlier suggestion, the join with multiple piping and renaming is difficult to read so we could separate that.
L124.out_EJ_R_heatfromelec_F_Yh %>% rename(value_heatfromelec = value) %>%
                       rename(temp = sector) -> L124.out_EJ_R_heatfromelec_F_Yh

and then 
left_join_error_no_match(L124.out_EJ_R_heatfromelec_F_Yh, by = c("GCAM_region_ID", "fuel", "year"))

jhoring: This is one of the output tables, so I would need to create an intermediary table with a different name or name those columns back. Might make things even more confusing. What do you think?

skim301: Ok. Leave it then.

jhoring: ok! Thanks for all your help on this!

bpbond:
@jhoring Take a look at my last commit. I changed a single line in your code, which makes it robust to the 'timeshift` test, i.e. non-crashing if `HISTORICAL_YEARS` changes.

That said, nice job again! Merging.

Issue #295, opened 2017-04-26
Error being thrown in tests for current master


rplzzz:
The current master branch (4ffab50) throws an error while running its tests:
```
DONE ===========================================================================
Error in left_join_error_no_match(., L101.Pop_thous_GCAM3_ctry_Y, by = c("iso",  : 
  left_join_no_match: NA values in new data columns
```
For some reason this isn't causing any of the tests to fail.  Perhaps it's happening during some setup in the testing code that isn't actually being subjected to an `expect` call?  We should investigate the cause.

bpbond:
The error is occurring in line 124 of `LA143.HDDCDD`, under the timeshift test:
```R
    # Add population data and region data
    L143.wtHDDCDD_scen_ctry_Y <- L143.HDDCDD_scen_ctry_Y %>%
      # Join with population data converted to long format
      left_join_error_no_match(L101.Pop_thous_GCAM3_ctry_Y,
                               by = c("iso", "year")) %>%
```

I haven't look at the relevant data frames in detail but pretty clearly when `HISTORICAL_YEARS` and/or `FUTURE_YEARS` get changed there are NAs in this step. I think we need to do two things:

1. Ask @russellhz to take a look at this chunk again. In particular, there probably needs to be an interpolation step to ensure that we have data for all years? Russell, the time shift test is changing these years variables and re-running the chunks, trying to find (as it did here) fragility.
2. Figure out why no 'real' error is getting thrown in `test_timeshift.R`. This is being run inside of a `with_mock` call, which may be causing odd effects. I tried putting an `expect_silent` inside it no luck; am out of time at the moment, but can come back to this later.

rplzzz:
Yeah, I had to do some extra work to get the time shift test to work properly (the problem is 2011, for which we typically don't have data as it is neither a historical year, nor a future year).

Also, in that code snippet you posted, we are joining with `L101.Pop_thous_GCAM3_ctry_Y`, which seems really weird, as I don't think these are GCAM3 scenarios.  We should have another look at that too.

russellhz:
L101.Pop_thous_GCAM3_ctry_Y gets read in with historical years 1971:2010 by 1 year and 2015:2100 by 5 years. The HDDCDD data begins with all years from 1971:2100. It later gets trimmed to match the population file years with 
`filter(year %in% c(HISTORICAL_YEARS, FUTURE_YEARS))`

[...]

PR #357, opened 2017-05-05
Russell-LA143.HDDCDD_fix

russellhz:
Fix to #295


R/zchunk_LA143.HDDCDD.R
```diff
@@ -118,10 +118,18 @@ module_energy_LA143.HDDCDD <- function(command, ...) {
         mutate(iso = if_else(iso == "scg", "mne", iso))
     }
 
+    # Extend population data to all years
+    iso_list <- tibble(iso = L101.Pop_thous_GCAM3_ctry_Y$iso %>% unique())
+    all_years <- tibble(year = seq(1971, 2100))
```
bpbond: These numbers should probably come from `min(HISTORICAL_YEARS)` and `max(FUTURE_YEARS)`, rather than be hardcoded.

PR #358, opened 2017-05-06
Build and run XML-generating chunks

bpbond:
Each Level 2 script generates one or more outputs datasets. Then a new category of code chunks, xml chunks, take those outputs and (via calling `create_xml`) return the data structures necessary for writing as a GCAM XML input. This final step occurs in `save_chunkdata`, after the driver has run.

Basically, our coders don't need to do anything w.r.t. XML writing--just, probably, check the relevant batch file(s) to ensure things look good.

This PR includes the XML chunk for modeltime -- see `zchunk_batch_modeltime.xml.R`.


R/utils.R
```diff
@@ -227,10 +228,16 @@ save_chunkdata <- function(chunkdata, write_inputs = FALSE, outputs_dir =
...
   dir.create(outputs_dir, showWarnings = FALSE, recursive = TRUE)
   unlink(file.path(outputs_dir, "*.csv"))
+  dir.create(xml_dir, showWarnings = FALSE, recursive = TRUE)
+  unlink(file.path(xml_dir, "*.xml"))
 
   for(cn in names(chunkdata)) {
     cd <- chunkdata[[cn]]
-    if(!isTRUE(identical(NA, cd))) {   # NA means an optional file that wasn't found
+    if(FLAG_XML %in% get_flags(cd)) {
+      # TODO: worry about absolute paths?
```
rplzzz: What do you mean by "worry" here?  Absolute paths are easy to implement, so I think we should make a decision.  Either the absolute path issue is important enough to deal with now, or we drop the TODO and just deal with it if and when it ever comes up.

R/driver.R
```diff
@@ -25,17 +25,27 @@ run_chunk <- function(chunk, all_data) {
 #' @param chunk_data Data produced by chunk
```
rplzzz: These param documentations should include the type of the data structure.  I assumed a list in some of my other comments, but maybe I'm confused and the actual structure is something else.

R/zchunk_batch_modeltime.xml.R
```diff
@@ -0,0 +1,36 @@
+#' module_modeltime_batch_modeltime.xml
```
rplzzz: Although I'm normally in favor of using `.` in place of `_`, I'm not a huge fan of naming files such that they appear to have two file extensions, such as ".xml.R" in this file.  In a UI that allows you to hide file extensions (itself not a great idea, but also rather common), this will appear as `zchunk_batch_modeltime.xml`, which looks makes it look like a stray XML file.

Maybe it's not really a big deal, but it seems like a potential source of needless confusion.

R/zchunk_batch_modeltime.xml.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' Construct XML data structure for \code{modeltime.xml}.
```
rplzzz: So, the idea is that XML conversion happens in specialized XML chunks, correct?  And every module that produces tables of data destined for XML outputs will have a corresponding XML chunk?

R/zchunk_batch_modeltime.xml.R
```diff
@@ -0,0 +1,36 @@
...
+    # Load required inputs
+    L200.ModelTime <- get_data(all_data, "L200.ModelTime")
+    L200.ModelTimeInterYears <- get_data(all_data, "L200.ModelTimeInterYears")
+
+    # Produce outputs
+    create_xml("modeltime.xml") %>%
+      add_xml_data(L200.ModelTime, "ModelTime") %>%
+      add_xml_data(L200.ModelTimeInterYears, "ModelTimeInterYears") %>%
+      add_precursors("L200.ModelTime", "L200.ModelTimeInterYears") ->
+      modeltime.xml
```
rplzzz: Should this have `FLAG_XML` added to its flags?  If not, then what is the intended use of `FLAG_XML`?

R/utils.R
pralitp: The code here, following the same strategy for outputting data to CSV, is adjusting the XML file path set by the user by some base path.  One reason such a feature is required is because the working directory for the tests is different when just running the package.  So when test cases are running the base path is set to `../../`.

This wouldn't work if the file paths set by the user are absolute.  That being said file paths are typically "modeltime.xml" and this isn't likely going to be a problem at all.  Maybe the best thing to do is just drop the comment and not worry about it.

R/zchunk_batch_modeltime.xml.R
pralitp: It would be a source of confusion and doesn't really add value.  It is also simple to fix so I'll just switch `.xml.R` to '_xml.R` by both renaming this model time example and also so the chuck generator will generate them as such.

R/zchunk_batch_modeltime.xml.R
pralitp: Yes, that is correct.  I know it might not be intuitive at first though that these should be their chunks and not have the Level 2 chunk just generate the XML.  After all in the old data system it _appears_ that is how it was done.

However I ultimately decided to design it like this for a couple of reasons:
1. There are a handful of cases where data is added to an XML file from multiple Level 2 .R files.
2. Fairly commonly the data output of the Level 2 files are in fact used again in some other Level 2 .R file in addition to in the XML file creation.  So we want to make sure we declare those as outputs of Level 2 processing as well instead of just .xml files.

That being said I could be convinced to just collapse some of the simple cases into just a single chunk.

R/zchunk_batch_modeltime.xml.R
pralitp: @bpbond, and I had this discussion.  I suggested that `create_xml` should just add this flag since we would always want to have that flag on objects created by it.  We could always do both since adding the flag twice should hurt anything right?

R/driver.R
```diff
@@ -25,17 +25,27 @@ run_chunk <- function(chunk, all_data) {
...
   # Check that the chunk has provided required data for all objects
   empty_precursors <- TRUE
   for(obj in names(chunk_data)) {
-    assert_that(tibble::is.tibble(chunk_data[[obj]]))
-    for(at in c(ATTR_TITLE, ATTR_UNITS, ATTR_COMMENTS, ATTR_LEGACY_NAME)) {
-      if(is.null(attr(chunk_data[[obj]], at))) {
-        warning("No '", at, "' attached to ", obj, " - chunk ", chunk)
+    obj_flags <- get_flags(chunk_data[[obj]])
+    # Chunks have to returns tibbles, unless they're tagged as being XML
+    if(!outputs_xml[which(obj == promised_outputs)]) {
```
pralitp: I think @bpbond can answer this one better than I can but I think the short answer is: because we don't trust ourselves and so would like some way of testing you said you are going to give us XML (`outputs_xml`) but did you really do that (has flag `FLAG_XML`).

R/driver.R
```diff
@@ -137,7 +148,8 @@ driver <- function(all_data = empty_data(), write_outputs = TRUE, quiet = FALSE,
       if(!quiet) print(paste("- make", format(round(tdiff, 2), nsmall = 2)))
 
       check_chunk_outputs(chunk, chunk_data, input_names,
-                          promised_outputs = subset(chunkoutputs, name == chunk)$output)
+                          promised_outputs = subset(chunkoutputs, name == chunk)$output,
+                          outputs_xml = subset(chunkoutputs, name == chunk)$to_xml)
```
pralitp: No, they are declared at the same time.  It is just that some outputs may be tagged as XML.  I don't think we have an instance where this would happen but a hypothetical example is:
```
  } else if(command == driver.DECLARE_OUTPUTS) {
    return(c("L200.ModelTimeInterYears_adjusted", XML = "modeltime.xml"))
```

R/driver.R
bpbond: Yeah, that's my answer--trying to ensure declarations match outputs. It is redundant, agreed.

R/driver.R
bpbond: I don't know if it'll ever happen, but I think the same was it's useful visually to see the `FILE =` callout, it's also useful to see `XML =`.

R/zchunk_batch_modeltime.xml.R
bpbond: I personally like this split design--it increases the number of chunks, obviously, but the benefit is a lot of flexibility in how we mix-and-match data into the xml files.

R/zchunk_batch_modeltime.xml.R
bpbond: Clarity versus concision. Having `create_xml` add the flag means it's always done there automatically, but not as obvious to anyone scanning the chunk's code. That's another argument for the `XML =` name accompanying the chunk outputs up top, perhaps.

bpbond:
Note to self: when this is merged go revisit closed #357 and put in the xml-generating chunk.

pralitp:
> I'm also a bit unclear on the development workflow. I gather that the XML chunks are generated completely automatically, but how does the system know what XML files need to be generated, which tables go in which files, and so on? I'm especially interested in how this will work in the future. Once the conversion effort is over, how does a user wanting to add new data get all of the necessary chunks added to the system?

Yes, this is a good question.  The XML generating chunks are only generated automatically in so far as the "chunk_generator.h" script can scan the old data system to come up with them.  In most cases we say there is minimal effort for folks to "convert" these chunks because they are so simple.  That being said the chunk generator can easily get fooled (for instance there are times when we loop over a list of data.frames in the old data system and add them to XML which won't get translated properly).  Still the conversion is really simple.  We just need to know:

1. What are we calling the XML file
2. Each table that needs to be converted and the corresponding model interface "header" name that defines the hierarchical relationship of the columns so they may be transformed to XML.

Of course for brand new chunks this won't be generated automatically at all.

R/zchunk_batch_modeltime.xml.R
rplzzz: FWIW, I like the split design too.  Many of these comments reflect the fact that I was figuring out how the system works as I was going through it.

R/zchunk_batch_modeltime.xml.R
rplzzz: That makes sense.  We just need to make sure it's clear so that everybody understands it, even bozos like me.  (I spent a lot of time trying to figure out where the flag got added).

rplzzz:
Re Pralit's last comment, I think that once we get through this conversion stage we'll need to spend a little time writing up some best practices for people adding to the data system.  The key here is to make sure that people know what is expected of them (i.e., how the system is intended to be used) and especially that FNGs starting work on the system for the first time can find that information.

Issue #267, opened 2017-04-18
L251.en_ssp_nonco2


cahartin:
@bpbond

bpbond:
Done.

bpbond:
@CLynchy you may need some `temp-data-inject/` files for this to run.

CLynchy:
Already done.  Pretty sure?

module_emissions_L251.en_ssp_nonco2 <- function(command, ...) {
  if(command == driver.DECLARE_INPUTS) {
    return(c( FILE = "emissions/A_regions",
              FILE = "temp-data-inject/L161.SSP2_EF",
              FILE = "temp-data-inject/L161.SSP15_EF",
              FILE = "temp-data-inject/L161.SSP34_EF",
              FILE = "temp-data-inject/L201.nonghg_steepness"))

bpbond:
Looks good! Then just copy those files from the old data system into `inst/extdata/temp-data-inject/`.

CLynchy:
Someone had already done this (ie put the files needed in the temp-inject folder).  Methinks CH?

cahartin:
not me.  i wonder if another file relied on them as well.

bpbond:
Yes, probably. Anyway, good.

CLynchy:
Level 2 problem:

@bpbond there is a function in the old data system called "add_region_name"
add_region_name <- function( data ) {
     data_new <- data
     data_new["region"] <- GCAM_region_names$region[
          match( data[["GCAM_region_ID"]], GCAM_region_names$GCAM_region_ID ) ]
     return( data_new )
	 }

Regions aren't defined by numbers in Level 2?  They are defined as "string". Is there an input file that I need to load?

bpbond:
@CLynchy It looks like this can be replaced by a simple `left_join_error_no_match` call, with `GCAM_region_names` as an argument. You'll need to go find where that's defined in the old data system.

bpbond:
How's it going? Would it be useful to touch base about this early next week?

CLynchy:
Very sorry Ben.  I have not spent much time last week on the DSR chunk.  I will be making up for it this week.  There isn't a problem/issue that I am working on in the script that has stymied me.  I just need to focus on this chunk for a sustained period of time.

bpbond:
No worries. I was just checking.

CLynchy:
There is a function in the old data system called "rename_SO2".  From the old data system that function does this:
[...]

PR #383, opened 2017-05-09
Add `rename_SO2`

bpbond:
Re query by @CLynchy in #267


bpbond:
@CLynchy note I haven't tried to run this; you may need to tweak later.

bpbond:
@pkyle sorry if that was unclear. I was actually hoping for you to (i) look it over for consistency (as far as you can see) with existing function and (ii) help fill in header information, i.e. Documentation. As Cary asked, what's AWB stand for? What's the broader logic, need, functionality of this? E.g. chunks call this when they need to...

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
```diff
@@ -59,18 +59,39 @@ module_emissions_L101.nonghg_en_USA_S_T_Y <- function(command, ...) {
...
+    # get_data calls followed by temporary (temp-data-inject) reshapes
+    L1322.in_EJ_R_indenergy_F_Yh <- get_data(all_data, "temp-data-inject/L1322.in_EJ_R_indenergy_F_Yh") %>%
+      gather(year, value, -GCAM_region_ID, -sector, -fuel) %>%
+      mutate(year = as.integer(substr(year, 2, 5)))
+
     L1231.in_EJ_R_elec_F_tech_Yh <- get_data(all_data, "L1231.in_EJ_R_elec_F_tech_Yh")
-    L1322.in_EJ_R_indenergy_F_Yh <- get_data(all_data, "temp-data-inject/L1322.in_EJ_R_indenergy_F_Yh")
+
+    # The following two data frames can't be reshaped from long to wide by the test system
+    # So leave them wide, and strip off X's
```
rplzzz: Really?  Why not?

R/module-helpers.R
```diff
@@ -45,5 +45,35 @@ set_water_input_name <- function(water_sector, water_type, water_mapping, GLU =
...
+#' rename_SO2
+#'
+#' Rename SO2 to regional SO2. (TODO)
+#'
+#' @param x Data object (typically from pipeline)
+#' @param so2_map TODO
+#' @param is_awb TODO
+#' @return Data object with \code{Non.CO2} changed to SO2 name for SO2 data.
+#' @importFrom tibble is_tibble
+#' @author BBL May 2017
```
bpbond: That's what I've tagged @pkyle for.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
bpbond: Because there's multiple observations per year and row combination. Simple example: `tibble(y=1:4, year=c(1,2,1,2)) %>% spread(year, y)` (errors).

R/module-helpers.R
```diff
@@ -45,5 +45,35 @@ set_water_input_name <- function(water_sector, water_type, water_mapping, GLU =
...
+  data_so2 <- filter(x, Non.CO2 == paste0("SO2", extension))
+  data_notso2 <- filter(x, Non.CO2 != paste0("SO2", extension))
+
+  so2_map %>%
+    mutate(SO2_name = paste0(SO2_name, extension)) %>%
+    # pull so2_map information into SO2 data
+    select(region, SO2_name) %>%
+    left_join_error_no_match(data_so2, ., by = "region") %>%
+    rename(Non.CO2 = SO2_name) %>%
+    bind_rows(data_notso2)
```
bpbond: I basically never use `split` for some reason, so it didn't occur to me. Don't know if it's better or worse.

R/module-helpers.R
rplzzz: Yep, just wanted to make sure it was captured.

R/module-helpers.R
rplzzz: Many uses of `split` can be replaced by a suitable `group_by`, but when you want to do something to just one group, split can still be convenient.

R/zchunk_L101.nonghg_en_USA_S_T_Y.R
rplzzz: Weird.  What does having multiple entries even mean?  Is there some distinguishing characteristic between them that isn't captured in the table?

bpbond: >some distinguishing characteristic between them that isn't captured in the table

Right. Probably something was stripped off, e.g. region ID, so now we can't reconstruct the wide form any longer.

R/module-helpers.R
CLynchy: so2_map uses information from column  "SO2_name" in A_regions.csv, I believe.  The "is.awb" is about agriculture waste burning, and is also defined by the column "SO2_name" in A_regions.csv, I think.  I don't know how A_regions.csv is created or what all the SO2_names mean.

pkyle:
OK I'll just do some documentation; I'm not going to review the fn without being able to test it, and what it's doing is ultimately pretty simple.
The purpose of this function is to add a suffix to the SO2 gas name indicating which of four major world regions to assign the emissions, as Hector considers the geographic location of sulfur emissions. Any code writing out SO2 emissions data should use this function to rename its "SO2" emissions. Agricultural waste burning emissions already have a suffix assigned (_AWB), so in this case, the SO2 region number is assigned between the "SO2" and "AWB" strings.

bpbond:
@CLynchy I'm going to merge this so you have access to it for testing. @rplzzz if you think the `split` thing is worth it, ping me and I'll follow up with that change. Just no time for it right now.

rplzzz:
@bpbond Nah, the split thing was just FYI.  The real request was for improved docs.

Issue #384, opened 2017-05-10
module_emissions_L102.ghg_en_USA_S_T_Y header info incomplete


bpbond:
```R
#' module_emissions_L102.ghg_en_USA_S_T_Y
#'
#' Briefly describe what this chunk does.
#'
#' @param command API command to execute
#' @param ... other optional parameters, depending on command
#' @return Depends on \code{command}: either a vector of required inputs,
#' a vector of output names, or (if \code{command} is "MAKE") all
#' the generated outputs: \code{L102.ghg_tgej_USA_en_Sepa_F_2005}. The corresponding file in the
#' original data system was \code{L102.ghg_en_USA_S_T_Y.R} (emissions level1).
#' @details Describe in detail what this chunk does.
#' @importFrom assertthat assert_that
#' @importFrom dplyr filter mutate select
#' @importFrom tidyr gather spread
#' @author HCM April 2017
```

@d3y419 can you complete this information (line 3 and and `@details`) and then open a PR please?

PR #385, opened 2017-05-10
description and details added to header

d3y419:
https://github.com/JGCRI/gcamdata/issues/384

added header info


bpbond:
Thanks! Need to rebuild documentation though - the Travis build is going to fail otherwise.

PR #393, opened 2017-05-11
Snyder_LB122_again per suggestion of RPL, added ungroup() to slow pipeline

abigailsnyder:
Speeds things up a ton, still passes Test Package and Check Package


abigailsnyder:
Specifically, chunk LB122.LC_R_Cropland_Yh_GLU

Added `ungroup()` on line 463; only change.

Test and Check fine on my machine at least.

bpbond:
Oh interesting. So before `complete` was operating within groups, so that really slowed things down? How much of a speedup, do you know?

abigailsnyder:
When I would run driver, this chunk usually had a "make" between 7 and 8 (it was that big gross progress bar). Now it's a little less than 1.

abigailsnyder:
quoting RPL on slack dsr channel:  "Putting an `ungroup` at the beginning of the pipeline cuts the total chunk time down from 5840ms to 920ms."

bpbond:
Awesome!

bpbond:
@abigailsnyder maybe add this to the "Hints" page, or somewhere else, on the wiki? Would be good for people to know and understand this.

PR #398, opened 2017-05-12
Fix bad interaction between dplyr NSE and data frame $ indexing.

rplzzz:
Apparently, whether this is a bug or not depends on exactly what package versions you have installed.  I'd say the best practice is to avoid using `$` in arguments to dplyr functions that use NSE.


bpbond:
@rplzzz agreed, we should flag these in our reviews.

Issue #366, opened 2017-05-08
Make an easy combine-old-and-new-data-and-plot function


bpbond:
To help people with debugging.

PR #399, opened 2017-05-12
Simple old-new load and merge debugging routine

bpbond:
Addresses #366


bpbond:
Suggestions welcome about improving this. I just find myself typing the same things (load old dataset; load new dataset; do a join; plot) over and over.

abigailsnyder:
I checked out the branch, did build and reload, did driver() -> x. Then in the console, I'm trying to test with a variety of different files, eg
`oldnew_compare("L122.LC_bm2_R_HarvCropLand_Yh_GLU.csv")`
`oldnew_compare("L133.ag_Cost_75USDkg_C.csv")`
`oldnew_compare("L133.ag_Cost_75USDkg_C")`
`oldnew_compare("L107.an_FeedIO_R_C_Sys_Fd_Y.csv")`


All return the same error:  "Error: is.logical(plot) is not TRUE". So I guess none of the outputs I tried to test were in the right layout for this to work?

bpbond:
No I screwed up while in a hurry. Sorry. Will push new commit in a bit.


On May 12, 2017, at 2:44 PM, abigailsnyder <notifications@github.com<mailto:notifications@github.com>> wrote:


I checked out the branch, did build and reload, did driver() -> x. Then in the console, I'm trying to test with a variety of different files, eg
oldnew_compare("L122.LC_bm2_R_HarvCropLand_Yh_GLU.csv")
oldnew_compare("L133.ag_Cost_75USDkg_C.csv")
oldnew_compare("L133.ag_Cost_75USDkg_C")
oldnew_compare("L107.an_FeedIO_R_C_Sys_Fd_Y.csv")

All return the same error: "Error: is.logical(plot) is not TRUE". So I guess none of the outputs I tried to test were in the right layout for this to work?

bpbond:
I began by having this routine plot results (old vs. new), but took that out at the last minute. Incompletely.

swd-turner:
Is the idea here to have a simple function that we can use to quickly compare the outputs without running the test every time??

bpbond:
@swd-turner basically, yeah. Easy way to get the old and new data for comparison/visualization.

swd-turner:
Couple of thoughts after a quick look...
(1) If the output doesn't have a column named "value" the function won't split it out. So, for example, `oldnew_compare("L133.ag_Cost_75USDkg_C")` contains duplicate rows rather than two columns for old/new.

(2) I think a message would be useful: e.g., 

> no discrepancies found

, or 

> X discrepancies found 

If there are discrepancies then it might be useful to have an option in the function to return a tibble that includes only rows for which old and new differ.

(3) not sure it will be easy to have an a good plot that works for all outputs. The user can probably make necessary plots to compare the data if necessary.

abigailsnyder:
building on Sean's observation, alternative useful messages could be "max difference =...; min difference =..."

swd-turner:
Noticed that `if("value" %in% names(olddata)) olddata <- rename(olddata, value.old = value)` is inside an if statement that requires year columns. So "value" in `olddata` doesn't get caught and renamed in cases like `oldnew_compare("L100.GTAP_LV_milUSD")`.

swd-turner:
Something like `select(value, value.old, all_other_cols)` could be useful to shunt the value columns together and to the left of the tibble. This would help in cases where there are so many columns that the last ones get excluded from the tibble summary (assuming it's mainly the value columns that folks will want to compare immediately).

rplzzz:
> Something like `select(value, value.old, all_other_cols)` could be useful ...

For what it's worth, the easiest way I've found to do this is something like
```
reorder_names <- union(value, value.old, names(my_table)
select(one_of(reorder_names))
```
The first line guarantees that the names you want are at the beginning, and the second one allows you to use that with select without having to figure out how to use `select_`.  Note, however, that if your column names are integers, you may have to protect them.

R/utils.R
```diff
@@ -537,3 +537,62 @@ screen_forbidden <- function(fn) {
...
+    if("value" %in% names(olddata)) olddata <- rename(olddata, value.old = value)
+  }
+
+  # Look for year columns in new data (unusual) and reshape if possible
+  yearcols <- grep("^X[0-9]{4}$", names(newdata))
+  if(length(yearcols)) {
+    cat("Reshaping new data...\n")
+    newdata %>%
+      tidyr::gather_("year", "value", gather_cols = names(newdata)[yearcols]) %>%
+      mutate(year = as.integer(substr(year, 2, 5))) ->
```
rplzzz: This could be made to work on both x-years and integer years by changing the substr call to `sub('^X?([0-9]{4})$', '\\1', year)`.

(It took me like 10 minutes to figure out that the replacement string comes before the match string, the opposite of every regex library anywhere, ever.  Seriously, R?)

R/utils.R
```diff
@@ -537,3 +537,62 @@ screen_forbidden <- function(fn) {
...
+      newdata
+  }
+
+  combined <- list("olddata" = olddata, "newdata" = newdata)
+
+  try({
+    combined <- full_join(newdata, olddata, suffix = c(".old", ""))
```
rplzzz: I would expect this to give the ".old" suffix to the left operand (newdata) and the "" suffix to the right operand (olddata), which is the opposite of what users would expect.  Is that not how the `suffix` argument works?  (The man page is maddeningly vague on this point.)

rplzzz: Another thought here:  without specifying the join columns, `full_join` should join by all columns with matching names.  Doesn't that mean that it's impossible to have "non-joined duplicate columns" for the `suffix` argument to work on?

bpbond:
Doesn't seem like anyone has been clamoring for this capability...close this? Or should I take the trouble to address comments and merge?

rplzzz:
@bpbond I'm indifferent, personally, but then I haven't spent as much time on this project as you guys have.

BTW, I finally turned off the codecov status indicators on the repositories I'm overseeing.  You still get the coverage reports, but it no longer returns a pass or fail status, so your build won't get a red X when the CI actually succeeded.  Here's what you add to your codecov.yml:
```
coverage:
  status:
    project:
      default:
        enabled: no

    patch:
      default:
        enabled: no
```

PR #400, opened 2017-05-12
Move emissions-related mapping files into mapping directory

ssmithClimate:
Move emissions-related mapping files into mapping directory. Remove duplicate file. Update CDIAC mapping file name to more logical name (is not used elsewhere).


ssmithClimate:
@bpbond All the mapping files have consistent line endings on my version (all Mac OS, which I suppose either mean that git converted everything to native MacOS line endings, or that's what they are originally?).

bpbond:
@ssmithClimate it doesn't look like it but no worries, I'll fix. FYI no - git doesn't do any conversion. Thanks!

PR #401, opened 2017-05-13
Various performance optimizations

rplzzz:
There is a lot more that could be done here, if anyone wants to take up the mantle.  This is more of a proof of concept than a complete treatment.

Some general observations on performance:
* Grouped operations are very expensive. Avoid them where possible.  _Leaving the grouping on a tbl_df will often trigger grouped operations_, so make sure to call `ungroup` when you're done with a necessary grouped operation.
* Try to filter data sets as early as you can in the pipeline.  The computational complexity of everything you do scales with the size of the data.
* Cross-products dramatically increase the size of the data, and therefore the complexity of everything you do to it afterward.  If you have to complete a data set, try to do so as late in the process as possible.  In particular, it is better to interpolate missing years on the final product, rather than interpolating at the beginning and carrying the interpolated data through the entire process.
* Recognize common or repeated operations and do them only once, storing the results for reuse, rather than recalculating them.  
* For very large tables (a few hundred thousand rows or more), there is now a `fast_left_join` function that uses a different library to do the join.  (We should probably consider a similar wrapper for large grouped operations, but I haven't done that yet.)

Ben adds:

* Use `replace_na()`, which is 10x as fast as `mutate(x = if_else(is.na(x), ...))`


bpbond:
Run time for the entire driver drops from ~80 to ~70 seconds on my system. Impressive.

@rplzzz note that I had to change lines 118 and 133 in `pipeline-helpers.R` to `as_tibble` to get the package to build cleanly.

R/pipeline-helpers.R
```diff
@@ -68,6 +69,71 @@ left_join_keep_first_only <- function(x, y, by) {
...
+#' @param by Character vector of column names to join by.
+#' @return The left join of \code{left} and \code{right}.  It will be returned
+#' as a \code{tbl_df}, irrespective of the type of the inputs.
+#' @importFrom data.table data.table
+#' @importFrom assertthat assert_that
+#' @importFrom tibble as.tibble
+fast_left_join <- function(left, right, by) {
+    assert_that(is.data.frame(left))
+    assert_that(is.data.frame(right))
+
```
bpbond: `assert_that(is.character(by))` ?

R/pipeline-helpers.R
```diff
@@ -68,6 +69,71 @@ left_join_keep_first_only <- function(x, y, by) {
...
+    ## To key or not to key?  A key is required for the right table, but it is
+    ## optional for the left, *provided* that the join columns are in order and
+    ## come before the non-join columns.  Keying takes time, but it makes the
+    ## join eventually go a little faster.  In the one example we have, it
+    ## keying the left table doesn't seem to pay for itself in the join, but
+    ## it's possible that depends on the specifics of the input.  For now we
+    ## *won't* key, instead opting to reorder the columns of the left table.
+    dtl <- data.table(left[ , union(by, names(left))])
+    dtr <- data.table(right, key=by)
+
```
bpbond: Style consistency - add spaces around `=` pls

rplzzz:
What is the difference between `as.tibble` and `as_tibble`?

bpbond:
@rplzzz None that I know of. But we use `as_tibble` everywhere else.

bpbond:
@rplzzz Added a commit that significantly improves build-and-reload speed, by reducing size of data returned by the driver.

R/zchunk_LA100.FAO_downscale_ctry.R
```diff
@@ -232,111 +234,112 @@ module_aglu_LA100.FAO_downscale_ctry <- function(command, ...) {
...
+    names(L100.FAOlist) <- lapply(names(L100.FAOlist), function(x) {paste0("L100.FAO_", x)})
+                                        # change list names to match the legacy
+                                        # names
+    fixup <- function(irows, legacy.name) {
+        FAO_data_ALL_5yr[irows,] %>%
+          add_comments("Downscale countries; calculate 5-yr averages") %>%
+          add_legacy_name(legacy.name) %>%
+          add_flags(FLAG_NO_XYEAR, FLAG_LONG_YEAR_FORM)
     }
+    L100.FAOlist <- Map(fixup, L100.FAOlist, names(L100.FAOlist))
```
rplzzz: Map can take more than one list as an argument.  In this case it takes both the elements of the list and their names.

rplzzz:
For what it's worth, you can further improve build and reload when you're working on a particular module by doing something like this:
```
driver() -> alldata
alldata[mymodule("DECLARE_INPUTS", alldata)] -> mymodule.data
rm(alldata)
```
Then you can make as many debugging runs as you want using
```
mymodule('MAKE', mymodule.data)
```

bpbond:
Doesn't work. What's `mymodule` supposed to be in above code?

...that said, I realized, in the cold light of day, that my well-intentioned commit of course destroys the ability to re-run a chunk via `chunk("MAKE", all data)`. Will revert it.

rplzzz:
`mymodule` is whatever module you happen to be working on.

rplzzz:
Pretty sure it works:
```
> driver() -> alldata
Found 62 chunks
Found 327 chunk data requirements
Found 200 chunk data products
225 chunk data input(s) not accounted for
...
> alldata[gcamdata:::module_aglu_LA101.ag_FAO_R_C_Y("DECLARE_INPUTS", alldata)] -> la101.data
> la101.out <- gcamdata:::module_aglu_LA101.ag_FAO_R_C_Y('MAKE', la101.data)
> names(la101.out)
[1] "L101.ag_Food_Mt_R_C_Y"   "L101.ag_Food_Pcal_R_C_Y" "L101.ag_kcalg_R_C_Y"     "L101.ag_HA_bm2_R_C_Y"   
[5] "L101.ag_Prod_Mt_R_C_Y"  
```

bpbond:
Oh I see what you mean. Humor me; starting slowly this morning.

R/zchunk_LB120.LC_GIS_R_LTgis_Yh_GLU.R
```diff
@@ -43,25 +43,33 @@ module_aglu_LB120.LC_GIS_R_LTgis_Yh_GLU <- function(command, ...) {
...
+    ## Reset WDPA classification to "Non-protected" where HYDE classification
+    ## is cropland, pasture, or urban land
+    hyde <- land.type$LT_HYDE
+    ltype <- land.type$LT_SAGE_5
+
+    land.type$LT_WDPA <- replace(hyde, hyde != "Unmanaged", "Non-protected")
+
+    land.type$Land_Type <-
+        ltype %>%
+          replace(hyde=='Cropland', 'Cropland') %>%
```
rynacui: So `replace` is more efficient than `mutate` + `if_else`? 
* space around `==`

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
```diff
@@ -51,82 +51,48 @@ module_aglu_LB171.LC_R_Cropland_Yh_GLU_irr <- function(command, ...) {
...
-      left_join(L171.LC_bm2_R_rfdHarvCropLand_C_Yh_GLU, by = c("GCAM_region_ID", "GCAM_commodity", "GLU", "year")) %>%
+    IrrRfdCropland <-
+        L122.LC_bm2_R_HarvCropLand_C_Yh_GLU %>%
+          left_join_error_no_match(L171.ag_irrHA_frac_R_C_GLU,
+                                   by = c("GCAM_region_ID", "GCAM_commodity",
+                                   "GLU")) %>%
+          mutate(irr.harvarea = value * irrHA_frac,
+                 irr.harvarea = replace(irr.harvarea, is.na(irr.harvarea), 0.0), # if no data, assume irrigated is zero
+                 rfd.harvarea = value * rfd_share,
+                 rfd.harvarea = replace(rfd.harvarea, is.na(rfd.harvarea), value)) # if no data, assume all rainfed
```
rynacui: Thanks @rplzzz! It's much more concise avoiding the repeated operations.

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
```diff
@@ -51,82 +51,48 @@ module_aglu_LB171.LC_R_Cropland_Yh_GLU_irr <- function(command, ...) {
...
+
+    ## Extend to cover all years 1700-2010.
+    idvars <- c('GCAM_region_ID', 'GCAM_commodity', 'GLU', 'year')
+    allyr <- seq(min(HISTORICAL_YEARS), max(HISTORICAL_YEARS))
```
rynacui: @rplzzz does this go back to 1971? Does it pass the current `timeshift` test?

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
```diff
@@ -51,82 +51,48 @@ module_aglu_LB171.LC_R_Cropland_Yh_GLU_irr <- function(command, ...) {
...
+    IrrRfdCropland %>%
+      tidyr::expand(nesting(GCAM_region_ID, GCAM_commodity, GLU),
```
rynacui: Just to confirm, `nesting` controls `GCAM_region_ID, GCAM_commodity, GLU` combinations in this tibble, and expands to `allyr` only, correct? Great to know!

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
```diff
@@ -51,82 +51,48 @@ module_aglu_LB171.LC_R_Cropland_Yh_GLU_irr <- function(command, ...) {
...
+                              year = allyr) %>%
+      left_join(IrrRfdCropland, by=idvars) %>%
```
rynacui: space around `=` please

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
```diff
@@ -51,82 +51,48 @@ module_aglu_LB171.LC_R_Cropland_Yh_GLU_irr <- function(command, ...) {
...
+                                             by = idvars) %>%
+      mutate(irr.yld = prod.irr / irr.harvarea,
+             irr.yld = replace(irr.yld, is.na(irr.yld), 0.0),
+             rfd.yld = prod.rfd / rfd.harvarea,
+             rfd.yld = replace(rfd.yld, is.na(rfd.yld), 0.0))
+
+
+    ## Produce outputs.  Note that for harvested area, we use the values without
+    ## interpolation, stored in IrrRfdCropland (as opposed to IrrRfdCropland.interp)
+    select(IrrRfdCropland, GCAM_region_ID, GCAM_commodity, GLU, year,
```
rynacui: How about `IrrRfdCropland %>% select(GCAM_region_ID, GCAM_commodity, GLU, year, value = irr.harvarea)` and below?

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
rplzzz: It goes back to the first of the `HISTORICAL_YEARS`, whatever it happens to be.  It passes the time shift test because the problem was actually in the other table being joined; it needed to be filtered to years that were actually being used.

R/zchunk_LB120.LC_GIS_R_LTgis_Yh_GLU.R
rplzzz: `replace(x, test, newvals)` is slightly more efficient than the equivalent`if_else(test, newvals, x)` because it only replaces the values that need replacing.  In most cases the difference is trivial, but I find the replace syntax a little easier to take in at a glance.  

You can use `replace` in conjunction with mutate.  In this case, since we do a series of replacements I thought the `$<-` notation was a little easier to read.  Also, this might have been early enough in the process that I still thought `mutate` was slower than `$<-`.  (That turned out to be an instance of the `group_by` problem; `mutate` on a grouped table can be very slow, but the problem is easily fixed by ungrouping if the groups aren't relevant to the new column that's being calculated.)

R/zchunk_LB171.LC_R_Cropland_Yh_GLU_irr.R
rplzzz: That's correct.  Basically (and I kind of had to piece this together because the docs for tidyr are a little sparse), `expand` does two things:
1. Interprets names in the context of the table in its first argument (`IrrRfdCropland` in this case).  As far as I can tell, that's the only thing the first argument is used for.
2. Returns as a tbl_df the cross product of its remaining arguments.  Using named arguments allows you to control the names of the columns in the resulting table.  

In this case it has two arguments.  One is `allyr`, which we name as `year`.  The other is `nesting(GCAM_region_ID, GCAM_commodity, GLU)`, which returns the combinations of those three variables that actually occur in the data set.  (As far as I can make out, this is equivalent to `cbind(GCAM_region_ID, GCAM_commodity, GLU)`, except for the class of the result.

This gives us a table with columns `GCAM_region_ID`, `GCAM_commodity`, `GLU`, and `year`.  When we join this to the original data frame (`IrrRfdCropland`), we get back the remaining columns from the original data, with `NA` filled in for combinations that didn't exist before.  I _think_ that this last step is what `complete` is supposed to do, but I've never been able to get it to work right for some reason, so I find it easier just to write it out explicitly.

R/zchunk_LB120.LC_GIS_R_LTgis_Yh_GLU.R
bpbond: FYI I just ran a test on a 1e8 row tibble and `replace_na()` was 10x as fast as `if_else(is.na())`.

bpbond: Replacing all `mutate(x = if_else(is.na(x), 0, x)) %>%` steps in the code with `replace_na(list(x = 0))` drops the driver's execution time **from 70 to 55 seconds**! I'm going to add this commit to the PR.

rynacui: Wow, that's impressive!

bpbond:
@rplzzz go ahead and merge whenever ready. Thanks!

R/zchunk_LB120.LC_GIS_R_LTgis_Yh_GLU.R
rplzzz: That's a little strange.  The problem seems to be with `if_else`, rather than mutate:
```
> x <- runif(1000000)
> df <- tibble(x=x, y=replace(x, x<0.1, NA))
> system.time({mutate(df, y = if_else(is.na(y), -1, y)) -> dfr1})
   user  system elapsed 
  0.196   0.016   0.212 
> system.time({mutate(df, y=replace(y, is.na(y), -1)) -> dfr3})
   user  system elapsed 
  0.005   0.000   0.005 
> system.time({tidyr::replace_na(df, list(y=-1)) -> dfr2})
   user  system elapsed 
  0.005   0.000   0.005 
```
Both `replace` + `mutate` and `replace_na` outperform `if_else` + `mutate` dramatically.  (Note this contradicts what I said above about `if_else` and `replace` having similar performance characteristics -- I must have been testing on vectors that were too small.)  It's hard to understand why `if_else` is behaving so badly.  Maybe I'll open an issue on their repository asking about it.  Note that in light of this, `if_else(condition(x), y, z)` is probably better written as
```
cond <- condition(x)
x <- replace(x, cond, y)
x <- replace(x, !cond, z)
```
Also, `replace_na` has the advantage that it can work on multiple columns with a different replacement value in each column, so that's a big win for clarity when it's applicable.

rplzzz:
Done.

PR #404, opened 2017-05-16
Aaron - LA.118.Hydro

aarony2j:
Please let me know if I should be mixing up the reviewers, so as not to always push these on you guys.

PR #413, opened 2017-05-18
Silence "no visible binding" notes on package check.

rplzzz:
These are caused by the nonstandard evaluation in dplyr functions.
They are harmless, but the deluge of notes can obscure more serious
messages.  We can silence them by creating dummy bindings for any
symbols used only in NSE expressions.

This change set fixes the notes in all currently merged modules.  After this it will be up to each author to ensure that the modules they are committing don't produce these notes.


bpbond:
I was wondering if you were working on this...thanks.

```R
[1] "module_aglu_LA107.an_IMAGE_R_C_Sys_Fd_Y"
 Show Traceback
 
 Rerun with Debug
 Error: invalid subscript type 'NULL' 
```

rplzzz:
Error appears to be on this line (LA107.an_IMAGE_R_C_Sys_Fd_Y.R)
```
mutate(GCAM_region_ID =  left_join_error_no_match(., iso_GCAM_regID, by = c("iso"))$GCAM_region_ID)
```
Why did we even allow this?  Having a join _inside_ a mutate is very weird.  Apparently it interacts in some way I haven't quite figured out yet with the assignments at the top of the file.

bpbond:
Ugh, ugh, ugh. Slipped through.

I've definitely been trying to crack down (e.g. #404 ) more consistently on funky/bizarre stuff like that.

bpbond:
...anyway one of us can fix or happy to tag original author and ask them to do it.

rplzzz:
I'm just going to change the `$` operator and move on.  Your call whether you want to fix the other stuff.

bpbond:
@aarony2j let me draw your attention to this - as you seem to be the primary offender :)  - please don't use `$`inside dplyr functions! It tends to cause problems.

aarony2j:
@bpbond Okay, now I know.

Issue #382, opened 2017-05-09
LA100.IEA_downscale_ctry


bpbond:
Repo needs to become private when this is started--will require commit of proprietary `en_nonOECD.csv` and `en_OECD.csv` files.

...or, use of the OPTIONAL_FILE mechanism.

rplzzz:
I'm not crazy about the idea of committing proprietary data to this repository.  A better strategy would be to use the optional file mechanism and to get the include the IEA files as a submodule (there is a repository on our internal BitBucket server that contains them).  If you don't have access to the internal server, you won't be able to update the submodule; therefore, the files won't be there, and the optional file mechanism will take over from there.

bpbond:
That's fine. Opened this issue b/c I didn't want to start someone on this without thinking about it, then have them commit the proprietary data open to the world.

rplzzz:
Perhaps we should add a pre-receive hook to reject any such attempt.

bpbond:
More trouble than it's worth IMO. This is the only chunk that uses those data; no one will start this chunk until they have the skeleton from me; so I think it'll be OK, barring a malicious proprietary-data-committer.

bpbond:
...that said, no harm in putting it in. The two files are listed above.

rplzzz:
Unfortunately, GitHub doesn't support pre-receive hooks, so the question is moot.

bpbond:
Basically, it looks like this chunk will take three optional inputs: the two listed above, as well as `L100.IEA_en_bal_ctry_hist.csv`. It will:

* If the proprietary data files are available, calculate `L100.IEA_en_bal_ctry_hist` (replicating existing data system logic) and return that to the driver
* If proprietary files are not available, instead return read-in `L100.IEA_en_bal_ctry_hist.csv`, i.e. pre-processed data that is committed into `inst/extdata/energy/`
* Otherwise, error.

bpbond:
Hey @pkyle whenever convenient - can you provide any metadata information (source, units, comments) you know re these files? Thanks.

* `en_nonOECD.csv` and `en_OECD.csv`
* `IEA_product_downscaling.csv`
* `IEA_ctry.csv`

pkyle:
**en_nonOECD.csv**
Title: Energy Balances of non-OECD Countries 1971-2010 (2012 Edition)
Source: International Energy Agency, https://www.iea.org/statistics/relateddatabases/energybalancesofnon-oecdcountries/
[...]

PR #433, opened 2017-05-31
Ben - LA100.IEA_downscale_ctry

bpbond:
This chunk is a bit different from many others, because it
- makes use of the `OPTIONAL_INPUT` mechanism;
- deals with very large input data, so does a lot of computing in 'wide' data format, not using dplyr-style pipelines; and
- can either run through its full computations, producing `L100.IEA_en_bal_ctry_hist`, if the proprietary IEA data area available, or if they're not, it will use a pre-stored version of it. In other words, its output is also an optional input! (See discussion in #382.) This means that the Travis tests will skip most of the chunk code.

unknown-1

Note also that this was failing its oldnew test, even though differences were very small (see below), so I have fallen back to `FLAG_SUM_TEST`.

```R
> summary(newdata)
     FLOW             PRODUCT              iso                year               value           old.value      
 Length:3549080     Length:3549080     Length:3549080     Length:3549080     Min.   :-899691   Min.   :-899691  
 Class :character   Class :character   Class :character   Class :character   1st Qu.:      0   1st Qu.:      0  
 Mode  :character   Mode  :character   Mode  :character   Mode  :character   Median :      0   Median :      0  
                                                                             Mean   :   1353   Mean   :   1353  
                                                                             3rd Qu.:     38   3rd Qu.:     38  
                                                                             Max.   :4354355   Max.   :4354355  
      diff           
 Min.   :-4.155e-05  
 1st Qu.: 0.000e+00  
 Median : 0.000e+00  
 Mean   : 0.000e+00  
 3rd Qu.: 0.000e+00  
 Max.   : 4.863e-05  
```

Finally, note that this PR also incorporates some minor tweaks to the driver and associated support code.


R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+                                 PRODUCT %in% c("Natural Gas", "Other Kerosene", "Total"),
+                                 c("Natural gas", "Other kerosene", "Total of all energy sources"))) ->
+        L100.IEAfull
+
+      # UP FRONT ADJUSTMENTS (UFA) original lines 42-67
+
+      # UFA1. Nearly the entire supply of natural gas in other Africa between 2001 and 2004 is allocated
+      # to GTL plants operating at nearly 100% efficiency. We adjust the energy input quantities
+      # to avoid negative values later on.
+      GTL_COEF <- 1.7
```
skim301: Could we include an additional comment about what 1.7 means and why we chose that? 70% more gas input for about 60% efficiency for GTL? Please comment Page. @pkyle

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        GTL_gas_entries <- with(L100.IEAfull, COUNTRY == "Other Africa" & FLOW == "TGTL" & PRODUCT == "Natural gas")
+        GTL_oh_entries <- with(L100.IEAfull, COUNTRY == "Other Africa" & FLOW == "TGTL" & PRODUCT == "Other hydrocarbons")
+        L100.IEAfull[GTL_gas_entries, GTL_ADJ_YEARS] <- L100.IEAfull[GTL_oh_entries, GTL_ADJ_YEARS] *
+          GTL_COEF * -1     # Multiply by -1 because other hydrocarbons are the output and have a different sign
+      }
+
+      # UFA2. South Africa has a coal-to-gas IO coef in the gas works sector of about 5:1, and low natural
+      # gas consumption in other sectors. The coal inputs are overridden here as the gas output times an
+      # exogenous IO coef
+      COAL_TO_GAS_COEF <- 1.3
```
skim301: So we're changing the coef from 5 to 1.3 to make it a more efficient process right? Is this justified @pkyle?

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        # Note that anthracite is not considered in these regions.
+        # NOTE: using round() to avoid NA's for any values whose base value is >1e6. This seems to work
+        prod <- L100.USSR_Yug$PRODUCT
+        prod_obc <- prod == "Other bituminous coal"
+        prod_hcind <- prod == "Hard coal (if no detail)"
+        prod_cc <- prod == "Coking coal"
+        L100.USSR_Yug[prod_obc, NO_DETAIL_COAL_YEARS] <-
+          L100.USSR_Yug[prod_hcind, NO_DETAIL_COAL_YEARS ] *
+          round(L100.USSR_Yug$`1978`[prod_obc] /
+                  (L100.USSR_Yug$`1978`[prod_cc] + L100.USSR_Yug$`1978`[prod_obc] + 1e-3), digits = 2)
```
skim301: Why do we add 1e-3 here? @pkyle

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: Yes, that's correct; ~60% is a pretty standard efficiency for gas-to-liquids. If we want a citation for that, we can use:
https://www.chevron.com/stories/gas-to-liquids
The input/output of that modern GTL plant resolves to about 1.65.

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: It isn't that we're making it more efficient; what we're trying to do is make the inputs and outputs of the coal-to-gas process consistent with one another. If we want a citation, here are some specs for a modern plant that was never built, but whose IO coef would have been about 1.3:
http://www.bizjournals.com/stlouis/stories/2008/12/15/daily25.html
An IO coef of about 5 on this process indicates that the coal is being used primarily for something else, and to some extent that is to be expected, as it's likely the gaseous fuels produced are a by-product from another activity. But the problem we'd run into if we allowed all of this to be assigned to the coal-to-gas process in GCAM is that there'd be an abrupt change in the economics in going from the calibration year to the future, when suddenly the stuff could be produced for a small fraction of the input fuel, and therefore cost. And for a sector whose capital isn't vintaged, that could cause an abrupt change in the model behavior.

R/zchunk_LA100.IEA_downscale_ctry.R
skim301: I'm fine with that. Lets add a comment that we are adjusting the GTL efficiency to 60% for consistency for these regions.

R/zchunk_LA100.IEA_downscale_ctry.R
skim301: I'm fine with this too. Let's add a comment that the 5:1 ratio is not representative of coal gasification and we are adjusting it here to be consistent with actual conversion efficiencies and also for smooth transition between historical and future technologies in GCAM.

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: Without adding something to the denominator, zero coal consumption (i.e., 0 * 0/0) would return NaN. Adding 1e-3 guards against NaN, and rounding to 2 decimal places keeps the result of the calculation generally consistent with the specificity of the source data (which is rounded to 0 decimal places).

skim301: Maybe we could filter and set prod_cc and prod_obc to 0 when prod_hcind is 0, and do the share without 1e-3 so as not to affect the valid share calculation.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+    L100.Pop_thous_ctry_Yh <- get_data(all_data, "L100.Pop_thous_ctry_Yh")
+    en_OECD <- get_data(all_data, "energy/en_OECD")
+    en_nonOECD <- get_data(all_data, "energy/en_nonOECD")
+    IEA_product_downscaling <- get_data(all_data, "energy/mappings/IEA_product_downscaling")
+    IEA_ctry <- get_data(all_data, "energy/mappings/IEA_ctry")
+    L100.IEA_en_bal_ctry_hist_preprocessed <- get_data(all_data, "energy/L100.IEA_en_bal_ctry_hist")
+
+    # If the (proprietary) raw IEA datasets are available, go through the full computations below
+    # If not, use the pre-saved summary file (i.e., the output of this chunk!) assuming it's available
+    if(!is.na(en_nonOECD) && !is.na(en_OECD)) {
```
rplzzz: `is.na` returns a vector value, and so it is not really the right thing to use in this test.  It will generate a warning when the relevant files are present (or worse, if the first entry in the table happens to be `NA`, it will read as true.  Generally, you want to represent a missing _object_ (as opposed to a missing _value_) with `NULL` and test it with `is.null`.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+    `1990` <- `1990_share` <- `1990_sum` <- COUNTRY <- EcYield_kgm2_hi <- EcYield_kgm2_lo <- FLOW <-
+      GCAM_commodity <- GCAM_region_ID <- GLU <- IEAcomposite <- Irr_Rfd <- LC_bm2_hi <-
+      LC_bm2_lo <- PRODUCT <- composite_population <- iso <- landshare_hi <- landshare_lo <-
+      level <- optional <- value <- year <- yield <- yieldmult_hi <- yieldmult_lo <- NULL # silence package check notes
+
+    all_data <- list(...)[[1]]
+
+    # Load required inputs
+    L100.Pop_thous_ctry_Yh <- get_data(all_data, "L100.Pop_thous_ctry_Yh")
+    en_OECD <- get_data(all_data, "energy/en_OECD")
```
rplzzz: How does `get_data` handle absent optional values?  Right now it looks like not finding a requested table is an error.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+      # parallels that in the original data system, and there are fewer dplyr-style pipelines.
+
+      # Subset only the relevant years and combine OECD with non-OECD
+      hy <- intersect(intersect(HISTORICAL_YEARS, colnames(en_OECD)), colnames(en_nonOECD))
+      cols <- c("COUNTRY", "FLOW", "PRODUCT", hy)
+      bind_rows(en_OECD[cols], en_nonOECD[cols]) %>%
+        # rename fuels with inconsistent naming between the two databases
+        mutate(PRODUCT = replace(PRODUCT,
+                                 PRODUCT %in% c("Natural Gas", "Other Kerosene", "Total"),
+                                 c("Natural gas", "Other kerosene", "Total of all energy sources"))) ->
```
rplzzz: This only works if there is exactly one of each of the values to be replaced, right?

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        L100.IEAfull
+
+      # UP FRONT ADJUSTMENTS (UFA) original lines 42-67
+
+      # UFA1. Nearly the entire supply of natural gas in other Africa between 2001 and 2004 is allocated
+      # to GTL plants operating at nearly 100% efficiency. We adjust the energy input quantities
+      # to avoid negative values later on.
+      GTL_COEF <- 1.7
+      GTL_ADJ_YEARS <- as.character(intersect(2001:2004, HISTORICAL_YEARS))
+      if(length(GTL_ADJ_YEARS)) {
```
rplzzz: Do we encourage using integer expressions in conditional tests (as opposed to, say `length(GTL_ADJ_YEARS)  > 0`)?  (I don't care one way or the other, but I gather it's considered bad style in higher level languages).

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+          GTL_COEF * -1     # Multiply by -1 because other hydrocarbons are the output and have a different sign
+      }
+
+      # UFA2. South Africa has a coal-to-gas IO coef in the gas works sector of about 5:1, and low natural
+      # gas consumption in other sectors. The coal inputs are overridden here as the gas output times an
+      # exogenous IO coef
+      COAL_TO_GAS_COEF <- 1.3
+      CTG_entries <- L100.IEAfull$COUNTRY == "South Africa" & L100.IEAfull$FLOW == "TGASWKS"
+      # Only use other bituminous coal; no need to maintain distinction between coal (if no detail) and other bituminous coal
+      L100.IEAfull[CTG_entries & L100.IEAfull$PRODUCT == "Hard coal (if no detail)", hy] <- 0
```
rplzzz: Is there a reason why we set these to zero instead of removing them from the table entirely?

R/zchunk_LA100.IEA_downscale_ctry.R
rplzzz: That's actually kind of a pain to do.  The best thing would be to use an expression like `max(1e-9,  L100.USSR_Yug$`1978`[prod_cc] + L100.USSR_Yug$`1978`[prod_obc])`.  That will ensure that the denominator is never zero, while leaving (almost) all calculations with nonzero values untouched.  (There is an edge case where 0 < prod_obc < 1e-9).

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+      # Isolate 1990 data for FSU and Yugoslavia; this will be used to back-project individual country shares
+      fsu_yug_composite_entries <- IEA_ctry_composite$IEA_ctry %in% c("Former Soviet Union (if no detail)", "Former Yugoslavia (if no detail)")
+      L100.IEAsingle %>%
+        filter(iso %in% IEA_ctry_composite$iso[fsu_yug_composite_entries]) %>%
+        left_join_keep_first_only(select(IEA_ctry_composite, iso, IEA_ctry), by = "iso") %>%
+        rename(IEAcomposite = IEA_ctry) ->
+        L100.USSR_Yug_ctry
+
+      L100.USSR_Yug_ctry %>%
+        filter(paste(PRODUCT, FLOW) %in% paste(IEA_product_downscaling$PRODUCT, IEA_product_downscaling$FLOW)) ->
```
rplzzz: Wouldn't be better to do this as a `semi_join`?  (It would certainly be clearer; I spent a good minute or two figuring out what this was supposed to do.)

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        L100.USSR_Yug_ctry_FLOW_PRODUCT
+
+      L100.USSR_Yug_ctry_FLOW_PRODUCT %>%
+        group_by(IEAcomposite, FLOW, PRODUCT) %>%
+        summarise(`1990_sum` = sum(`1990`)) %>%
+        ungroup ->
+        L100.USSR_Yug_FLOW_PRODUCT
+
+      L100.USSR_Yug_ctry_FLOW_PRODUCT %>%
```
rplzzz: This pipeline needs a little explanation, especially since it uses a negative select, which adds confusion.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        mutate(`1990_share` = `1990` / `1990_sum`) %>%
+        replace_na(list(`1990_share` = 0)) ->
+        L100.USSR_Yug_ctry_FLOW_PRODUCT_1990
+
+      # Calculate the energy balances of the individual countries during the USSR years as
+      # the total in the composite region times the country-wise shares in 1990 (121-131)
+      POST_USSR_YUG_YEARS_IEA <- intersect(POST_USSR_YUG_YEARS_IEA, names(L100.USSR_Yug_ctry))
+      L100.USSR_Yug_ctry %>%
+        select(one_of("iso", "FLOW", "PRODUCT", "IEAcomposite", POST_USSR_YUG_YEARS_IEA)) %>%
+        left_join_keep_first_only(select(L100.USSR_Yug, one_of("COUNTRY", "FLOW", "PRODUCT", USSR_YUG_YEARS)),
```
rplzzz: Is the `one_of` actually needed here?  Since all of the select columns are string literals, the NSE version should work fine, right?

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+                                  by = c("IEAcomposite" = "COUNTRY", "FLOW", "PRODUCT")) %>%
+        left_join_keep_first_only(select(L100.USSR_Yug_ctry_FLOW_PRODUCT_1990, iso, PRODUCT, `1990_share`),
+                                  by = c("iso", "PRODUCT")) ->
+        L100.USSR_Yug_ctry_bal
+
+      USSR_YUG_COLUMNS <- which(names(L100.USSR_Yug_ctry_bal) %in% USSR_YUG_YEARS)
```
rplzzz: Is `which` actually needed here?  It should be ok to index with a logical, as opposed to the vector of indices returned by `which`.  I think that's what we do in previous instances, so seeing the `which` makes me wonder how or if this one is different.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        rename(IEAcomposite = COUNTRY) ->
+        L100.Others_ctry_bal
+
+      L100.Others_repCtry %>%
+        select(iso) %>%
+        left_join_error_no_match(L100.Others_pop_share, by = "iso") %>%
+        select(-iso) ->
+        pop_share
+
+      hyc <- which(names(L100.Others_ctry_bal) %in% HISTORICAL_YEARS)
```
rplzzz: Same comment as in previous usage of `which`.  When we calculated `hy` (line 59) we didn't seem to need `which`.  Is there something different going on here?  If not, then we should do it the same way in both places.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        add_title("IEA energy balances downscaled to 201 countries by iso code, FLOW, PRODUCT, and historical year") %>%
+        add_units("ktoe and GWh") %>%
+        add_comments("Combine OECD and non-OECD data; perform upfront adjustments for other Africa, Turkey, and South Africa;") %>%
+        add_comments("split out and handle the 1990 split of Yugoslavia and USSR; use population to downscale IEA composite regions") %>%
+        add_comments("to individual countries; filter out countries without data in any year.") ->
+        L100.IEA_en_bal_ctry_hist
+
+    } else {
+      # raw IEA datasets not available, so used presaved data loaded above
+      if(isTRUE(is.na(L100.IEA_en_bal_ctry_hist_preprocessed))) {
```
rplzzz: I don't think `is.na` is the right tool to be using here (see note above).

Also, is `L100.IEA_en_bal_ctry_hist_preprocessed` data that we're allowed to share?  I thought it became ok to share at the point where it was aggregated to GCAM regions.

R/zchunk_LA100.IEA_downscale_ctry.R
```diff
@@ -0,0 +1,308 @@
...
+        add_comments("** DATA PRE-GENERATED; NOT COMPUTED FROM RAW IEA FILES **") ->
+        L100.IEA_en_bal_ctry_hist
+    }
+
+    # Produce final output
+    L100.IEA_en_bal_ctry_hist %>%
+      add_legacy_name("L100.IEA_en_bal_ctry_hist") %>%
+      add_precursors("L100.Pop_thous_ctry_Yh", "energy/en_OECD", "energy/en_nonOECD",
+                     "energy/mappings/IEA_product_downscaling", "energy/mappings/IEA_ctry") %>%
+      # Outputs match the old data system within 10^-5 but still failing, so use sum test
```
rplzzz: That's really weird.  If we can find the time, we really need to look at our comparison code.  Something isn't quite right with it.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Right. What happens is the CSV file reader looks for a requested file and if it's not found either throws an error (most of the time) or puts an NA value into the global data store (if the file is marked optional).

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Yeah, you'll see below I used `if(isTRUE(is.na(...)))` which is I think the correct way to test this, _if_ we use `NA`.

We can use `NULL` instead, but then we'll need to change `get_data` which as you note above currently throws an error in this case.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Hmm. I don't _think_ so, but will check.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Probably clearer to include the `> 0`, agreed.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: @pkyle ?

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: I feel like it wasn't working because of the integer-ish column names, i.e. years; but will check.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: > is L100.IEA_en_bal_ctry_hist_preprocessed data that we're allowed to share?

@skim301 @pkyle who would know the answer to this question?

R/zchunk_LA100.IEA_downscale_ctry.R
rplzzz: Probably the easiest thing to do is to stick a sentinel value in the list, and have `get_data` use that to determine that it doesn't need to throw the error, but then have `get_data` translate the sentinel value to a `NULL`.

Note that `isTRUE(is.na(...))` still has an edge case failure if the first value in its argument happens to be `NA`.

R/zchunk_LA100.IEA_downscale_ctry.R
rplzzz: Example 1 (generates warning):
```
> x <- c(1,2,3,4,5,1)
> replace(x, x %in% c(1,5), c(91,95))
[1] 91  2  3  4 95 91
Warning message:
In replace(x, x %in% c(1, 5), c(91, 95)) :
  number of items to replace is not a multiple of replacement length
```
(Putting the two 1's at the beginning would produce a wrong answer in addition to the warning.)

Example2 (silently gives wrong answer):
```
> x <- c(1,1,2,2,3,3,4,4,5,5)
> replace(x, x %in% c(1,5), c(91,95))
 [1] 91 95  2  2  3  3  4  4 91 95
```

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: Just to clarify, there are no decimals in the data; the minimum non-zero value in the dataset is 1. The current method (adding 1e-3 to the denominator, rounding the result to the nearest 1e-2) does not actually change any values returned as compared with not adding 1e-3. But it is klugey and the max() expression looks better; as long as we continue rounding to the nearest 1e-2, we can change to using the max() expression w/o changing any of the numbers returned. If it's deemed worthy of the effort.

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: At this stage, zeroes are included in the table, so the category is left that way for consistency. It could be removed with no consequences. Of the ~40 million data points in the IEA energy balances, 95% are zeroes.

bpbond: 95% of 40 million points are zero that can be removed with no consequence? That could be a big speed gain. Thanks, will try.

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: The agreement that we have with the IEA is that we can distribute their data aggregated to our regions and sectoral categories. Not sure why it's an issue tho...nobody can run this system without the IEA data (en_OECD and en_nonOECD) anyway, right?
If we're trying to develop a publicly available (i.e., no need for license to IEA data) version of gcamdata, we would have to insert a few data files (L1011.in_EJ_ctry_intlship_TOT_Yh, L1011.en_bal_EJ_R_Si_Fi_Yh, L101.in_EJ_ctry_bld_Fi_Yh, L101.in_EJ_ctry_trn_Fi_Yh), and get a different agreement from the IEA, as several of those files are country-level.

R/zchunk_LA100.IEA_downscale_ctry.R
pkyle: Ha, didn't mean to imply that! This particular row can be removed b/c it's with categories (sub-bituminous coal, other bituminous coal) that are redundant. We CAN absolutely remove zeroes earlier in the processing, but without further downstream checks and modifications, it might wreck later on due to missing/null categories.

R/zchunk_LA100.IEA_downscale_ctry.R
rplzzz: The intent is that you should be able to run the system without the IEA files.  That's why they're optional.

pkyle: Huh, I did not know that! And the planned work-around is to commit some of the data tables written out by our scripts in the inst/extdata folder? Note that doing that would hard-wire our regional and sectoral mappings.

rplzzz: Yes, it would.  The main advantages would be that it would allow users to generate the stock output files, simplifying the process of downloading and installing GCAM, and it would allow them to tinker with things that don't require changing the regional and sectoral mappings.

bpbond: Exactly. I think we can make the requirement that if you're going to mess with regional/sector mappings, you need to have access to the underlying proprietary IEA data.

skim301: I would agree with that too. There's no way to create new gcam regions without having access to the IEA data, and we are not supporting that capability as our open source model.

rplzzz: Sure, I think we all agree on that.  The point I was trying to get across is that there is still value in being able to run the rest of the system, even if you can't change the regional and sectoral mappings.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done, although note that it uses `pmax` not `max`, and is enclosed in an `OLD_DATA_SYSTEM_BEHAVIOR` block, as this change causes tests to fail.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: P.S. for future reference, differences between old and new method are negligible:

```
> summary(combined$diff)
     Min.   1st Qu.    Median      Mean   3rd Qu.      Max. 
0.000e+00 0.000e+00 0.000e+00 1.582e+14 0.000e+00 1.012e+19 
```

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Thanks. Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Good suggestion, thanks. Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Good simplification, thanks. Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: Done.

R/zchunk_LA100.IEA_downscale_ctry.R
bpbond: I've pushed commits addressing all other comments, I think, but this remains unresolved. Are we allowed to commit, i.e. make semi-publicly available, `L100.IEA_en_bal_ctry_hist`, a processed but still country-specific product of the proprietary IEA files?

If not...we either need to make `en_OECD` and `en_nonOECD` required to run the new data system (as now), or push the optional fill-in data 'deeper' into the data system, i.e. to the aggregation step(s).

pkyle: No! At least as per existing agreements. `L100.IEA_en_bal_ctry_hist` is basically the same as committing the IEA energy balances; it's the same dataset, expanded and made more usable. That said, it is five years out of date, so they might allow us to distribute it? Sometimes they make old datasets freely available. But there is no practical difference, for us or for them, between the output of this chunk and its inputs (en_OECD, en_nonOECD).

bpbond: Right. So the current mechanism (in this PR) won't work.

Do we want the ability to run the data system without access to the proprietary files? To me, that seems to have significant value. So I'll change this chunk to return NA, basically, if the IEA files aren't present, and then downstream chunks will be responsible for checking this and substituting their own pre-built aggregated data as necessary.

R/zchunk_LA100.IEA_downscale_ctry.R
rplzzz: There's always something that needs doing on the farm.

Issue #402, opened 2017-05-15
Circle back to L100.Population_downscale_ctry


bpbond:
Lines 104-124 in this chunk reshape the data several times and use hard-coded years to fill in missing values.  It seems like there should be a way to do this more cleanly and robustly.

russellhz:
Is a for-loop OK here? @d3y419 wrote a fairly compact (5-10 lines of code) one that works. I can't think of any other painless way to do it, since we need to first fill in 1990, then use that corrected 1990 value to fill in 1850, etc down to 1700.

bpbond:
Go for it. More important, I think, is to remove the hard-coded years.

PR #436, opened 2017-06-01
H1_Russell-L100.population_downscale_ctry

russellhz:
#402 
Replaced with a for loop and removed all hardcoding.
Also changed the next line to avoid spreading and regathering data.


bpbond:
Thanks @russellhz . Note that runtime of `L100.Population_downscale_ctry` is unchanged by this PR.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -101,45 +102,47 @@ module_socioeconomics_L100.Population_downscale_ctry <- function(command, ...) {
...
-      gather(year, pop_ratio_1950, -iso) %>%
-      mutate(year = as.integer(substr(year, 2, 5)))
     # Need to create matching iso codes for three countries in UN, but not Maddison.
     # Set Serbia and Montenegro ratio equal to Serbia & Montenegro and use Indonesia population ratio for East Timor
     mne_srb_tls <- filter(maddison_hist_ratio, iso %in% c("idn", "scg")) %>%
-      spread(iso, pop_ratio_1950) %>%
-      rename(mne = scg, tls = idn) %>%
-      mutate(srb = mne) %>%
-      gather(iso, pop_ratio_1950, -year)
+      mutate(iso = replace(iso, iso == "idn","tls"), iso = replace(iso, iso == "scg", "mne"))
```
bpbond: Use `change_iso_code`

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -155,7 +158,7 @@ module_socioeconomics_L100.Population_downscale_ctry <- function(command, ...) {
       select(-year) %>%
       full_join(maddison_hist_ratio, by = "iso") %>%
       complete(year, nesting(iso)) %>%  # Completes tibble to include all years for all iso (creates missing values)
-      mutate(pop = if_else(!is.na(pop_ratio_1950), (pop * pop_ratio_1950), 0)) %>% # Set the remaining mismatched (very small) countries to zero
+      mutate(pop = if_else(!is.na(pop_ratio), (pop * pop_ratio), 0)) %>% # Set the remaining mismatched (very small) countries to zero
```
bpbond: Reordering these clauses to `if_else(is.na(), ..., ...)` might be clearer.

R/zchunk_L100.Population_downscale_ctry.R
```diff
@@ -101,45 +102,47 @@ module_socioeconomics_L100.Population_downscale_ctry <- function(command, ...) {
...
+      filter(hist_interp, iso != "world_total") %>%
+      left_join(pop_missing, by = "year")
+
+    maddison_hist_years <- unique(maddison_hist_ratio$year)
+
+    # Fill in missing values for period t using ratio of country to total population in period t+1 times the missing share in period t.
+    for( i in rev(maddison_hist_years)[-1]){
+      maddison_hist_ratio %>%
+        ungroup() %>%
+        mutate(pop2 = if_else(is.na(pop), (lead(pop, n = 1L, order_by = iso)), 0)) %>%
```
d3y419: i know this is my fault. but can we make this better? 
i think pop2 should be previous_pop/
and else should probably be NA. 
0 and pmax was a hack.

bpbond: It seems like this is an algorithm change that should be entered as issue and saved for V2?

d3y419: russell talked me out of it. apparently, this "hack" is indeed the most efficient method. 
my other suggestion involves another if_else statement, with no obvious gain.

bpbond:
Thanks for dealing with this @russellhz

R/zchunk_L100.Population_downscale_ctry.R
d3y419: i'm not entirely sure what happened here. is this resolved?

Issue #419, opened 2017-05-19
LB162.ag_prodchange_R_C_Y_GLU_irr


bpbond:
OK @abigailsnyder @kvcalvin ?

abigailsnyder:
Good for me

abigailsnyder:
@ben  I'm about to head out for the weekend, but I'm incredibly stuck on my current chunk. Namely, I can't figure out how to replicate the behavior of a few lines involving `stats::complete.cases` (line 176 in old DS).


What it does is catch things like the first two rows:
``` 
GCAM_region_ID GCAM_commodity    GLU Irr_Rfd X2010        X2015        X2020        X2025        X2030        X2035        X2040        X2045
1868             15      FiberCrop GLU054     IRR    NA           NA           NA           NA           NA           NA           NA           NA
1869             15      FiberCrop GLU055     IRR     1 -0.055912489 -0.077892089 -0.129449437 -1.000000000          NaN          NaN          NaN
1870             15      FiberCrop GLU070     IRR     1  0.004617643  0.004513432  0.004413821  0.004318511 0.0008694472 0.0008656838 0.0008619529
```

and replace the entire row with some other vector. Because it's catching any element in the row and replacing the entire row, I just can't figure out how to get it to work in long form with `group_by`. Maybe the problem is just getting to it after having spent the entire day on issues with the rest of the code.

bpbond:
```R
L162.ag_YieldRate_R_C_Y_GLU_irr[ !complete.cases(L162.ag_YieldRate_R_C_Y_GLU_irr), X_spec_ag_prod_years ] <-
  L162.defaultYieldRate[
    match( L162.ag_YieldRate_R_C_Y_GLU_irr[[C]][ !complete.cases(L162.ag_YieldRate_R_C_Y_GLU_irr) ],
           L162.defaultYieldRate[[C]] ),
    X_spec_ag_prod_years ]
```

"If any element in the row is missing, replace the entire row with corresponding commodity's values in `L162.defaultYieldRate`." Yes? Off the top of my head (haven't tried this) I guess I'd do something like this:

```R
L162.ag_YieldRate_R_C_Y_GLU_irr %>%
  # isolate the incomplete rows, wipe out their existing data, and pull in default yield rates
  filter(!complete.cases(.)) %>%
  select(GCAM_commodity) %>%
  left_join(L162.ag_YieldRate_R_C_Y_GLU_irr, by = "GCAM_commodity") %>%
[...]

PR #441, opened 2017-06-02
Snyder-LB162.ag_prodchange_r_c_y_glu_irr

abigailsnyder:
https://github.com/JGCRI/gcamdata/issues/419


R/constants.R
```diff
@@ -97,6 +97,7 @@ CONV_MIL_THOUS <- 1000
 CONV_ONES_THOUS <- 0.001
 CONV_TON_MEGATON <- 1e-6
 CONV_T_KG <- 1e3
+CONV_KG_T <- 1e-3
```
bpbond: Maybe just `1 / CONV_T_KG`?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+    L162.agBio_YieldRatio_R_C_Ysy_GLU_irr %>%
+      left_join_error_no_match(timesteps, by = "year") %>%
+      mutate(lagyear = year + timestep)  %>%
+      # There is no lag for SPEC_AG_PROD_YEARS[1] but there is for a year not in SPEC_AG_PROD_YEARS
+      # SPEC_AG_PROD_YEARS[1] gets left alone, so for lagyear = not in SPEC_AG_PROD_YEAR, overwrite
+      # the ratio to be 0.52, the timestep to be 1, and lagyear = SPEC_AG_PROD_YEAR[1]. This allows
+      # the same pipeline to be used for all SPEC_AG_PROD_YEARS
+      mutate(YieldRatio = replace(YieldRatio,
+                                  ! lagyear %in% SPEC_AG_PROD_YEARS,
+                                  0.5),
```
bpbond: Comment above says 0.52, but here it's 0.5?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+      mutate(YieldRate = replace(YieldRate,
+                                 is.na(YieldRate),
+                                 defaultRate)) %>%
+      ungroup() %>%
+      select(-defaultRate) ->
+      L162.ag_YieldRate_R_C_Y_GLU_irr
+
+
+    # Step 4: Write out biomass yields
+    # Essentially repeating steps 1-3 with biomass only.
```
bpbond: It seems like it. Would be it make any sense to put this logic into a function, given how much code is repeated? Just wondering.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
     #
-    # if(OLD_DATA_SYSTEM_BEHAVIOR) {
-    #   ... code that replicates old, incorrect behavior
-    # } else {
-    #   ... new code with a fix
-    # }
+    # CROSIT_ctry identifiers come from the AGLU_ctry table, which needs preprocessing to avoid unwanted
+    # extra rows due to Yugoslav FSR.
+
+    # deal with unfilled Yugoslavia entries in AGLU_ctry without impacting other code chunks.
```
kvcalvin: @bpbond: I feel like we've had to do this step in other chunks. Is there a better solution to the `AGLU_ctry` file?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+    L151.ag_rfdHA_ha_ctry_crop %>%
+      mutate(Irr_Rfd = "RFD") %>%
+      rename(HA = rfdHA) %>%
+      bind_rows(L151.ag_irrHA_ha_ctry_crop) %>%
+      # Join CROSIT country and crop information and aggregate
+      # keeping NA's for later processing
+      left_join(AGLU_ctry_iso_CROSIT, by = "iso") %>%
+      left_join(dplyr::distinct(select(FAO_ag_items_PRODSTAT, CROSIT_crop, GTAP_crop)),
+                               by = "GTAP_crop") ->
+      L162.ag_HA_ha_ctry_crop_irr # agrees to 10e-8
```
kvcalvin: Could you expand on the comment? I don't know what `agrees to 10e-8` means. I'm guessing you are saying this table would fail the oldnew test, but does actually match. If so, say that explicitly.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+
+    # Aggregate the above table of LDS area to CROSIT country and crop levels.
+    # Then, filter to only the country-crop-irrigation combinations present in the CROSIT multiplier
+    # table, L162.ag_Yieldmult_Rcrs_Ccrs_Y_irr.
+    # Repeat the resulting tibble for all SPEC_AG_PROD_YEARS, and join in the yield multipliers.
+    # This results in a table of harvested area and yield multipliers by CROSIT country and crop, glu,
+    # and irrigation for each SPEC_AG_PROD_YEARS year, based on LDS harvested area tables L151....
     #
+    # In this method, compositional shifts of CROSIT commodities within GCAM commodities do not
+    # translate to modified yields. Harvested area multipliers are 1 in all periods.
```
kvcalvin: What does this mean?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
abigailsnyder: Not Ben, but per Page, there are some columns that are fine to update in `AGLU_ctry` because only one chunk uses them.  When I checked in about the CROSIT columns, he indicated they were used by another chunk and I _may_ be better off not risking damaging that. So I erred on the side of caution. 

It's all Yugoslavia, so to me, it would make sense if iso=hrv in Yugoslavia had ALL the same information as iso=hrv in Croatia. But I guess because this issue has been dealt with differently in different chunks at this point, we run the risk of having to go through and tweak each of those.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+      # the full GTAP table,  L162.ag_HA_ha_ctry_crop_irr
+      L162.ag_HA_ha_Rcrs_Ccrs_Ysy_GLU_irr %>%
+        select(CROSIT_ctry, CROSIT_crop, year, Mult) %>%
+        semi_join(L162.ag_HA_ha_ctry_crop_irr, by = c("CROSIT_ctry", "CROSIT_crop")) %>%
+        dplyr::distinct() ->
+        CROSIT_mult
+
+      L162.ag_HA_ha_ctry_crop_irr %>%
+        na.omit() %>%
+        repeat_add_columns(tibble::tibble(year = SPEC_AG_PROD_YEARS)) %>% # agree fine to here; it's the next join
```
kvcalvin: What does this comment mean?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
abigailsnyder: Whoops, this is a leftover comment from tracking my agreement as I translate for use in debugging; I usually catch all of them to remove but I missed this one. I'll go through very carefully to make sure I catch the others

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
+        na.omit() %>%
+        left_join_error_no_match(select(iso_GCAM_regID, iso, GCAM_region_ID), by = "iso") %>%
+        left_join_error_no_match(select(FAO_ag_items_PRODSTAT, GTAP_crop, GCAM_commodity), by = "GTAP_crop") %>%
+        # Multiply base-year harvested area by the future productivity multipliers to calculate prod_mod and aggregate
+        mutate(Prod_mod = HA * Mult)  %>%
+        group_by(GCAM_region_ID, GCAM_commodity, year, GLU, Irr_Rfd) %>%
+        summarise(HA = sum(HA), Prod_mod = sum(Prod_mod)) %>%
+        ungroup() %>%
+        # Calculate YieldRatio = Prod_mod/HA by region-commodity-glu-irrigation-year; subset and output the YieldRatios
+        mutate(YieldRatio = Prod_mod / HA) %>%
```
kvcalvin: Is this really a `YieldRatio` or just a yield?

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
```diff
@@ -41,67 +45,460 @@ module_aglu_LB162.ag_prodchange_R_C_Y_GLU_irr_DISABLED <- function(command, ...)
...
```
kvcalvin: See above.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
abigailsnyder: same debugging comment that I missed. I should start doing them with `##` so I can just use find to make sure I remove all. There are very possibly other similar comments and I will go through line by line to make sure they're all removed

bpbond: :)  I leave similar comments ("GOOD TO HERE BUT THEN WTF?!?") too

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
abigailsnyder: I have always wondered that whenever I've worked with these files (like for the integration branch); I always assumed it was called `YieldRatio` because it used a modified production. @pkyle ?

pkyle: I used YieldRatio whenever the value was the yield the given year divided by the yield in the base year. YieldRate is the annual growth rate applied in the given model time step. The YieldRatio is more versatile because its value is not influenced by the choice of model future years, but the YieldRate is the parameter that is ultimately read in to the model.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
abigailsnyder: Actually, I'm pretty sure the only reason this is done separately in the old DS is because some intermediate products don't have Biomass.

So I think I should be able to streamline this into a single pipeline with maybe one intermediate pipeline ahead of it, and then separate out the ag and bio for outputs at the end. Testing now.

abigailsnyder:
I think all review comments have been addressed @bpbond @kvcalvin

abigailsnyder:
It's error-ing on Travis; I think merging in master deleted some of the inputs it needs. Tracking down now.

abigailsnyder:
Ooh the error on Travis was due to FAO files moving.  Updated and passing Check Package locally at least

bpbond:
Looks great @abigailsnyder - thank you! Holding on @kvcalvin approval.

R/zchunk_LB162.ag_prodchange_R_C_Y_GLU_irr.R
kvcalvin: But, isn't this yield? It is production / harvested area, not ratio of yields from one year to another.

kvcalvin: Never mind. I see. It isn't really production.

abigailsnyder: Yes, I was just about to reply. I updated the comments in the last round I pushed with the following:  "YieldRatio = Prod_mod / HA rather than Yield = Prod / HA.
 And modified Production, Prod_mod = Harvested Area * Yield Multiplier -> Prod_mod is not a physical quantity as production is, and so the YieldRatio is distinct from Yield."

abigailsnyder: though looking now, that's still not the most accurate way to convey the information

kvcalvin: I think what it is is productivity growth weighted by harvested area.

bpbond:
@abigailsnyder nice job! Do you need another chunk assigned?

abigailsnyder:
@bpbond  I do, but I'm also about to head to Vienna for a week (next Sunday-Saturday), so I probably won't start until I get back.

bpbond:
OK! Note - I just noticed that this chunk is actually failing the timeshift test (b/c we still haven't fixed this test, the 'failure' just gets printed and doesn't actually trigger a failure).

```
Error in left_join_error_no_match(., L162.defaultYieldRate, by = c("year",  : 
  left_join_no_match: NA values in new data columns
```

If you get a chance to take a look before you leave, great; otherwise I'll take care of it. Thanks! -B

abigailsnyder:
Oh I didn't realize that was a real error because it was after the DONE line. 
I'll try to take a look at it today or tomorrow before I dive into other stuff

bpbond:
Yeah, sorry, I've got to deal with this non-error-throwing thing asap. Again, if you don't have time, no worries--just let me know. Thanks.

abigailsnyder:
@bpbond  Ok I definitely get why it's failing that test but I'm not quite sure how to fix it. 

It's the `left_join_error_no_match` calls on lines 382 and 391.

What's happening is that externally read in default rates for each model year are read in and interpolated to the model years (pipeline starting on line 322)  and used to fill in any missing data in these two pipelines. So I think it's failing because when the chunk shifts years, that externally defined file 
or the interpolation pipeline does not. 
In practice, I think we would just interpolate that externally defined file to cover yearly  or whatever time steps. But I'm just not sure how exactly to implement it in the pipeline starting on line 322 to pass the time shift test.

pkyle:
Sounds complicated! I know nothing of the specifics here, but wondering how zchunk_LB112.ag_prodchange_R_C_Y_GLU.R didn't have the same problem. In that chunk, the default yield rates are handled like this:
`    A_defaultYieldRate %>%
      select(GCAM_commodity) %>%
      repeat_add_columns(tibble(FUTURE_YEARS)) %>%
      rename(year = FUTURE_YEARS) -> interp.future`
Sorry if that wasn't helpful!

bpbond:
I'm about to rewrite the timeshift anyway, and will (hopefully) fix this as part of that. More later. -B

Issue #111, opened 2017-03-27
LA100.IMAGE_downscale_ctry_yr


bpbond:

bpbond:
Hi @crvernon it looks like this file uses a `downscale_IMAGE_regions` routine. Please check whether it's called elsewhere in the data system; if not, it can be defined locally (in your new file), but if yes, please put it in a separate file--we can talk. Thanks.

bpbond:
@crvernon @kvcalvin Should this chunk be reassigned to someone else? Don't know your (CV) time commitments, but it would be good to get it done one way or another.

crvernon:
@bpbond @kvcalvin Apologies.  I have been swamped.  I will let you know by Wednesday, May 10 COB about getting to this.  If I cannot get a chance to do it by then, I will get you to reassign.  I am interested in helping out on this.  Thanks.

crvernon:
@bpbond @kvcalvin Still working to make time for this.  If this this chunk is a dependency for another where someone is waiting, I would reassign.  If not, I will try to get to it by the end of the week.  Thanks.

bpbond:
OK, I'm reassigning this. Chris, if you get time, please talk to Kate about re-engaging with DSR.

bpbond:
@pkyle any comments on the metadata for `IMAGE_an_Feedfrac_Rimg_C_Sys_Fd_Y.csv`, `IMAGE_an_FeedIO_Rimg_C_Sys_Y.csv`, and `IMAGE_an_Prodmixfrac_Rimg_C_Y.csv`? In particular, units? Here's what there is currently:

```
# File: IMAGE_an_Feedfrac_Rimg_C_S_F_Y.csv
# Title: IMAGE animal feed fractions by IMAGE region / animal type / feeding system / feed type / and year
# Units: NA
# Author: Page Kyle
# Original source: aglu-data/RawData/IMAGE/Agr Production.xls [Agr Production]

# File: IMAGE_an_FeedIO_Rimg_C_S_F_Y.csv
# Title: IMAGE animal feed input-output coefficients by IMAGE region / animal type / feeding system / feed type / and year
# Units: NA
# Author: Page Kyle
# Original source: aglu-data/RawData/IMAGE/Input Fractions.xls [IO Coeffs]

# File: IMAGE_an_Prodmixfrac_Rimg_C_Y.csv
# Title: IMAGE animal production system fractions by IMAGE region / animal type / and year
# Units: NA
# Author: Page Kyle
[...]

PR #443, opened 2017-06-03
Ben - LA100.image_downscale_ctry_yr

bpbond:
#111 

Note that one of the downstream products, `L107.an_FeedIO_R_C_Sys_Fd_Y` now fails the oldnew test. This isn’t a rounding error problem, either; there are some big differences. My best guess is that the LA107 chunk makes assumptions about the ordering of its input data (e.g. perhaps in line 96), as I have verified that the three inputs datasets from `LA100.IMAGE_downscale_ctry_yr` are identical to within 10^-10.

@aarony2j would you have time to circle back to this chunk? If yes, please check out this branch and isolate the problem in LA107.


R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
```diff
@@ -30,61 +29,119 @@ module_aglu_LA100.IMAGE_downscale_ctry_yr_DISABLED <- function(command, ...) {
...
-    # }
-    #
-    #
-    # ===================================================
+    IMAGE_an_Feedfrac_Rimg_C_Sys_Fd_Y <- get_data(all_data, "aglu/IMAGE/IMAGE_an_Feedfrac_Rimg_C_Sys_Fd_Y")
+    IMAGE_an_FeedIO_Rimg_C_Sys_Y <- get_data(all_data, "aglu/IMAGE/IMAGE_an_FeedIO_Rimg_C_Sys_Y")
+    IMAGE_an_Prodmixfrac_Rimg_C_Y <- get_data(all_data, "aglu/IMAGE/IMAGE_an_Prodmixfrac_Rimg_C_Y")
+
+    # Helper function: copy 1970 data to new year 1960
+    extrapolate_1960 <- function(x) {
```
abigailsnyder: Not really extrapolating (or at least way I think of extrapolating) and maybe a more descriptive name, like `extrapolate_1970_to_1960` if you want to stick with extrapolate language or `copy_1970_to_1960` if not. 
Also, if this type of helper function might be more widely useful, it might make sense define externally and make the years an input option, eg
`copy_year1_to_year2 <- function(x, year1, year2){`

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
```diff
@@ -30,61 +29,119 @@ module_aglu_LA100.IMAGE_downscale_ctry_yr_DISABLED <- function(command, ...) {
...
+    # Extrapolate each IMAGE table to all historical years
+    # For each table, same basic steps: extrapolate to 1960; fill out all AGLU
+    # historical years; interpolate.
+
+    IMAGE_an_Feedfrac_Rimg_C_Sys_Fd_Y %>%
+      extrapolate_1960 %>%
+      gather(IMAGE_region_ID, value, -commodity, -system, -input, -year) %>%
+      complete(year = union(AGLU_HISTORICAL_YEARS, year), nesting(commodity, system, input, IMAGE_region_ID)) %>%
+      arrange(year) %>%
+      group_by(commodity, system, input, IMAGE_region_ID) %>%
```
abigailsnyder: do these tables all need to be `ungroup`ed to be safe?

abigailsnyder:
@bpbond  @aarony2j  I'm pretty sure I'm the one who did `LA107`; I'll take a look at it tomorrow when I work on DSR more. I've been trying to use my middle initial since Aaron and I have the same first and last. But that's not helpful if he and I happen to have the same middle as well

bpbond:
Oh jeez, that didn't occur to me! Though by now, you'd think I would recognize people by their coding styles alone...anyway, thank you @abigailsnyder . If you don't have time, I can dive in later this week.

abigailsnyder:
@bpbond I'm just glad we aren't also competing with Adria for AS dominance in DSR

aarony2j:
@bpbond @abigailsnyder Well then, I take that as a compliment that you thought I wrote Abigail's code. I'll be sure to include my middle initial, which I think is different. AJS (mine) vs ACS

bpbond:
@abigailsnyder whoah you're right, we could have had _three_ "AS" coders. Yikes!

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
```diff
@@ -30,61 +29,119 @@ module_aglu_LA100.IMAGE_downscale_ctry_yr_DISABLED <- function(command, ...) {
...
+    # Change AGLU_ctry$IMAGE_region_ID to character for easy joining in helper function
+    AGLU_ctry$IMAGE_region_ID <- as.character(AGLU_ctry$IMAGE_region_ID)
+
+    # Helper function: filter, repeat and add columns, join with iso data
+    downscale_IMAGE_regions <- function(x, AGLU_ctry, by) {
+      hyd <- filter(x, year %in% AGLU_HISTORICAL_YEARS) %>% spread(year, value)
+
+      x %>%
+        ungroup %>%
+        filter(IMAGE_region_ID == 1, ! year %in% AGLU_HISTORICAL_YEARS) %>%
```
kvcalvin: Why region `1`? Or is it irrelevant which you choose?

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
```diff
@@ -30,61 +29,119 @@ module_aglu_LA100.IMAGE_downscale_ctry_yr_DISABLED <- function(command, ...) {
...
+
+
+    # Downscale IMAGE region-level data to all countries
+
+    # Change AGLU_ctry$IMAGE_region_ID to character for easy joining in helper function
+    AGLU_ctry$IMAGE_region_ID <- as.character(AGLU_ctry$IMAGE_region_ID)
+
+    # Helper function: filter, repeat and add columns, join with iso data
+    downscale_IMAGE_regions <- function(x, AGLU_ctry, by) {
+      hyd <- filter(x, year %in% AGLU_HISTORICAL_YEARS) %>% spread(year, value)
```
kvcalvin: Can we make this name more meaningful? I thought `hyd` meant `hydropower`.

rplzzz:
@bpbond @abigailsnyder For what it's worth, `git blame` is a better guide to who wrote a section of code than the Author: line in the comments.  For example, for LA107:
https://github.com/JGCRI/gcamdata/blame/1dce1e44b676145194b2b153433777d5245625c6/R/zchunk_LA107.an_IMAGE_R_C_Sys_Fd_Y.R#L147
Here you can see that the commented-out code left in the source is actually Ben's doing, not Abigail's, as the author comment would have suggested. 😈

abigailsnyder:
@rplzzz that's really cool and probably helpful when I dig into the issues with it today!  Thanks!

rplzzz:
You bet.  It's available from the command line with `git blame` and in emacs using the command `vc-annotate`.  Emacs has a slightly more useful color scheme for indicating the age of a line than github does, IMO.  The command line version doesn't use color at all, so it's not quite so easy to spot newer or older sections.

abigailsnyder:
So I have checked out and run the branch and Just for reference, here's the old-new plot for the output that is now broken. It's a definite problem I am now digging into. 
[L107.an_FeedIO_R_C_Sys_Fd_Y.pdf](https://github.com/JGCRI/gcamdata/files/1055263/L107.an_FeedIO_R_C_Sys_Fd_Y.pdf)

abigailsnyder:
Because it's just the IO and the other two tables (Feed and Prod) are fine, I'm guessing it's an issue in the pipeline from lines 173-182 where Feed and Prod come together to calculate IO. Continuing to dig in, just updating

abigailsnyder:
Vert Line of points at old = 100 shows there are entries in Old DS that are NA and overwrite that are not NA in in New DS; that's definitely not the entire issue but at least part of what's going on

abigailsnyder:
@bpbond  I'm completely stumped. It literally should just be Feed (L107.an_Feed_Mt_R_C_Sys_Fd_Y) divided by Production (L107.an_Prod_Mt_R_C_Sys_Fd_Y). It occurs in the very last pipeline before output.
Both of those output tables match old DS completely,  unless I'm somehow missing a character/level of information (like country, somehow). But I copied and pasted to double check.

I included dummy steps to arrange both of those tables the exact same way, even though the left join should do that. It's literally just dividing two correct numbers. And that somehow gives an incorrect number. 

I corrected the line 96 implicit ordering assumption (no change); I checked for other similar assumptions (none I could see). I even tried reverting the replace_na call in the FeedIO pipeline to a mutate + if_else + is.na call. To no one's surprise, that was not the issue. 

I've pushed what I tried so you can take a look and discard or not. I'm going to work on my current PR and then some other things I have to get done, and hopefully circle back to this tomorrow.  I will dig more into the actual mismatching outputs then and try to identify what the heck is actually going wrong. 


One other test I did was to take the L100 outputs from THIS branch, and use them as the temp-data-inject files to LA107 in another branch (just from master).  That has the same LA107_an_FeedIO issues as this branch. The original temp-data-inject files (as on the master) still give perfect old vs new plots.
So I have compared the BBL L100 output files with the previous temp-data-inject L100 files from master. And they match identically up to formatting.

In other words, master L107 takes wide temp data inject L100 files and gathers to long. FeedIO gives perfect agreement. 
On my test branch, L107 takes quantitatively identical, long temp data inject L100 (sourced from BBL branch outputs). FeedIO fails. 

My head is feeling a little swirly so I don't think I'm describing this last experiment well.
I still have no idea how to fix this, or why dividing two correct numbers gives an incorrect number (in the case of using BBL L100 files anywhere, this branch or my L100 debugging branch). But it seems like it possibly has to do with reading in long versus wide data, or skipping the step to process wide data to long. 


image

bpbond:
Thank you @abigailsnyder - love the picture too! I'm taking a look at this now.

bpbond:
@abigailsnyder thanks again for investigating. Here's what (I think) is going on.

Basically, this is–again–a machine precision/rounding issue. Some of the numbers in the line 186 division (`feedVal / prodVal`) are _very_ small in the denominator, and some of them are being rounded to zero in the old data system in the old one but not in the new. This produces two categories of discrepancies: division results that are a bit off, e.g. 29 versus 27, because the denominator differed by 10^-15 or something; and ones that are a lot off, 29 versus 100 (the latter because the old d.s. `NA` has been replaced by a 100 later in the pipeline).

I'm not sure how we fix this. My thought at the moment is to go back to the old data system and add a rounding step before this computation, produce a new comparison data set, and use that. Then we add a similar rounding step in the new system. Other suggestions? @rplzzz @kvcalvin

abigailsnyder:
That doesn't sound like a bad way to handle it to me. I'm excited you got to digging into it before I got back around to it today.

kvcalvin:
@bpbond: how is this file used downstream? are the changes here going to affect other files later in the system in a meaningful way?

bpbond:
Yeah, from a quick look there are at least five downstream dependencies.

kvcalvin:
So, if we update the old-data file, we would need to update it for the 5+ downstream files as well?

bpbond:
@kvcalvin probably.

bpbond:
Following up on this: so the solution I proposed above (rounding to 10 significant digits before dividing) works, but it does change things. Implementing this in the old data system code produces mostly the same answers, but for 1658 out of 40960 rows, the denominator gets rounded to zero, so the division produces `Inf`, and is then transformed into the default value of 100. Here's the distribution of values across the resulting `L107.an_FeedIO_R_C_Sys_Fd_Y` dataset:

density

But then the new code is able to match these rounded `L107.an_FeedIO_R_C_Sys_Fd_Y` values exactly:

l107 an_feedio_r_c_sys_fd_y_oldnew

So the question is, do we mind that these 4% of rows used to be 0.7 to 514, and now will also receive values of 100? (This is for "Animal production input-output coefficients".) I kind of assume that any number that came out of division of numbers < 10^-10 isn't particularly robust, but don't know.

@kvcalvin @pkyle any thoughts? In brief, the question is do we make a change to an old data system calculation that involves dividing tiny numbers so that the new system is able to replicate?

bpbond:
Also note that `L107.an_Prod_Mt_R_C_Sys_Fd_Y` needs FLAG_SUM_TEST to pass, but matches the old data system's values exactly.

```R
> summary(xall$value-xall$old.value)
-7.105e-14 -1.740e-18  0.000e+00 -4.061e-17  0.000e+00  8.527e-14 
```

pkyle:
This is digging deep! No @bpbond we do not care about these 1e-17 values; the question to me is how they got there in the first place, which in the current data system is the input file IMAGE_an_Feedfrac_Rimg_C_Sys_Fd_Y.csv. That file in turn was constructed from one that we downloaded from the IMAGE 2.4 website (but that is no longer available). In that file, the estimate of the fraction of pastoral beef in China in 1970 that used feed crops as an input was 4.16333634234434E-17. Right down to the mass of the electrons!

bpbond:
>Right down to the mass of the electrons!
😆

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
bpbond: Good suggestion, thanks, done. I don't know whether this occurs elsewhere in the code so leaving it in `zchunk_LA100.IMAGE_downscale_ctry_yr.R` for now.

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
bpbond: Will do.

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
bpbond: @pkyle any comment on this? Thanks.

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
bpbond: I thought it meant hydra 🐉 

Done.

bpbond:
I think all comments have been addressed, except Kate's question about region 1 above, which hopefully @pkyle can answer. Otherwise, holding on OK by @kvcalvin at this point.

pkyle:
@bpbond not sure what question from @kvcalvin about region 1 that you meant? Maybe under a different issue?

R/zchunk_LA100.IMAGE_downscale_ctry_yr.R
pkyle: OK just saw this (I swear this was hidden when I looked an hour ago...). There's no significance of region 1 (which for IMAGE 2.4 is Canada, not the US); it could have been any region.

bpbond: Thanks. I'll add a comment saying this.

rplzzz:
> ... so the division produces Inf, and is then transformed into the default value of 100...
Am I the only one that's troubled by this?  Why is the default value 100?  It appears to be neither the lowest, nor the highest value in the output, nor does it have any other distinguishing feature that makes it a plausible default.  What is the deal with it?

pkyle:
The reason for using 100 is that it is a conservative default IO coefficient for regions without the necessary production data from which to compute one. Tends to be pastoral production in regions with zero pastoral production. If we were to allow this tech in the future (currently it is zero-shareweighted out), we'd need to have something plausible. Might be good to define it in its own object indicating that it's a default IO coef, rather than just hard-wiring 100 into the code, and a more realistic default would be lower and would differentiate dairy from the others.

bpbond:
Thanks @pkyle . I've added this comment to the code and also in a new issue #462 for revisiting later.

bpbond:

bpbond:
Thanks for merging this @rplzzz, although I'm going to miss @abigailsnyder 's "Crazy Pills" pic above.

The problematic `L107.an_FeedIO_R_C_Sys_Fd_Y` is used only by `L202.an_input` (notwithstanding my comment above). I've opened #471 to remind us to be careful with that one.

rplzzz:
That's ok @bpbond, I'm sure there will be more crazy pills in the weeks and months to come.

PR #450, opened 2017-06-08
Check for chunk inputs that don't appear in any outputs

bpbond:
I realized that we weren't checking for chunks that had inputs that weren't declared as precursors for any of their outputs. Added a driver check for this, and test code for the check. This in turn exposed a bunch of mistakes in the chunks' code, so fixed those.


R/zchunk_L125.bcoc_unmgd_R_S_T_Y.R
```diff
@@ -63,14 +67,16 @@ module_emissions_L125.bcoc_unmgd_R_S_T_Y <- function(command, ...) {
 
     # Define function to map from EDGAR to GCAM, and select year 2000,
     # rename the value column by forest fire type, and add a Non.CO2 emission species
-    edgar_to_gcam <- function(x, fire_type, em_name){ x %>%
-      mutate(iso = tolower(Country), Country = NULL) %>%
+    edgar_to_gcam <- function(x, fire_type, em_name) {
+      Country <- GCAM_region_ID <- `2000` <- NULL # silence package check notes
```
rplzzz: This shouldn't be necessary.  Mentioning all of these in the list on lines 39-42 (which `Country` isn't, but the others are) should do the trick because of R's lexical scoping.  My preference would be to have all the note-silencing to happen in one go at the top of the function so that someone reading the code can more easily ignore it.

R/zchunk_L125.bcoc_unmgd_R_S_T_Y.R
```diff
@@ -150,8 +156,8 @@ module_emissions_L125.bcoc_unmgd_R_S_T_Y <- function(command, ...) {
     # calculate deforestation and forest fire emission factors
     # from deforestation and FF drivers (calculated above) and BC and OC emissions
     L125.bcoc_tgbkm2_R_forest_2000_calculations <- left_join_error_no_match(L125.bcoc_tgbkm2_R_forestfire_2000,
-                                                               L125.bcoc_tgbkm2_R_defor_2000,
-                                                               by = c("GCAM_region_ID", "Land_Type")) %>% # combine FF and Deforestation drivers
+                                                                            L125.bcoc_tgbkm2_R_defor_2000,
```
rplzzz: The indentation on the next few lines seems to have gotten garbled.

R/driver.R
```diff
@@ -31,6 +31,7 @@ check_chunk_outputs <- function(chunk, chunk_data, chunk_inputs, promised_output
 
   # Check that the chunk has provided required data for all objects
   empty_precursors <- TRUE
+  pc_all <- c()
```
bpbond: Thanks. I'm punting this to new issue #451

PR #453, opened 2017-06-09
HM&RH-L210.Pop_GDP_scenarios

russellhz:
#325 
All GCAM3 inputs/outputs removed


R/zchunk_L201.Pop_GDP_scenarios.R
```diff
@@ -1,193 +1,173 @@
...
+      lapply(function(df) {select(df, -scenario) %>%
+          add_units("Unitless (annual rate of growth)") %>%
+          add_comments("Per capita GDP growth rate is used for labor productivity growth rate, by scenario") %>%
+          add_precursors("common/GCAM_region_names", "L102.pcgdp_thous90USD_Scen_R_Y")})
+    # Assign each tibble in list
+    for (i in names(L201.pcgdpGrowth_Scen_R_Y_split)){
+      assign(paste0("L201.LaborProductivity_", i), L201.pcgdpGrowth_Scen_R_Y_split[[i]] %>%
+      add_title(paste0("Labor productivity: ", i)) %>%
+      add_legacy_name(paste0("L201.LaborProductivity_", i)))
+    }
```
d3y419: this actually works? so elegant.

R/zchunk_L201.Pop_GDP_scenarios.R
```diff
@@ -1,193 +1,173 @@
...
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L201.BaseGDP_Scen %>%
+      add_title("GDP in base scenario and year") %>%
+      add_units("Million 1990USD") %>%
+      add_comments(paste("Base scenario is", BASE_GDP_SCENARIO)) %>%
+      add_comments(paste("Base year is", min(BASE_YEARS))) %>%
```
d3y419: clever!

R/zchunk_L201.Pop_GDP_scenarios.R
```diff
@@ -1,193 +1,173 @@
...
-      L201.BaseGDP_GCAM3
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L201.LaborForceFillout %>%
+      add_title("Labor force participation and productivity for all scenarios") %>%
+      add_units("Unitless") %>%
+      add_comments("Constant used for all regions") %>%
```
d3y419: shouldn't we clarify that these value are placeholders and have no meaning?

R/zchunk_batch_interest_rate.xml.R
```diff
@@ -0,0 +1,35 @@
...
+    create_xml("interest_rate.xml") %>%
+      add_xml_data(L201.InterestRate, "InterestRate") %>%
+      add_precursors("L201.InterestRate") ->
+      interest_rate.xml
+
+    return_data(interest_rate.xml)
+  } else {
+    stop("Unknown command")
+  }
+}
```
d3y419: there should be a few more of these xml chunks.
batch_socioeconomics_SSPnnn.xml
for each scenario.

bpbond:
Added xml chunks for the `SSP1`...`SSP5` and `gSSP1`...`gSSP5` outputs. I had to construct these by hand, as chunk-generator.R wasn't picking these up (because the original data system code generates them programmatically).

@pralitp Could you look at these files, and run through the full XML-generating process, to confirm they're OK? Thanks.

d3y419:
I can't remember if i wrote the original for-loop. I want to take the full credit regardless. 
I completely agree with Pralit: it's a beautifully elegant piece of coding.

bpbond:
@pralitp I have collapsed the 10 xml-generating chunks into one, duh, thanks, and fixed the problems you notes above. If Pralit approves, are we waiting on a review by @gokuliyer for this, or not?

tests/testthat/test_timeshift.R
```diff
@@ -11,6 +11,8 @@ if(require(mockr, quietly = TRUE, warn.conflicts = FALSE)) {
 
   test_that("chunks handle timeshift", {
 
+    skip_on_travis()
```
rplzzz: If we do this, then we will have to instruct reviewers to run tests manually before approving PRs, since the CI will no longer be running the full suite of tests.

bpbond:
Thanks for the fixes @pralitp

rplzzz:
Hey, did we retain the "skip on travis" thing above when we merged this?  I really don't think we ought to do that if we can help it.

bpbond:
@rplzzz yeah I realized that too and reverted it in 2ef5c8c .

Issue #455, opened 2017-06-12
Need to clearly distinguish the kinds of time series chunks are producing


bpbond:
Right now we conflate two things: historical time series and time series that match `HISTORICAL_YEARS`. Generally we want to produce continuous historical (small h) time series early in the data system, and only filter to the GCAM HY later on. 

For example, currently `L100.GDP_hist` filters its output (historical GDP) to HY, but this causes downstream problems if, for example, our definition of HY changes.

pkyle:
The intention behind `HISTORICAL_YEARS` was to set common years that could potentially be model calibration years in GCAM. Wasn't intended to be all historical years in a general sense. So maybe just a problem of naming and/or documentation? The reason why the Big Bang is in 1971 is that it's the first year that the IEA has stats on developing nations.

bpbond:
Hey @pkyle in your view are these expanded comments correct, in terms of how they define these time constants? cc @abigailsnyder @pralitp 

```R
 # Historical years for level 1 data processing. All chunks are required to 
# fill this period, i.e. producing annual data that covers this entire span.
HISTORICAL_YEARS        <- 1971:2010

 # Future years for level 1 data processing. A few chunks use this to 
# produce future data, e.g. 21st century XXX
FUTURE_YEARS            <- 2011:2100

# Calibrated periods in the model. Only level 2 chunks should reference these
MODEL_BASE_YEARS        <- c(1975, 1990, 2005, 2010)

# Future (i.e. not calibrated) periods in the model.  Only level 2 chunks 
# should reference these
MODEL_FUTURE_YEARS      <- seq(2015, 2100, 5)
MODEL_YEARS             <- c(MODEL_BASE_YEARS, MODEL_FUTURE_YEARS)
```
Are these comment (above) correct, in terms of defining the difference

pkyle:
# Historical years for level 1 data processing. All chunks that produce
# historical data for model calibration are required to cover this entire span.

# Future years for level 1 data processing, for the few chunks that
# produce future data (e.g., population projections)

pkyle:
Ha, that's hard to write comments that start with a `#`! That means something in Markdown!

PR #456, opened 2017-06-13
Better timeshift

### LINKED ISSUE #388: test_timeshift needs to fail correctly!
Currently it doesn't - prints an error message with no context at end of testing. Very confusing.

bpbond:
1. A better timeshift test: one that makes more sense (see #388 ) as it changes `HISTORICAL_YEARS` to 1971-2005, and `FUTURE_YEARS` to 2010, 2015, ...2100. The next step will be a timeshift that rolls history back to 1971-1975, or something like that, but that will involve some trickier steps.

2. It's also better because it will fail with a real, and informative, error; fixes #388 .

3. In a variety of chunks, to pass this new test we now need to handle things more carefully, in particular filtering any `test-data-inject/` data and removing hardcoded years.

4. Previously we filtered the USDA GDP database to `HISTORICAL_YEARS` in the chunks that first handled it, `L100.GDP_hist`. But this conflates two things (cf. #455 ): constructing smooth time series and assigning data to GCAM's historical versus future periods. For example, if we did this and `HISTORICAL_YEARS` no longer overlapped with the IMF GDP projections used by `L102.GDP`, then this latter chunk would fail under timeshift. Now, we pass the full time series and let L102 do its thing; it is responsible for appropriate filtering of its outputs.

5. ~Unfortunately this change means the output of `L100.GDP_hist` is now flagged `NO_TEST`, because it no longer matches that of the old data system. I have created #457 to follow up on this later.~

6. The driver now has `stop_before` and `stop_after` parameters, which are useful for debugging.

7. Cleanup of unneeded comparison files; stylistic tweaks.


rplzzz:
Regarding the no-test flag on the GDP history file, maybe the thing to do here is to update the comparison data to the new correct value.  This is something we are going to run into once the conversion is complete and we move into normal development.  Sometimes a change breaks a test because because we _want_ to change the old behavior.  The thing to do in those cases is to note the failure, validate that it can be explained by the intended change in behavior, and then update the comparison data to reflect the new baseline.

bpbond:
@rplzzz Good point. Thanks.

kvcalvin:
I think this is a good start on this test. This will at least make sure we are moving in the direction of being able to do a hindcast.

R/zchunk_L102.GDP.R
```diff
@@ -181,14 +184,13 @@ module_socioeconomics_L102.GDP <- function(command, ...) {
...
       change_iso_code('rou', 'rom') %>%
       gather(year, gdp.rate, -iso) %>%
-      full_join( gdp_mil90usd_ctry %>% select(iso) %>% unique,
-                 by = 'iso') %>%
+      full_join(gdp_mil90usd_ctry %>% select(iso) %>% unique, by = 'iso') %>%
       mutate(gdp.rate = if_else(gdp.rate == 'n/a', '0', gdp.rate)) %>% # Treat string 'n/a' as missing.
       mutate(year = as.integer(year),
              gdp.rate = as.numeric(gdp.rate)) %>%
-      replace_na(list(year=2010)) %>% # have to do this for `complete` to work as expected.
+      replace_na(list(year = 2010)) %>% # have to do this for `complete` to work as expected.
```
kvcalvin: Do we want to do something more flexible here than hard-coding 2010?

(I know this is unrelated to your changes, just flagging it)

R/zchunk_LA106.ag_an_NetExp_FAO_R_C_Y.R
```diff
@@ -30,6 +30,9 @@ module_aglu_LA106.ag_an_NetExp_FAO_R_C_Y <- function(command, ...) {
 
     all_data <- list(...)[[1]]
 
+    Exp_t <- GCAM_commodity <- GCAM_region_ID <- Imp_t <- curr_table <- element <- item <-
+      netExp <- scaler <- value <- year <- NULL   # silence package check notes
+
```
kvcalvin: Definitely not urgent. But, I wouldn't mind understanding what you are doing here.

bpbond: When you run `R CMD CHECK` (the "check package" option in RStudio) it looks for undefined symbols and prints a NOTE for each. The problem is that dplyr's uses [non-standard evaluation](http://adv-r.had.co.nz/Computing-on-the-language.html). This allows us to have nice clear and clear pipelines without restating `somedataframe$somecolumn` over and over...but the price is that R CMD CHECK can't determine whether the 'bare' column names being used in a pipeline are legitimate or not. As @rplzzz noted on Slack, this can make it hard to see legitimate notes and warnings.

What we're doing here is providing dummy (NULL) definitions for each term that will appear in a pipeline in the following code. This way R CMD CHECK doesn't have a problem with them.

kvcalvin: Thanks!

tests/testthat/test_timeshift.R
```diff
@@ -10,32 +10,41 @@ context("timeshift")
...
+    fyr <- FUTURE_YEARS
+    FUTURE_YEARS <<- seq(2010, 2100, 5)  # normally seq(2015, 2100, 5)
+    byr <- BASE_YEARS
+    BASE_YEARS <<- c(1975, 1990, 2005) # normally (1975, 1990, 2005, 2010)
+    myr <- MODEL_YEARS
+    MODEL_YEARS <<- c(BASE_YEARS, FUTURE_YEARS)
+    ahyr <- AGLU_HISTORICAL_YEARS
+    AGLU_HISTORICAL_YEARS <<- 1971:2005  # normally 1971:2010
+    fhyr <- FAO_HISTORICAL_YEARS
+    FAO_HISTORICAL_YEARS <<- 1961:2011
```
kvcalvin: Why don't these end in 2005 too?

bpbond: Good point. Changed.

R/zchunk_L102.GDP.R
bpbond: I'm going to put this into an issue to follow up on.

bpbond:
Thanks for your comments @rplzzz @kvcalvin . Since this PR introduces a test change that's going to affect all other open PRs and other development work, I'm going to get it merged when Travis tests pass.

rynacui:
Sorry I didn't get a chance to help this week! But I'll take a look when I can to learn the improvements!

R/zchunk_LA154.Transport.R
```diff
@@ -34,12 +34,14 @@ module_gcam.usa_LA154.Transport <- function(command, ...) {
...
-      mutate(year = as.integer(substr(year, 2, 5))) ->
+      mutate(year = as.integer(substr(year, 2, 5))) %>%
+      filter(year %in% HISTORICAL_YEARS) ->   # ensure temp data match our current history
       L154.in_EJ_R_trn_m_sz_tech_F_Yh
 
     get_data(all_data, "temp-data-inject/L154.out_mpkm_R_trn_nonmotor_Yh") %>%
       gather(year, value, -GCAM_region_ID, -mode) %>%
-      mutate(year = as.integer(substr(year, 2, 5))) ->
+      mutate(year = as.integer(substr(year, 2, 5))) %>%
+      filter(year %in% HISTORICAL_YEARS) ->   # ensure temp data match our current history
```
bpbond: @rplzzz note that these comments apply _only_ to the `temp-data-inject` data that we use when an upstream chunk isn't yet available. They and their accompanying `filter`s will go away when the real data become available.

R/zchunk_L102.GDP.R
```diff
@@ -141,18 +143,19 @@ module_socioeconomics_L102.GDP <- function(command, ...) {
     ## Add region IDs to historical GDP and aggregate by region. Hang onto the country-level data
     ## because we will use them again when we make the IMF adjustments
     gdp_mil90usd_ctry <-
-      left_join_error_no_match(L100.gdp_mil90usd_ctry_Yh, iso_region32_lookup,
-                               by = 'iso') %>%
+      left_join_error_no_match(L100.gdp_mil90usd_ctry_Yh, iso_region32_lookup, by = 'iso') %>%
       rename(gdp = value)
+
     gdp_mil90usd_rgn <- gdp_mil90usd_ctry %>%
+      filter(year %in% HISTORICAL_YEARS) %>%
```
bpbond: I think become the regionally-summed data don't get used again until being matched with the `FUTURE_YEARS` data...I assume we're not responsible, here, for guaranteeing that the user's choices of HY and FY are seamless or make sense.

R/zchunk_L102.GDP.R
bpbond: Well now you have #459 to think about it! 🎁

R/zchunk_LA124.heat.R
```diff
@@ -139,7 +142,7 @@ module_energy_LA124.heat <- function(command, ...) {
     # in the zero years. The code below sets an exogenous year as the one to use to extract fuel shares for each region. In all 0 region / years,
     # output is equal to the output in the fuel-share year times 1e-3 to avoid throwing off the energy balances.
 
-    norm_year <- 2010
+    norm_year <- max(HISTORICAL_YEARS) # originally was hardcoded as 2010
```
bpbond: Opened issue #460 to discuss this.

bpbond:
The Debian worker(s) on Travis seem to be having a bad day and stalling until the whole build errors out. Merging, as at least one finished OK.

R/zchunk_L102.GDP.R
rplzzz: Interesting.  I don't think the code should be responsible for validating those choices, but if there's a requirement, then we should at least document it for users.  For example, as it stands, I'd have to go spelunking in the code to tell you whether `HISTORICAL_YEARS` and `FUTURE_YEARS` are supposed to overlap or not allowed to.

R/zchunk_LA154.Transport.R
rplzzz: Ok, I hadn't realized that.

Issue #369, opened 2017-05-08
Column ordering matters when converting to XML


pralitp:
We have thus far been flexible with regards to column ordering, allowing the order to differ from the old data system so long as the content was the same.

Unfortunately converting tables to XML does currently assume some column ordering since the Model Interface "header" maps columns based on column positions ignoring what the actual column names in the table being converted are.

I could see two possible solutions to this.  First is what was essentially done in the old data system: maintain a "level2_data_names" which contains a database of column name ordering as expected in the Model Interface headers as written.  In the old data system we used to re-order columns based off of these as the last step before doing CSV to XML conversions.

An alternative could be to change the behavior of the CSV to XML tool so that it does in fact check the column names in the data to be converted and permute the columns to match columns in the header file as written.  This implies that the column names in our Level 2 tibbles must match the column names as written in the header which may not always be the case at present.

pralitp:
I could imagine @pkyle and @rplzzz may have ideas on this issue.

pkyle:
I don't see any way that the model interface could always correctly guess the column ordering from the strings we're using as column names, at least not unless we make the column names much more verbose. For example, where we currently name a column "year", we'd need to name it "technology/+{year}period", or "subsector/+{year}share-weight", or whatever the correct string is for the given header/column.
What I'd think of doing if we're looking to automate this is to programmatically translate the headers in ModelInterface_headers.txt into column names. I'd imagine we're almost always consistent there (e.g., the column name for "world/+{name}region" is always "region"). One place where I know we'd have exceptions is the CalOutput and CalInput tables, where subsector and technology share-weights are read using different strings than what is in the header and XML file ("subs.share.weight" and "tech.share.weight", because we don't want two columns with the same name). Of course doing this programatically might take more time than just keying them in manually...

rplzzz:
Keeping a database of expected column orderings seems like the lest bad solution in this case.  It should be easy to supply a utility subroutine to reorder the columns according to what's in the DB.

In the long run my concern with all of this is that anybody attempting to add a feature to GCAM is going to have a very hard time figuring out how to get the XML conversion to work.  For purposes of getting the conversion working we can just let it go, but at some point I think we are going to want to do some thinking about how we specify our input database schema.

pralitp:
Yea, I didn't fully appreciate all of the cases where trying to match column names would be difficult as Page points out.

[...]

PR #458, opened 2017-06-14
Xml column ordering

pralitp:
This PR addresses issue #369 by adding level 2 data names from the old data system and modifying `add_xml_data` to automatically look these up and reorder columns before doing the XML to CSV conversion.

In addition we more utilities to deal the special cases of LandNode rename and EQUIV_TABLEs by adding a couple of specialized add_xml_data and updating the chunk generator accordingly.


R/xml.R
```diff
@@ -0,0 +1,576 @@
...
+
+  curr_table <- list(data = data, header = header)
+  dot$data_tables[[length(dot$data_tables)+1]] <- curr_table
+
+  dot
+}
+
+# Note: we have put the definition of run_xml_conversion inside of the "closure"
+# make_run_xml_conversion so that we may stash the XML_WARNING_GIVEN flag in an
+# environment that can only be reached by run_xml_conversion.
```
rplzzz: Why do we need to do this?  (It's obvious if you read the code, but it's a little mysterious if you're just reading the comment.)

R/xml.R
```diff
@@ -0,0 +1,576 @@
...
+  }
+}
+
+#' run_xml_conversion
+#'
+#' Run the CSV to XML conversion using the model interface tool.  This method
+#' should be the final call in a pipeline started with \code{\link{create_xml}}
+#' and one or more calls to \code{\link{add_xml_data}}.
+#'
+#' Not that this method relies on Java to run the conversion.  To avoid errors for
```
rplzzz: "Not" -> "Note"

R/xml.R
```diff
@@ -0,0 +1,576 @@
...
+#' generate_level2_data_names
+#'
+#' Generate the list of "level 2 data names" or really a list of column orderings
+#' keyed by the ModelInterface header so that we can ensure tables being sent to be
+#' converted to XML by the ModelInterface have their columns arranged in the order
+#' the ModelInterface is expecting them.
+#' @return A list object where [[header]] <- character vector of column names
+#' corresponding to header.
+#' @author Pralit Patel
+generate_level2_data_names <- function() {
```
rplzzz: This shouldn't be a function.  Create the data structure in your workspace and add it to the `data/` directory using `devtools::use_data()`.  This could probably be made an internal data structure.  Don't forget to add a data.R function to document the data structure, as explained in http://r-pkgs.had.co.nz/data.html#documenting-data

Also, see the notes on reproducing the data set in http://r-pkgs.had.co.nz/data.html#data-extdata .  Briefly, this function should go in a source file in 'data-raw/'.  It's not part of the package, but it's still in the repository so that someone maintaining the package (i.e., adding or deleting names) can rerun it and generate a new version of the data.

pralitp:
I've addressed most of the changes requested by @rplzzz with the exception to those on the `chunk-generator.R`.  To be perfectly honest I think the answer to many of your questions was "I made these changes to get through the day and avoid breaking Ben's code".  I'm not sure how much effort we want to spend redesigning this code as presumably it should be a throw away at some point soon.  Perhaps @bpbond would like to comment on that.

In addition to the comments with regards to writing documentation on how the XML conversion and model interface headers work:  I didn't think writing that in the code comments was the appropriate place so I write it in a .md file (attached but renamed .txt because Github doesn't like it otherwise) with the intentions of adding it to the gcamdata wiki.  Does that sounds reasonable?  If so I can push this up to the wiki once this PR is approved.

[XML-Conversion.txt](https://github.com/JGCRI/gcamdata/files/1110157/XML-Conversion.txt)

rplzzz:
The comments on `chunk-generator.R` were just comments, not actual requests for changes.  I agree that in the long run the chunk generator will probably be thrown away, so it's not important to get the best code there.  I still think there's a role in code reviews for observations about how to do things better in the future.  I apologize if I ruffled any feathers.

Mostly I was concerned about the documentation issues.  This part of the process is more mysterious than some of the straightforward data wrangling.  I agree, however, that much of that extra documentation belongs in the wiki.  As long as we don't forget about it, I'm ok with not letting it hold up the merge.

R/utils.R
```diff
@@ -399,116 +399,6 @@ chunk_outputs <- function(chunks = find_chunks()$name) {
   dplyr::bind_rows(chunkoutputs)
 }
 
-#' create_xml
```
bpbond: Would it make sense to move these functions into `xml.R`?

R/xml.R
```diff
@@ -0,0 +1,183 @@
...
+add_rename_landnode_xml <- function(dot) {
+  land_name_table <- tibble(from=paste0("LandNode", seq(1,5)),to="LandNode")
+
+  add_xml_data(dot, land_name_table, "NodeRename", NULL)
+}
+
+#' add_node_equiv_xml
+#'
+#' Add a table to an XML pipeline that instructs the ModelInterface to treat
+#' tags in the same class as the same when enabled.  Thus not requiring multiple
```
bpbond: "when enabled, thus not requiring"

R/xml.R
```diff
@@ -0,0 +1,183 @@
...
+#' headers for instance to read in a share-weight into a technology
+#' intermittent-technology or tranTechnology.  A user taking advantage of this
+#' feature would then read this table early in the XML pipeline.
+#' @param dot The current state of the pipeline started from \code{create_xml}
+#' @param equiv_class A name of the equivalence class to add.  This could be any
+#' key in the list \code{XML_NODE_EQUIV}.
+#' @return A "data structure" to hold the various parts needed to run the model
+#' interface CSV to XML conversion.
+#' @author Pralit Patel
+#' @export
```
bpbond: Will this (and others) ever be reasonable called from command line? If not, let's not export it.

pralitp: Yes, I foresee users will need to make "one off" XML files from time most frequently to add some policy lever into some technology.  They could grab the tables necessary to define the structure out of the results from gcamdata, add in their levers,  and use these XML methods to create the XML input file and write it in some R script so it is reproducible.
Although I realize this is likely just pure fantasy and what really happens is they take the CSV results from gcamdata, load it into excel and do the step manually to create the XML files in which case exporting the functions isn't necessary.
I'd prefer to at least make it as easy as possible to do it in a reproducible way.

R/utils.R
pralitp: Well I thought it made sense to move `create_xml` and the others into it's own `xml.R` the XML specific stuff was ballooning.  Although after moving the LEAVEL2_DATA_NAMES into a data object it isn't so bad.  I guess I don't have a strong opinion on moving it back.

pralitp:
@bpbond, yes the `xml-column-names-aglu.R` are duplicative with `LEVEL2_DATA_NAMES` I created.  I can drop those and change the code files that use them to use `LEVEL2_DATA_NAMES` instead.

bpbond:
Thanks @pralitp !

Issue #452, opened 2017-06-08
L203.demand_input


bpbond:
This has a _lot_ of lines of code but I think–hope–that there's a lot of repetition which will help. There's also some funkiness in the output section, looks like; we can touch base as necessary.

Thanks @rynacui !

rynacui:
@bpbond This chunk involves a GCAM defined function `get_logit_fn_tables`. How would you like me to handle it?

bpbond:
Wow, it looks like that's used all over the place. We'll probably want to translate and put it into `module-helpers.R` to make it broadly available.

rynacui:
Got it. I'll give it a try. 
Looks like `write_to_all_regions` is in a similar situation.

rynacui:
I suggest I do it in a separate branch and PR.

bpbond:
Sounds good. Thanks! I can also handle this if you want--just let me know.

rynacui:
Let's see how it goes. I'll let you know!

rynacui:
Hi @bpbond, I just took a closer look, and I have to admit it will be much more efficient to let you handle the functions. More complicated than I thought...

For `write_to_all_regions`, I think I can simply use `repeat_add_columns` instead, but it also seems doing sth else (although not applied here).

bpbond:
>More complicated than I thought

No kidding. Working on this but it will take a few days.

bpbond:
Ryna, re the just-merged functions in #464, I _think_ they're operating correctly, but remember they're brand new. "Trust but verify."

rynacui:
Got it. I'll be careful. Thanks!

rynacui:
Hi @bpbond, I'm working through this chunk and towards the end, there are two functions that are commonly used for level two: `write_mi_data` and `insert_file_into_batchxml`. Would you mind taking care of them as well? Thanks!

bpbond:
Don't worry about these. We now use a different mechanism for all that. Your chunk can just return the datasets to the driver.

rynacui:
woa, I see. Fantastic!

rynacui:
Okay, we are now getting to the outputs. It seems `L203.SectorLogitTables[[curr_table]]` and `L203.SubsectorLogitTables[[curr_table]]` is not recognized as tibble. 

Maybe I should specify each of the outputs such as `L203.SubsectorLogit_relative-cost-logit`?

bpbond:
[...]

PR #464, opened 2017-06-19
`get_logit_fn_tables` and other support functions

bpbond:
These functions are needed for `L203.demand_input` ( #452 ). Basically straight translations of the originals, although I've tweaked parameters names and ordering, and provided assertions and test code.

@rynacui - please review.

@pkyle - I understand technically what's going on in these, but not always what the rationale is, and so the documentation for all these functions needs improvement so that it clearly states _why_ they do what they do. Can you look through and, either in comments or by pushing to this PR, help? Thanks.


bpbond:
Thanks @rynacui . Since you're holding on these functions, I'm going to go ahead and merge, and open an issue for @pkyle and/or me to bring their documentation up to par later.

R/module-helpers.R
```diff
@@ -85,3 +85,171 @@ rename_SO2 <- function(x, so2_map, is_awb = FALSE) {
...
+  }
+
+  # Loop through each of the logit types and create a table for it
+  # using the appropriate header name and writing to all regions if requested.
+  for(curr_logit_type in gcam.LOGIT_TYPES) {
+    tables[[curr_logit_type]]$header <- paste0(base_header, curr_logit_type)
+    currdata <- data[data$logit.type == curr_logit_type,]
+    if(write_all_regions) {
+      if(nrow(currdata) > 0) {
+        currdata <- write_to_all_regions(currdata, names, ...)
```
rynacui: `write_to_all_regions` needs to specify `GCAM_region_names`. When use separately, `write_to_all_regions(data, names, GCAM_region_names = GCAM_region_names)`, first is the parameter and second is the tibble. But here the tibble is not specified, generates error: `argument "GCAM_region_names" is missing, with no default`.

bpbond: Ah, sorry. I'll fix.

bpbond: Opened #467

Issue #465, opened 2017-06-19
Documentation for `get_logit_fn_tables` and other support functions


bpbond:
The documentation for the following functions in `module-helpers.r` needs to be filled out and/or improved/clarified:

* `get_logit_fn_tables`
* `write_to_all_regions`
* `set_traded_names`
* `set_years`

@pkyle when you have time, can you go through their roxygen headers and code comments, succinctly clarifying/adding where necessary, and then open a PR? Thanks.

See also #464

pkyle:
About to open the PR. @bpbond did you test whether get_logit_fn_tables works when write_all_regions=F? I vaguely remember that every time this function is called, that value is set to T, but that if it were set to F, the function would fail.

PR #466, opened 2017-06-19
clarification of description of four functions (get_logit_fn_tables, …

pkyle:
…set_traded_names, set_years, write_to_all_regions)

Addresses #465


bpbond:
Whoah, that's a weird Travis error. Let's try restarting it.

PR #467, opened 2017-06-20
Fix calls to write_to_all_regions from get_logit_fn_tables

bpbond:
Noted by @rynacui


bpbond:
@rynacui merging this

PR #469, opened 2017-06-22
Fix hang during coverage testing

rplzzz:
The after_success command to run the coverage test now uses travis_wait to generate a heartbeat while the coverage test runs.  By default this will keep the coverage test alive for up to 20 minutes.  This limit can be increased if necessary by giving an additional argument to travis_wait.


rplzzz:
@bpbond This should allow us to leave the time shift test in the test configuration, at least until it starts taking more than 10 minutes to run the tests in the normal package check (the coverage test is quite a bit slower than the normal test run.)

bpbond:
Nice. Thanks for looking into this, Robert.

PR #481, opened 2017-06-26
Snyder - LB163 bio yield r glu irr

abigailsnyder:
@pkyle do I have the right Wullschleger citation for the  MAX_BIO_YIELD_THA in constants.R?

#470

Issue #164, opened 2017-03-31
LB113.bio_Yield_R_GLU


bpbond:
Sorry I messed up and made this your "in progress" (not the "up next" one). If not a problem, let's keep it this way please. Thanks.  cc @kvcalvin

bpbond:
```
#Bioenergy yields are equal to this region/zone-specific index multiplied by a base yield
# The base yield is taken to be the maximum of the yields in the USA region, or the region containing the USA,
# because the Wullschleger paper from which the yield estimate was derived was for the USA
```

@pkyle do you have a reference for this paper please?

pkyle:
@abigailsnyder found it and put it into a pull request; not sure if this is still needed but here it is anyway, in full:
Wullschleger, S.D., E.B. Davis, M.E. Borsuk, C.A. Gunderson, and L.R. Lynd. 2010. Biomass production in switchgrass across the United States: database description and determinants of yield. Agronomy Journal 102: 1158-1168. doi:10.2134/agronj2010.0087

bpbond:
Got it, thanks. I just saw her question too--funny.

PR #486, opened 2017-06-27
Ben - LB113.bio_Yield_R_GLU

bpbond:
#164 

Note this and #481 have overlapping definitions of `aglu.MAX_BIO_YIELD_THA` and `aglu.BIO_GJT`.


R/zchunk_LB113.bio_Yield_R_GLU.R
```diff
@@ -25,53 +29,77 @@ module_aglu_LB113.bio_Yield_R_GLU_DISABLED <- function(command, ...) {
...
+    # Calculate each region / zone / crop's comparative yield (41-50)
+    L100.LDS_ag_HA_ha %>%
+      rename(HA = value) %>%
+      left_join_error_no_match(L100.LDS_ag_prod_t, by = c("iso", "GLU", "GTAP_crop")) %>%
+      mutate(Yield = value / HA) %>%
+      # Drop the missing values, where the harvested area was above the min threshold but production was not
+      na.omit ->
+      LDS_ag_Yield_tha
+
+    # Match in the global avg yield for each crop, and compute the ratio from that yield (52-61)
```
abigailsnyder: also aggregating HA and Ratio_weight to GCAM region in this pipeline. might be worth a brief additional comment.

R/zchunk_LB113.bio_Yield_R_GLU.R
```diff
@@ -25,53 +29,77 @@ module_aglu_LB113.bio_Yield_R_GLU_DISABLED <- function(command, ...) {
...
+
+    # Bioenergy yields are equal to this region/zone-specific index multiplied by a base yield
+    # The base yield is taken to be the maximum of the yields in the USA region, or the region
+    # containing the USA, because the Wullschleger paper (10.2134/agronj2010.0087)
+    # from which the yield estimate was derived was for the USA.
+    iso_GCAM_regID %>%
+      filter(iso == "usa") %>%
+      .[["GCAM_region_ID"]] ->
+      USAreg
+    L113.base_bio_yield_tha <- aglu.MAX_BIO_YIELD_THA / max(L113.YieldIndex_R_GLU$YieldIndex[L113.YieldIndex_R_GLU$GCAM_region_ID == USAreg])
```
abigailsnyder: I did this and the next line as pipeline style; it ends up making a 1x1 dataframe, which is a little silly. The call to this value is then dataframeName$columnName. 

I don't know that I have a preference for how to do it and am happy to change LB163, but I think we should decide and be consistent between our chunks

R/zchunk_LB113.bio_Yield_R_GLU.R
bpbond: Done, thanks.

R/zchunk_LB113.bio_Yield_R_GLU.R
bpbond: I don't have a strong preference. I guess I'd say this style, because dplyr 0.7 apparently has a new `pull` function for this, and it'll be easy to replace the `.[[` construct then.

bpbond:
>we may have an opportunity to make more efficient

Opened #487 to remind us about this.

bpbond:
Thoughts @pkyle ?

pkyle:
Since we're no longer maintaining the capacity to run the model without irrigation, the XML files constructed from the data in LB113 aren't relevant anymore. The one possible problem with cutting this file out entirely is that the L201 and L205 code chunks depend on it, and while those are over-written later on, it's possible that their use of the output of LB113 is still somehow in the pipeline. Unlikely, but worth keeping in mind before just removing it.

bpbond:
I guess I'd prefer to merge this, check it off, and then (see #487 ) come back to this question.

bpbond:
Otherwise, waiting on approvals from @kvcalvin and @abigailsnyder

R/constants.R
```diff
@@ -106,10 +106,7 @@ aglu.FERT_COST <- 363 # 2007$ per ton NH3
 # Source: Wullschleger doi:10.2134/agronj2010.0087
 aglu.MAX_BIO_YIELD_THA <- 20
 
-aglu.BIO_ENERGY_CONTENT_GJT <- 17.5  # Energy content of biomass
-
-
-# Energy constants ======================================================================
+aglu.BIO_ENERGY_CONTENT_GJT <- 17.5  # Energy content of biomass, GJ/ton
```
kvcalvin: Does this mean I won?

bpbond: 🏅

Issue #39, opened 2017-03-16
Some GCAM inputs have bad line endings.


rplzzz:
For example: [A12.U_curves](https://github.com/JGCRI/gcam-data-system-OLD/blob/master/energy-data/assumptions/A12.U_curves.csv).  You can't see it in the github interface, but this file has <CR> line endings (like you would find in Mac OS 9 and previous).  It also doesn't have a final newline. Where possible, when we encounter these we should convert to unix line endings (<LF>) and a final newline.

bpbond:
When we're all done with the first translation effort, a simple script should be able to scan all the input files, read, and re-write them with standardized line endings.

bpbond:
Note we may want to re-compress files using a standardized modern algorithm while we're at it. See discussion in #490 .

bpbond:
Addressed by #860 .

PR #490, opened 2017-06-27
Compress files

bpbond:
Shrinks `inst/` from 129 MB to 78 MB.


rplzzz:
@bpbond , are you able to read these compressed files?  When I try I get an error:
```
> l100area <- read.csv('L100.Land_type_area_ha.csv.zip')
Error in read.table(file = file, header = header, sep = sep, quote = quote,  : 
  duplicate 'row.names' are not allowed
In addition: Warning messages:
1: In read.table(file = file, header = header, sep = sep, quote = quote,  :
  line 1 appears to contain embedded nulls
2: In read.table(file = file, header = header, sep = sep, quote = quote,  :
  line 3 appears to contain embedded nulls
3: In read.table(file = file, header = header, sep = sep, quote = quote,  :
  line 4 appears to contain embedded nulls
4: In scan(file = file, what = what, sep = sep, quote = quote, dec = dec,  :
  embedded nul(s) found in input
```
Reading the uncompressed csv file works as expected, so I'm wondering if something got messed up in the compression process.  Also, there appears to be some Mac junk folded in with the data, which may or may not be related:
```
% unzip L100.Land_type_area_ha.csv.zip 
Archive:  L100.Land_type_area_ha.csv.zip
  inflating: L100.Land_type_area_ha.csv  
   creating: __MACOSX/
  inflating: __MACOSX/._L100.Land_type_area_ha.csv  
```

Finally, R supports xz compression, which looks like it creates smaller files than zip, which is kind of an archaic algorithm:
```
% du -h L100*
 18M	L100.Land_type_area_ha.csv
3.9M	L100.Land_type_area_ha.csv.zip
2.4M	L100.whatever.csv.xz
```

bpbond:
Interesting. For me, `readr::read_csv` reads that file fine, but `read.csv` does not.

I agree about using zip--I've been doing just because it seems to be what everyone, especially on the Windows side, has access to and is expecting. xz or gz would be great to shift to though.

rplzzz:
The point is that the user doesn't ever have to uncompress the files.  R has the ability to read from compressed connections transparently.

`read_csv` worked for me once I remembered that you had to specify the comment character.

bpbond:
I agree, but people regularly want/need to go _look_ at those files, e.g. to check formats/headers etc., and I was concerned that seeing `.(g|x)z$` would be confusing.

bpbond:
As you mentioned in #39 we're going to want to do some V1.1 cleanup on all files, I suspect. I'll add a note to that issue re potentially changing compression algorithm.

PR #498, opened 2017-06-28
Snyder lb164 ag costs usa c 2005 irr

abigailsnyder:
#472


R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
```diff
@@ -36,43 +42,130 @@ module_aglu_LB164.ag_Costs_USA_C_2005_irr_DISABLED <- function(command, ...) {
...
+      ungroup %>%
+      select(-iso) %>%
+      left_join_error_no_match(L164.waterCostFrac_Cusda, ., by = "GTAP_crop") %>%
+      mutate(waterCostFrac_wt = waterCostFrac * weight) %>%
+      group_by(GCAM_commodity) %>%
+      summarise(waterCostFrac_wt = sum(waterCostFrac_wt),
+                weight = sum(weight)) %>%
+      ungroup %>%
+      # compute weighted average water cost fraction by GCAM commodities that are represented in USDA cost spreadsheets.
+      mutate(waterCostFrac = waterCostFrac_wt / weight) ->
```
bpbond: Can above lines be replaced simply by `summarise(waterCostFrac = weighted.mean(waterCostFrac_wt, weight))` ?

R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
```diff
@@ -36,43 +42,130 @@ module_aglu_LB164.ag_Costs_USA_C_2005_irr_DISABLED <- function(command, ...) {
...
+      L164.waterCostFrac_C
+
+
+    # Step 3: Compute water cost fractions for crops not in the USDA cost spreadsheets.
+    # For the crops with missing values, make a simple linear regression to predict water cost fraction as a function of
+    # irrigation fraction
+    L161.ag_irrHA_frac_R_C_GLU %>%
+      filter(GCAM_region_ID == iso_GCAM_regID[["GCAM_region_ID"]][iso_GCAM_regID$iso == "usa"]) %>%
```
bpbond: `filter(GCAM_region_ID == gcam.USA_CODE)`

R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
```diff
@@ -36,43 +42,130 @@ module_aglu_LB164.ag_Costs_USA_C_2005_irr_DISABLED <- function(command, ...) {
...
+      # join in water costs for the commodities that ARE covered by USDA spreadsheets, L164.waterCostFrac_C
+      # preserve NA's to process
+      left_join(select(L164.waterCostFrac_C, GCAM_commodity, waterCostFrac),
+                               by = "GCAM_commodity") ->
+      L164.ag_irrHA_frac_USA_C
+
+    # Use L164.ag_irrHA_frac_USA_C to form the linear model waterCostFrac = f(irrHA_frac)
+    # and predict the missing water cost fractions in L164.ag_irrHA_frac_USA_C.
+    L164.ag_irrHA_frac_USA_C %>%
+      glm(waterCostFrac ~ irrHA_frac, data =.) %>%
```
bpbond: Ridiculously picky, but: space after equals please.

R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
```diff
@@ -36,43 +42,130 @@ module_aglu_LB164.ag_Costs_USA_C_2005_irr_DISABLED <- function(command, ...) {
...
+      add_title("Production costs of GCAM commodities not including purchased irrigation water") %>%
+      add_units("1975USD/kg") %>%
+      add_comments("USDA cost data is used to calculate total production costs - purchased irrigation water costs for") %>%
+      add_comments("GCAM commodities covered in USDA spreadsheets. This produces a commodity level water cost fraction for some") %>%
+      add_comments("but not all GCAM commodities. L161.ag_irr_HA_frac_R_C_GLU irrigated vs rainfed harvested area data for the") %>%
+      add_comments("covered GCAM commodities is used to form a linear model describing water cost fraction as a function of the") %>%
+      add_comments("irrigated fraction of harvested area. This linear model then predicts the water cost fraction based on") %>%
+      add_comments("irrigated fraction of harvested area for the missing commodities. Finally, cost data from LB133.ag_Cost_75USDkg_C") %>%
+      add_comments("is adjusted as LB133_cost * (1 - water cost fraction) = production cost - purchased irrigation water for each commodity.") %>%
       add_comments("can be multiple lines") %>%
```
bpbond: Remove

R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
abigailsnyder: I think I ran the tests wrong before pushing this change. Now when I run test package, it fails. also looks like it failed travis. Reverting to original (not using weighted.mean)

bpbond: What's the intent of this calculation then? I mean, it _says_ it's computing a "weighted average water cost fraction" for each GCAM commodity, but that's not what's happening (at least as far as I understand those words). You'd need to change line 87 to `waterCostFrac_wt = sum(waterCostFrac_wt * weight)`...and that would produce a result identical to `weighted.mean`.

So is this a mistake?

abigailsnyder: I don't believe so.
A weighted average of some set of quantities x_i is generally `sum(w_i * x_i) / sum(w_i).`

Your definition only holds for a weighted average when all the weights already sum to 1 (which they do not here).  Essentially, the calculations in the code now are introducing that fractional nature.
To use `weighted.mean`, I believe we would have to go through and adjust the weights to be a fraction out of total. Not difficult, but also the same number of operations (I think) as the code here. 


In other words, with the approach in the code now, we get terms like ` (w1*x1 + w2*x2) / (w1 + w2)`, which is what we want.

(We have these desired terms because `waterCostFrac_wt = waterCostFrac * weight`)

bpbond: Sorry, in my example above, I was omitting the final step of dividing by the sum of the weights.

Well, I'll take your word for it, although it seems like a lot of operations (summarise, mutate, summarise, mutate) to get to what ostensibly is just a weighted mean.

abigailsnyder: I can definitely play with getting it down

abigailsnyder: I think I got it down to just two summarise's; testing now

R/zchunk_LB164.ag_Costs_USA_C_2005_irr.R
```diff
@@ -36,43 +42,124 @@ module_aglu_LB164.ag_Costs_USA_C_2005_irr_DISABLED <- function(command, ...) {
...
+    USDA_cost_data %>%
+      gather(year, cost, matches(YEAR_PATTERN)) %>%
+      mutate(year = as.integer(year)) %>%
+      filter(Item == "Purchased irrigation water" | Item == "Total operating costs",
+             year %in% MODEL_COST_YEARS) %>%
+      spread(Item, cost) %>%
+      rename(IrrCost = `Purchased irrigation water`,
+             TotCost = `Total operating costs`) %>%
+      mutate(CostFrac = IrrCost / TotCost) %>%
+      # filter out any crops with IrrCost in all years
```
kvcalvin: Are we only include crops with IrrCost in all years? Or excluding crops with IrrCost in all years?

abigailsnyder: whoops, that's an incomplete comment. We filter out any crops missing IrrCost in all MODEL_COST_YEARS. So if they have an IrrCost in at least one of the years, they stay

Issue #500, opened 2017-06-29
module_aglu_batch_pasture_ssp34.xml needs node_rename=T


bpbond:
Once #458 is merged, go back to `module_aglu_batch_pasture_ssp34.xml` and add the node_rename. See discussion in #491 .

bpbond:
@pralitp could you push this change to #501 please?

pralitp:
Yup, running tests now

Issue #491, opened 2017-06-28
L242.ssp34_pasture 


bpbond:
OK @kvcalvin ?

bpbond:
@pkyle This chunk reads in both `basin_to_country_mapping` and `GCAMLandLeaf_CdensityLT` but doesn't use them at all. OK to remove both as dependencies?

kvcalvin:
When I wrote this code `basin_to_country_mapping` didn't exist and I have no recollection of `GCAMLandLeaf_CdensityLT`. I don't know if these got added during some kind of merge or if someone went in afterwards to do it. I suppose we should make sure they _shouldn't_ be using those files.

pkyle:
I added the basin_to_country_mapping file in order to allow the GLU number code to be replaced with the short basin name upon data read-in. The line:
L125.LC_bm2_R_LT_Yh_GLU <- readdata( "AGLU_LEVEL1_DATA", "L125.LC_bm2_R_LT_Yh_GLU", replace_GLU = T )
requires basin_to_country_mapping because replace_GLU=T

bpbond:
@pkyle Thanks.

```R
write_mi_data( L242.LN2_MgdAllocation_SSP34, "LN2_MgdAllocation", "AGLU_LEVEL2_DATA", "L242.LN2_MgdAllocation_SSP34", "AGLU_XML_BATCH", "batch_pasture_ssp34.xml", node_rename=T )
```

Note the `node_rename` parameter on this. @pralitp will our existing xml step handle this?

pralitp:
Not until #458 is merged.

bpbond:
OK, thanks. I've added #500 to remind us to do this.

PR #501, opened 2017-06-29
Ben - L242.ssp34_pasture

bpbond:
#491 

Note see also #500 - there’s one thing related to XML generation that’s unimplemented here.


bpbond:
@kvcalvin and/or @pkyle Please check that documentation reads clearly and accurately.

R/module-helpers.R
```diff
@@ -8,7 +8,7 @@
 #' @param water_sector A character vector containing water sector information
 #' @param water_type A character vector containing water type information
 #' @param water_mapping A tibble with a mapping between \code{water.sector} and  \code{supplysector}
-#' @param GLU An optional character vector containing GLU information (only used for irrigation with mapped water types)
+#' @param GLU An optional character vector containing GLU information(only used for irrigation with mapped water types)
```
kvcalvin: Did we really want to remove this space?

R/module-helpers.R
```diff
@@ -62,7 +62,7 @@ set_water_input_name <- function(water_sector, water_type, water_mapping, GLU =
 #' @details Add a suffix to the SO2 gas name indicating which of four major world regions to assign the emissions to,
 #' as Hector considers the geographic location of sulfur emissions. Any code writing out CSVs for conversion to XML
 #' handling SO2 related data should use this function. Agricultural waste burning emissions already have a suffix
-#' assigned (_AWB), so in this case, the SO2 region number is assigned between the "SO2" and "AWB" strings.
+#' assigned(_AWB), so in this case, the SO2 region number is assigned between the "SO2" and "AWB" strings.
```
kvcalvin: Same question here (and for the rest of this comment block)

R/xml-column-names-aglu.R
```diff
@@ -1,7 +1,17 @@
+# XML-related constants
+
+aglu.GLU_NDIGITS          <- 3    # number of digits in the geographic land unit identifier codes
+aglu.GLU_NAME_DELIMITER   <- ""   # delimiter between the GLU name and number
+aglu.LT_GLU_DELIMITER     <-      # delimiter between the land use type name and GLU name. should be the same as the crop-glu delimiter
+  aglu.CROP_GLU_DELIMITER <- "_"  # delimiter between the crop name and GLU name
```
kvcalvin: No spaces at the beginning of the line?

R/zchunk_L242.ssp34_pasture.R
```diff
@@ -0,0 +1,161 @@
...
+      L242.LN2_HistMgdAllocation_ALL
+
+    # Model base period land cover, managed land in the second nest
+    L242.LC_bm2_R_Mgd2_Yh_GLU.mlt %>%
+      filter(year %in% BASE_YEARS) %>%
+      select(region, LandAllocatorRoot, LandNode1, LandNode2, LandLeaf, year, allocation) ->
+      L242.LN2_MgdAllocation_ALL
+
+    # Create SSP4 pasture inputs, isolating low-growth regions for subsequent filtering
+    L102.pcgdp_thous90USD_Scen_R_Y %>%
```
kvcalvin: I think we might do this a lot. I wonder if it is worth making a function.

R/zchunk_L242.ssp34_pasture.R
```diff
@@ -0,0 +1,161 @@
...
+      select(region, LandAllocatorRoot, LandNode1, LandNode2, LandLeaf, year, allocation) ->
+      L242.LN2_HistMgdAllocation_ALL
+
+    # Model base period land cover, managed land in the second nest
+    L242.LC_bm2_R_Mgd2_Yh_GLU.mlt %>%
+      filter(year %in% BASE_YEARS) %>%
+      select(region, LandAllocatorRoot, LandNode1, LandNode2, LandLeaf, year, allocation) ->
+      L242.LN2_MgdAllocation_ALL
+
+    # Create SSP4 pasture inputs, isolating low-growth regions for subsequent filtering
```
kvcalvin: It's actually not "low growth" regions. It is regions that are poor today.

R/module-helpers.R
bpbond: Hahah...oops. I copied over some code from the old data system and did a find-and-replace to fix the GCAM spacing convention. Will fix.

R/xml-column-names-aglu.R
bpbond: Notice the `aglu.LT_GLU_DELIMITER` is assign to it in previous line (i.e. all one line syntactically).

R/zchunk_L242.ssp34_pasture.R
bpbond: The filter is for regions with very low GDP, looks like; call them "poor (low GDP) regions"?

R/zchunk_L242.ssp34_pasture.R
```diff
@@ -0,0 +1,161 @@
...
+    L102.pcgdp_thous90USD_Scen_R_Y %>%
+      filter(scenario == "SSP4", year == 2010) %>%
```
bpbond: What's magic about 2010 here? Should we replace with a constant, either a new one or `max(AGLU_HISTORICAL_YEARS)` or something like that? @kvcalvin @pkyle

R/xml-column-names-aglu.R
kvcalvin: I did not catch that at all.

bpbond:
@kvcalvin thoughts on that hardcoded 2010 value? (Lines 101-104.)

kvcalvin:
@bpbond: I had commented on that chunk that we may want to pull it out into a function (like get_low_income_regions). For now, I think we need to ensure that the year is linked to the one used in the threshold definition. So, we can either hard-code it to 2010 or set a new constant that is the year with which the low_income thresholds are defined.

bpbond:
Thanks. OK, I'll leave it for now but open an issue to remind us to revisit later.

pralitp:
I've verified that `pasture_ssp34.xml` rewrites the LandNode properly but I checked the contents of the generated XML file and what I have in a only branch of the `integration/water-ssp` branch and it seems the gcamdata version is missing `land-use-history` is that expected?

bpbond:
@pralitp Not that I know of. Is this a problem with this PR, or something that's missing in `master`? Coming to chat.

pralitp:
Ok I found the problem.  I think the chunk generator failed to pick up the following line because it was split into two lines:
```R
write_mi_data( L242.LN2_HistUnmgdAllocation_SSP34, IDstring="LN2_HistUnmgdAllocation", domain="AGLU_LEVEL2_DATA", fn="L242.LN2_HistUnmgdAllocation_SSP34",
               batch_XML_domain="AGLU_XML_BATCH", batch_XML_file="batch_pasture_ssp34.xml" )
```

Putting `L242.LN2_HistUnmgdAllocation_SSP34` seems to replicate the XML exactly.

Issue #522, opened 2017-07-07
Data system output list getting large


bpbond:
`driver() -> x` now results in a very large list, ~1.5 GB, and the data system rewrite is only half-ish done. Although this can be filtered for a task at hand (`x <- x[my_chunk("DECLARE_INPUTS")]`) it's a resource drain on the computer, and slows down e.g. building and reloading.

Most data objects are small but some are not:

![sizes](https://user-images.githubusercontent.com/1956468/27968458-1174d9e4-6315-11e7-81f5-cb7508963d94.png)

```
                              object      size
                               <chr>     <dbl>
1   L181.LC_bm2_R_C_Yh_GLU_irr_level  31774696
2              L101.EIA_use_all_Bbtu  34464936
3                  L100.FAO_ag_Imp_t  38558816
4               ag_Fert_IRR_MGMT.xml  41161368
5       L121.nonco2_tg_R_awb_C_Y_GLU  42106672
6                     energy/en_OECD  54737488
7     socioeconomics/SSP_database_v9  71300456
8                        EDGAR_gases  75422696
9  L1211.nonco2_tg_R_awb_C_Y_GLU_IRR  91842328
10                 energy/en_nonOECD 145154456
```
Possibilities:
1. Use factors in the largest tables. If they're heavily repetitive text that might save a lot.
2. Allow chunks to mark input data for deletion after being used. This would help with the IEA data in the list above.
3. Allow the driver to delete `FLAG_INPUT_DATA` objects that will no longer be used. This seems straightforward and would take care of four of the five biggest objects.
4. Data system backend could compress the largest objects. This would be transparent for the chunks, and apply to objects that are (i) large and (ii) accessed infrequently. https://stat.ethz.ch/R-manual/R-devel/library/base/html/memCompress.html
5. Data system backend could cache large objects on disk. This seems wasteful and slow.

Thoughts @rplzzz @pralitp @kvcalvin ?

pralitp:
Huh, I was just thinking about this mostly around the compressing option.  About as far as I got is I should test this out and see the pros and cons (memory saved vs speed lost).

The delete options seem more likely to become more problematic (what if a user wanted to inspect those tables after the driver was done for instance) and so are more a secondary option in my opinion.

bpbond:
[...]

PR #530, opened 2017-07-11
Prune unneeded data as driver runs

bpbond:
This addresses the issues raised #522, in particular memory usage of the driver. It involves significant changes to the driver's behavior: now no data are returned by default, but the user can request (via a variety of parameters) that certain data be returned.

~I can't figure out why the timeshift test is currently dying. Will come back and fix this, but thought it would be useful for now to open a PR for review and testing.~


bpbond:
It would be good to have a bunch of people test this, since it's a big change. I've tagged you @abigailsnyder @pralitp but if busy or you'd prefer not to, no worries, just reassign to someone else. Thanks.

pralitp:
Hey @bpbond, I just checked out the branch and gave it a go and got:

[1] "module_data_Maddison_population"
[1] "- make 0.04"
 Hide Traceback
 
 Rerun with Debug
 Error: data_list is not a character vector 
4.
stop(assertError(attr(res, "msg"))) 
3.
assert_that(is.character(data_list)) at utils-data.R#259
2.
remove_data(names(which_zero), all_data) at driver.R#220
1.
driver()

pralitp:
Note, changing the assert to allow `is.character(data_list) || is.null(data_list)` is sufficient to solve this for me.

bpbond:
Thanks Pralit! :doh:

bpbond:
Though it didn't fail on Travis. Huh. Anyway, tweaked that assertion.

bpbond:
Adding Corinne as a reviewer - please test if you have time.

bpbond:
@pralitp re your question in #522 , I see one "Skipping XML conversion as global option gcamdata.use_java is not TRUE" message but would have expected many? Anyway please check on your system.

pralitp:
Yup, I verified they get generated correctly.  The message only gets printed once per your recommendation so it doesn't spew hundreds of warning messages (was especially an issue when the XML files were all written at the end).

pralitp:
@bpbond, what is the correct name to use for the parameter `return_inputs_of` or `return_outputs_of`?  I tried something like the following but doesn't seem to work.  Using `return_data_names` seems to work as expected.
```R
x <- driver(return_inputs_of = c("module_aglu_batch_demand_input_xml"))
x
$<NA>
NULL
```

bpbond:
@pralitp Um...I would expect that to work. 🐛

bpbond:
As a workaround until I investigate (soon), what about 

```R
x <- driver(stop_before = "module_aglu_batch_demand_input_xml")
```
?

pralitp:
Huh, that works.  Also note to self pick a chunk that gets run earlier in the system for testing.

bpbond:
Thanks @pralitp . Sorry about that. It's tricky to write tests for the driver, because it interacts with so many parts of the data system...but I should have caught that.

pralitp:
No worries at all.  I figured you were asking us to take it for a test spin to iron out these kinks sooner rather than later anyway.

bpbond:
@abigailsnyder @pralitp @cahartin If possible it would be great to get this merged in next 48 hours, so if you have feedback or find bugs, please let me know by end of tomorrow if possible. Thanks.

abigailsnyder:
@bpbond I _should_ have time to look at it after about 12pm tomorrow. So I'll be sure to prioritize this at that point

cahartin:
@bpbond on it after my 2pm call!

cahartin:
@bpbond
It looks like you and Pralit addressed this earlier...?  

 `1] "module_modeltime_batch_modeltime_xml"
[1] "- make 0.00"
Skipping XML conversion as global option gcamdata.use_java is not TRUE`

bpbond:
@cahartin yup thanks. He confirmed that behavior is good.

pralitp:
@cahartin, sort of. That is not affected by this PR.

By default the options `options("gcamdata.use_java")` is `FALSE` so as to assume people don't have Java installed and therefore don't attempt to run the XML conversion.  If you do want the XML files and do have Java installed you can simply run options("gcamdata.use_java" = TRUE)` and you will no longer get that warning.

cahartin:
Is the only option in `all_data`, to return an empty tibble?

Other than that, i went through and ran all of the other options and everything seems to be working.  However, I went to try and do the `debug(module.....)` step and I can't get that working when I stop at particular chunk. This has probably changed since this driver update...?

bpbond:
@cahartin That should be the same as before:
```R
driver(stop_before = "module_aglu_LA100.FAO_downscale_ctry") -> x
debug(gcamdata:::module_aglu_LA100.FAO_downscale_ctry)
gcamdata:::module_aglu_LA100.FAO_downscale_ctry("MAKE", x)
```

cahartin:
@bpbond ah, gotcha, that works on my end.  I can't find any issues with this branch.

abigailsnyder:
I just checked out and ran the driver as-is on my machine. 
It looks like all the outputs were written correctly, based on timestamps in the output folders. 
But I find the "returning 0 tibbles." bit weird in the following:
```
[1] "module_aglu_batch_demand_input_xml"
[1] "- make 0.05"
Writing chunk data...
Returning 0 tibbles.
All done.
```

I would expect plenty of people to assume that's an error?

bpbond:
Good suggestion @abigailsnyder - thanks!

PR #568, opened 2017-07-26
Temporarily disable L112 and switch back to temp-data-inject

bpbond:
See #569


bpbond:
Stupid coverage checks are taking FOREVER. This is however passing Travis and Appveyor builds so merging.

Issue #584, opened 2017-07-29
Write test code for get_ssp4_regions


bpbond:
Merged in #570

PR #585, opened 2017-07-30
Add test code for get-ssp-regions

bpbond:
Also tweaked logic, changed name, and updated documentation. Addresses #584


bpbond:
@rplzzz seems busy, and no review needed here, honestly. Merging.

PR #590, opened 2017-08-01
Re-enable code coverage tests using covr

### LINKED ISSUE #581: Investigate codecov performance problems
As noted in #580 , where it was disabled, the code coverage step of the CI (called from `coverage.R`) has gotten unusable; it's essentially never terminating, now, slowing everything down and resulting in false errors everywhere.

pralitp:
Re-enable coverage tests however explicitly using covr version 2.2.2 which is a known working version (although at this point we are not sure when exactly that library broke).

This closes #581


bpbond:
Hey, slow but it got done!

pralitp:
@bpbond, I was testing to turn off old/new and potentially time shift tests when running the code coverage by using something like the following but the option doesn't seem to make it through and it never skips.  Any thoughts?  Should I give up and just set environment variable or something of the like instead?

```
diff --git a/coverage.R b/coverage.R
index e0f80728..90b51015 100644
--- a/coverage.R
+++ b/coverage.R
@@ -3,6 +3,7 @@
 ## put the commands inline in the after_success section
 
 library(covr)
+options("gcamdata.is_coverage_test" = TRUE)
 x <- package_coverage()
 cat("coverage = ", percent_coverage(x), " %\n")
 codecov(coverage = x)
diff --git a/tests/testthat/test_oldnew.R b/tests/testthat/test_oldnew.R
index 625ca38b..2d08e20f 100644
--- a/tests/testthat/test_oldnew.R
+++ b/tests/testthat/test_oldnew.R
@@ -6,6 +6,14 @@ library(readr)
 
 test_that("matches old data system output", {
 
+  # If we are running the code coverage tests then let's skip this since
+  # it will take a long to time run and the purpose of this test is to
+  # make sure the chunk outputs match the old data system and not to test
+  # the functionality of any chunks
+  if(isTRUE(getOption("gcamdata.is_coverage_test"))) {
+    skip("Skip old new when only interested in code coverage")
+  }
+
```

bpbond:
@pralitp Yeah, I think I'd do it via an environment variable set in `.travis.yml`.

pralitp:
Note on my laptop code cover tests take:

Full: 155.001
Skip old/new: 134.324
And skip time shift: 31.890

I'll commit one at a time to see how it compares on travis too

pralitp:
Ok, with full tests the coverage took 1265.93 seconds.

pralitp:
Skip old / new the coverage took 438.37 seconds.

bpbond:
Time shift is the easy one: it adds nothing to coverage.

On reflection, I think it's fine to skip oldnew, too. It means we won't have coverage for the main body of code chunks, but that's OK--there's no easy way to test them for "correct" behavior anyway. What we really care about is coverage for the support utilities, driver, data subsystem, etc.

bpbond:
@pralitp Did you get a chance to open an issue in the `covr` repository about performance degradation?

pralitp:
@bpbond, yes I did open an issue there (it is number 274 on `jimhester/covr`) and they were able to verify the performance regression and will work on it.

As a side note, another issue that might be work keeping an eye on is number 144 -- getting test results and coverage at the same time.  If we didn't have to essentially run tests twice we wouldn't have to had to disable some tests when running the coverage.

PR #573, opened 2017-07-27
Aaron - LA144.building_det_en

aarony2j:
LA144.building_det_en #396

PR #595, opened 2017-08-02
Aaron - LA144.building_det.en_fix

aarony2j:
See #573


bpbond:
I'm assuming Sha has no further comments. Merging. Thanks and nice work @aarony2j !

sypnl:
Looks good! I don't have any further comments.

PR #603, opened 2017-08-03
Russell-L112_FIX

russellhz:
I added rounding at `zchunk_LB1322.Fert` and changed the comparison data for `zchunk_L101.nonghg_en_USA_S_T_Y`, `zchunk_LB1322.Fert`,  and `zchunk_L112.ghg_en_R_S_T_Y`


russellhz:
Having trouble submitting pull request for old data system
screen shot 2017-08-03 at 5 46 18 pm

bpbond:
Hey @russellhz this looks good I think. Two things: (i) can you post info documenting the changes introduced introduced into the _old_ data system? (I.e. how the outputs dataset(s) differ.) (ii) What about the downstream objects that depend on this?

bpbond:
Note this is holding only on documentation–see comment above.

russellhz:
The change I made was in lines 174-175 and 179-180 in `energy-processing-code/level1/LB1322.Fert.R`. I added a `round(..., 10)` to both lines.

This changed the output of `L1322.in_EJ_R_indenergy_F_Yh` & `L1322.in_EJ_R_indfeed_F_Yh`.

Then, I ran `emissions-processing-code/level1/L101.nonghg_en_USA_S_T_Y.R`, which changed the output of `L101.so2_tgej_USA_en_Sepa_F_Yh`, `L101.co_tgej_USA_en_Sepa_F_Yh`, `L101.nox_tgej_USA_en_Sepa_F_Yh`, `L101.voc_tgej_USA_en_Sepa_F_Yh`, and `L101.in_EJ_R_en_Si_F_Yh`.

Last, I ran `emissions-processing-code/level1/L112.ghg_en_R_S_T_Y.R`, which changed the output of `L112.ghg_tg_R_en_S_F_Yh` and `L112.ghg_tgej_R_en_S_F_Yh`

bpbond:
Sorry--what I meant was, do a numeric comparison of the old and new outputs (both in the old data system) and document to what degree things have changed.

russellhz:
Ah, OK I'll get that done

russellhz:
@bpbond Not sure if this is what you were thinking of, let me know if there are some other metrics/examples you want to see.

Overall, nearly all the downstream changes are insignificant (differences in individual observations of e-10 and smaller) and simply a result of rounding LB1322 to the 10th digit after the decimal. The changes are also almost exclusively in gas industrial energy use, so most values in these files are unchanged. Here I took the sum of the absolute value of the difference in each file from changes to master.

diffs

The only exception is the file that was causing the testing failure, `L112.ghg_tgej_R_en_S_F_Yh`. In this file, we were often dividing values around e-23 by values around e-19, and ending up with something significant. These values are all 0 now, which is the desired behavior. Below is showing that the significant changes are only in gas industrial energy use, and then a specific example.

l112 ghg_tgej

l112 ghg_tgej_full

bpbond:
Thanks @russellhz - perfect.

bpbond:
@russellhz can you send/post the diff for old data system?

russellhz:
image

Issue #604, opened 2017-08-03
LA114.wind very slow for some people


bpbond:
`LA114.wind` executes in 3.5 seconds on my laptop, and @pralitp reports 6s.

But:

@fengly20 50s.
@abigailsnyder 60-62s.
@rynacui 68s.

?

bpbond:
Fixed in #611

Issue #606, opened 2017-08-04
LA114.wind needs some cleanup


bpbond:
Comments originally by @rplzzz :

1. The functions on lines 173 and 209 are way too long to be anonymous functions.  They should be pulled out and documented.

2. We should also consider pulling the inner functions at the beginning of the chunk out and documenting them.  They don’t make closures, so there isn’t really any reason to define them inside of the chunk.

3. ~Some of the inner functions make local data frames for no good reason.  As a result, we do a lot of gc in this chunk.~ (This is being handled in #604 )

4. Long-term: do we really need to create this data at all?  All of it can be calculated analytically, so maybe the best thing is to just have other chunks call those functions as required.

russellhz:
Closing this, as none of these comments seem to apply to the latest version (mostly done in ebd927f1111dba504c0fe239077565a1caf220fd, but also in other commits as well). As to the last point about whether it needs to be done at all,  the output is used in two downstream chunks and I don't see how it would be more transparent to move all of the calculation into module-helpers functions.

PR #611, opened 2017-08-04
snyder - La114.wind.redux

abigailsnyder:
Passing check package on my machine at least

Addresses #604


bpbond:
@rplzzz and @pralitp - to fix #604 about LA114 wind being _very_ slow for some folks, @abigailsnyder and I decided, since the offending code is in a OLD_DATA_SYSTEM_BEHAVIOR block, and the new code is lots faster, that the simplest thing was just to plug in the old data system's values. When we get rid of this old behavior, we'll toss away the pre-set points too obviously. Sound good?

bpbond:
@rplzzz thanks. We have this note in #606 .

bpbond:
Thanks for taking this on @abigailsnyder 👏

Issue #484, opened 2017-06-26
L2112.ag_nonco2_IRR_MGMT


cahartin:

cahartin:
@bpbond can you add in any necessary files?  Thanks!

PR #617, opened 2017-08-07
Kd - L2112.ag_nonco2_IRR_MGMT

kdorheim:
This chunk disaggregates agricultural emissions on the basis of production. 

#484


R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
...
 #'
 #' @param command API command to execute
 #' @param ... other optional parameters, depending on command
 #' @return Depends on \code{command}: either a vector of required inputs,
 #' a vector of output names, or (if \code{command} is "MAKE") all
-#' the generated outputs: \code{object}, \code{L2112.AnEmissions}, \code{L2112.AnNH3Emissions}, \code{L2112.AWBEmissions}, \code{L2112.AGREmissions}. The corresponding file in the
+#' the generated outputs: \code{object}, \code{L2112.AWBEmissions}, \code{L2112.AGREmissions}. The corresponding file in the
 #' original data system was \code{L2112.ag_nonco2_IRR_MGMT.R} (emissions level2).
-#' @details Describe in detail what this chunk does.
+#' @details Disaggregates agricultural emissions the basis of production by scaling emissions by a technology factor
```
cahartin: can you add in what emissions?

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
 #' module_emissions_L2112.ag_nonco2_IRR_MGMT
 #'
-#' Briefly describe what this chunk does.
+#' Disaggregates agricultural emissions the basis of production
```
bpbond: Doesn't make grammatical sense.

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
...
+
+    # First calculate the total emissions for each region, supply sector, subsector,
+    # technology level in the most recent model base year. This total will latter be
+    # used to  calculate share weights.
+    L2012.AgProduction_ag_irr_mgmt %>%
+      filter(year == max(BASE_YEARS)) %>%
+      mutate(AgProductionTechnology_nolvl = gsub("_hi|_lo", "", AgProductionTechnology)) %>%
+      group_by(region, AgSupplySector, AgSupplySubsector, AgProductionTechnology_nolvl, year) %>%
+      summarise(total = sum(calOutputValue)) %>%
+      ungroup(.) ->
```
bpbond: Just `ungroup` or `ungroup()`

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
...
+      mutate(share_tech = calOutputValue / total) ->
+      L2112.AgProduction_ag_share
+
+
+    # Combine non agricultural waste burning emissions and agricultural waste burning
+    # emissions into a single data frame. Add management level information to production
+    # technology column.
+    L2111.AWBEmissions %>%
+      bind_rows(L2111.AGREmissions) %>%
+      repeat_add_columns(tibble::tibble(level = c("hi", "lo"))) %>%
```
bpbond: Here and throughout, this is OK, but you can also just write `tibble` (no namespace qualifier)

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
...
+    # Match the shares(fraction of emissions) to the data frames containing
+    # emissions quantities for the agricultural water burning and agricultural
+    # non waste burning emissions.
+    L2112.AgProduction_ag_share %>%
+      select(-year, -AgSupplySubsector) %>%
+      left_join(L2112.awb_agr_emissions, by = c("region", "AgSupplySector", "AgProductionTechnology_lvl")) ->
+      L2112.awb_agr_emissions
+
+
+    # Where shares allocated to lo/hi are NA but emissions are positive, split it 50/50 between
```
cahartin: nice comments!

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
```diff
@@ -1,135 +1,193 @@
...
+    # the techs. For all others, set share to zero. Then scale the input emissions by the emission
+    # shares.
+    L2112.awb_agr_emissions %>%
+      mutate(share_tech = if_else(is.na(share_tech) & input.emissions > 1e-6, 0.5, share_tech)) %>%
+      mutate(share_tech = if_else(is.na(share_tech), 0, share_tech)) %>%
```
bpbond: Use `replace_na` please--it's faster and clearer

R/zchunk_batch_all_aglu_emissions_IRR.xml.R
```diff
@@ -1,14 +1,14 @@
-#' module_emissions_batch_all_aglu_emissions_IRR.xml
```
bpbond: Wait...why are there two xml-generating files here? You should just need `zchunk_batch_all_aglu_emissions_IRR_MGMT.xml.R` I think?

R/zchunk_L2112.ag_nonco2_IRR_MGMT.R
kdorheim: Is it possible to use `replace_na` to replace something with multiple conditions? When I tried to use it to evaluate 2 statements I kept on getting `Evaluation error: no applicable method for 'replace_na' applied to an object of class "c('double', 'numeric')".`

If you really want me to used `replace_na` I would have to filter by the input.emission conditions, replace na, and then left join. Or I could do use `replace` for the first mutate statement and then `replace_na` in the second mutate statement. But this method will still use a `is.na`. Which method would prefer?

bpbond: It's only the second `mutate` that can (easily) be replaced by `replace_na`. The first, no, agreed.

R/zchunk_batch_all_aglu_emissions_IRR.xml.R
kdorheim: Ah!  `zchunk_batch_all_aglu_emissions_IRR.xml.R` generates the xml from L2111. I didn't mean to change it. I went back and found the original version of the file and can include it in my next commit.

Issue #619, opened 2017-08-08
VERSION 1 CHECKOFF - MODELTIME


bpbond:
Verify that

- [x] new data system produces exactly same `modeltime` XML files as old data system
```
batch_hector.xml
batch_magicc.xml
batch_modeltime.xml
```
- [x] structure of files matches

- [x] contents of files matches (we assume true for data, based on `oldnew` test)

- [x] GCAM runs with these files substituted in

bpbond:
This and socioeconomics are the two areas already done, and because they're small, they're handy test cases. I'm assigning (if I may) this to us as the DSR XML lead and modeltime lead, respectively. (Modeltime was the very first thing done in DSR, before we had teams or leads.)

**We'd like to declare them officially done for Version 1**: producing exactly the same XML files as the old data system. (Not exactly: we have made some changes, there's numerical precision issues, etc.) I'm hoping that Pralit has time to check the XML files, and work with me to resolve any problems.

pralitp:
@bpbond, given these files are exactly the same do we really need to run them through GCAM or can we just mark this good to go?

bpbond:
@pralitp Up to you. I was just thinking that's the acid test, so to speak. But you're in charge of this final certification process, and if you think not worth it, fine.

pralitp:
Maybe for just these.  For other modules it will definitely make sense to run them through GCAM.

bpbond:
Makes sense to me.

bpbond:
Reopening so @kdorheim can get some practice at this.

kdorheim:
Results from the model time run-xml-tests.py 
``` 
### oldfile: ../old-xml/modeltime.xml    newfile: ../new-xml/modeltime.xml
1 file tested.
0 missing output files.
0 file discrepancies.
```

bpbond:
Why did it only test one file but three are listed above?

kdorheim:
There is only one xml file in 
```
gcam-core/input/gcam-data-system/xml/modeltime-xml
```

bpbond:
@pralitp has wisdom to share here

pralitp:
For _legacy_ reasons the `hector.xml` and `magicc.xml` got saved to a different location just outside the old `gcam-data-system`.  You can find then in the GCAM repository in `input/climate`

pralitp:
@kdorheim, as I typed that I realized that you need to get those files from, exactly the right revision of GCAM.  Instead of explaining that I've just attached them
[...]

PR #621, opened 2017-08-08
Add missing XML file chunks

bpbond:
As part of #619 I added these missing XML-generating chunks, but it’s throwing an error.

```
[1] "module_climate_xml_batch_magicc.xml"
 Show Traceback
 
 Rerun with Debug
 Error: Unknown columns 'default.emiss.file' 
```

@pralitp could use help


pralitp:
@bpbond it looks like somehow we missed a change (maybe when we switched from core data system to integration/water-ssp data system since `module_modeltime_L200.modeltime` was amongst the first to get translated?

Anyways, I recreated it however I'm worried old/new will fail now.  Where is the old data stored that is checked against and if it does indeed fail I could just update the old CSV files right?

bpbond:
Thank you! Looks good and is passing tests.

bpbond:
P.S. the data it's compared against is stored in `tests/testthat/comparison-data`

pralitp:
Thanks, yea I poked around and updated it.

Issue #337, opened 2017-05-02
Deprecate GCAM3


d3y419:
Per @rplzzz @pkyle @d3y419 @leon:

we're proposing to deprecate GCAM3 population and GDP.

please tag any files that use GCAM3 population and GDP here.

d3y419:
@pkyle said there are a few more places that GCAM3 is used. https://confluence.pnnl.gov/confluence/display/JGCRI/246+Core+Model+Proposal%3A+Deprecate+GCAM3+socioeconomics+in+DSR+2017-05-03

1.  level2 electricity (L223.electricity), in going from the share-weights assumed by GCAM3 region to the country and back up to the new regions, it's just hard-wired to use the GCAM3 socioeconomics
2. HDD/CDD population weighting,

my suggestions is:
old-system-behavior: permanently-inject-gcam3-data
new-system-behavior: switch to gSSP2 (or make this global constant: "standard scenario")

is this an acceptable get around?: @bpbond

russellhz:
moved to GCAM JIRA issue (JGCRI-437)

PR #625, opened 2017-08-09
Fix inadvertent SSP3/5 that should have been gSSP3/5 in bld and trn scio xml.

pralitp:
This fixes some of the missing XML files noted in #620, still no GCAM3 inputs files.  Was that a conscious decision to leave out?

Perhaps the bld / trn socioeconomic files should have been implemented in a loop which could have avoided the inadvertent SSP instead gSSP bug.  @d3y419 can clean that up if he so chooses.


bpbond:
Thanks!

bpbond:
Is this merge affected by the GCAM3 question - #337 ?

pralitp:
Nope, this PR doesn't have to do with the GCAM3 question.  We will have to deal with that separately before we can finish off #620

Issue #588, opened 2017-07-31
L225.hydrogen


bpbond:

bpbond:
Hi @fengly20 - we looked and thought this might be good one to try. It is a "Level 2" script, which means there are some things to be aware of--feel free to touch base with me, Russell, Abigail, or other folks who have done this before. Sound OK? Thanks!

cc @d3y419

fengly20:
Sounds great! Thanks!

PR #633, opened 2017-08-11
LY-L225.hydrogen

fengly20:
Chunk re-write finished and passed test. 

#588


R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+        write_to_all_regions(LEVEL2_DATA_NAMES[['SubsectorShrwt']], GCAM_region_names) ->
+        L225.SubsectorShrwt_h2
+    }
+    if(any(!is.na(A25.subsector_shrwt$year.fillout))){
+      A25.subsector_shrwt[!is.na(A25.subsector_shrwt$year.fillout),] %>%
+        write_to_all_regions(LEVEL2_DATA_NAMES[['SubsectorShrwtFllt']], GCAM_region_names) ->
+        L225.SubsectorShrwtFllt_h2
+    }
+
+    if(exists("A25.subsector_interp")){
```
bpbond: This is always guaranteed to exist after a call to `get_data`.

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      }
+    }
+
+    # 1c. Technology information
+
+    # L225.StubTech_h2: Identification of stub technologies of hydrogen
+    # Note: assuming that technology list in the shareweight table includes the full set (any others would default to a 0 shareweight)
+    A25.globaltech_shrwt %>%
+      write_to_all_regions(LEVEL2_DATA_NAMES[['Tech']], GCAM_region_names) -> L225.StubTech_h2
+    names(L225.StubTech_h2) <- LEVEL2_DATA_NAMES[['StubTech']]
```
bpbond: Mmmm....is it guaranteed columns will be in same order? Seems unlikely.

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+
+    # L225.GlobalTechEff_h2: Energy inputs and efficiencies of global technologies for hydrogen
+    # Efficiencies of global technologies
+    A25.globaltech_eff %>%
+      gather(year, efficiency, -supplysector, -subsector, -technology, -minicam.energy.input) %>%
```
bpbond: `gather(year, efficient, matches(YEAR_PATTERN))` might be clearer.

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      mutate(year=as.numeric(year)) ->
+      A25.globaltech_eff_raw_long
+
+    df_years <- unique(A25.globaltech_eff_raw_long$year)
+    int_years <- c(BASE_YEARS, FUTURE_YEARS)[which(!(c(BASE_YEARS, FUTURE_YEARS) %in% df_years))]
```
bpbond: `?setdiff`

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+
+    A25.globaltech_eff_raw_long %>%
+      select(-efficiency,-year) %>%
```
bpbond: Space after comma please

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      unique() %>%
+      repeat_add_columns(tibble::tibble(year=int_years)) %>%
```
bpbond: No need to `tibble::`; spaces around mathematical operators like `=` pls

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      bind_rows(A25.globaltech_eff_raw_long) %>%
+      group_by(supplysector, subsector, technology, minicam.energy.input) %>%
+      mutate(efficiency=approx_fun(year, efficiency, rule=1)) %>%
+      mutate(efficiency=round(efficiency,energy.DIGITS_EFFICIENCY)) %>%
```
bpbond: Combine into single mutate; also spaces around `=`

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      filter(year %in% c(BASE_YEARS,FUTURE_YEARS)) %>%
+      rename(sector.name=supplysector) %>% #Assign the columns "sector.name" and "subsector.name", consistent with the location info of a global technology
+      rename(subsector.name=subsector) ->
```
bpbond: As above

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+    # Costs of global technologies
+    A25.globaltech_cost %>%
+      gather(year, input.cost, -supplysector, -subsector, -technology, -minicam.non.energy.input) %>%
+      mutate(year=as.numeric(year)) ->
+      A25.globaltech_cost_raw_long
+
+    df_years <- unique(A25.globaltech_cost_raw_long$year)
+    int_years <- c(BASE_YEARS, FUTURE_YEARS)[which(!(c(BASE_YEARS, FUTURE_YEARS) %in% df_years))]
+
+    A25.globaltech_cost_raw_long %>%
```
bpbond: This pipeline could use a good explanatory comment. What's going on here?

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      select(-input.cost,-year) %>%
+      unique() %>%
+      repeat_add_columns(tibble::tibble(year=int_years)) %>%
+      bind_rows(A25.globaltech_cost_raw_long) %>%
+      group_by(supplysector, subsector, technology, minicam.non.energy.input) %>%
+      mutate(input.cost=approx_fun(year, input.cost, rule=1)) %>%
```
bpbond: Again, comment this: clarify what's happening, esp. w/ `rule = 1`

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      mutate(input.cost=round(input.cost,energy.DIGITS_COST)) %>%
+      filter(year %in% c(BASE_YEARS,FUTURE_YEARS)) %>%
+      rename(sector.name=supplysector) %>% #Assign the columns "sector.name" and "subsector.name", consistent with the location info of a global technology
+      rename(subsector.name=subsector) ->
```
bpbond: Combine into single rename call

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
     # Produce outputs
-    # Temporary code below sends back empty data frames marked "don't test"
-    # Note that all precursor names (in `add_precursor`) must be in this chunk's inputs
-    # There's also a `same_precursors_as(x)` you can use
-    # If no precursors (very rare) don't call `add_precursor` at all
+
+    tibble() %>%
+      add_title("Supply sector information for hydrogen sectors") %>%
+      add_units("Unitless") %>%
+      add_comments("Expand sector information for all GCAM regions") %>%
```
bpbond: This is not very clear. Work with @mwisepnnl to clarify and expand comment please.

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+      L225.GlobalTechCapture_h2
+
+    # ===================================================
     # Produce outputs
-    # Temporary code below sends back empty data frames marked "don't test"
-    # Note that all precursor names (in `add_precursor`) must be in this chunk's inputs
-    # There's also a `same_precursors_as(x)` you can use
-    # If no precursors (very rare) don't call `add_precursor` at all
+
+    tibble() %>%
```
bpbond: Here and below--this is passing an empty data frame in, which is **not** what you want! Replace with the names of the final output data frames. If this is confusing, let's talk on Monday--I'm happy to clarify.

inst/extdata/energy/A25.globaltech_co2capture.csv
```diff
@@ -0,0 +1 @@
+# File: A25.globaltech_co2capture.csv# Title: Hydrogen technology capture fractions# Units: Unitlesssupplysector,subsector,technology,1971,2020,2035,2050,2065,2100H2 central production,coal,coal chemical CCS,0.91,0.91,0.92,0.93,0.94,0.94H2 central production,gas,natural gas steam reforming CCS,0.91,0.91,0.92,0.93,0.94,0.94H2 central production,biomass,biomass to H2 CCS,0.91,0.91,0.92,0.93,0.94,0.94
```
bpbond: Line endings messed up (windows, not *nix). I can fix, or re-save in a good text editor.

R/zchunk_L225.hydrogen.R
```diff
@@ -1,237 +1,287 @@
...
+
+    if(exists("A25.subsector_interp")){
+      # L225.SubsectorInterp_h2 and L225.SubsectorInterpTo_h2: Subsector shareweight interpolation of hydrogen sectors
+      if(any(is.na(A25.subsector_interp$to.value))){
+        A25.subsector_interp[is.na(A25.subsector_interp$to.value),] %>%
+          write_to_all_regions(LEVEL2_DATA_NAMES[['SubsectorInterp']], GCAM_region_names)->
+          L225.SubsectorInterp_h2
+      }
+      if(any(!is.na(A25.subsector_interp$to.value))){
+        A25.subsector_interp[!is.na(A25.subsector_interp$to.value),] %>%
```
bpbond:
```R
A25. subsector_interp %>%
  filter(!is.na(to.value)) %>%
```

bpbond:
Hi @fengly20 - I fixed the line endings, corrected the chunk so it will correctly return its data to the driver, and made a number of other syntax tweaks. Note however that right now this doesn't run--on both my system and Travis it errors on one of the `select` statements.

R/zchunk_L225.hydrogen.R
fengly20: It's one of my concerns...I can directly use `rename` to rename the columns since I know the old and new names. But in that way it seems that there's no point using `LEVEL2_DATA_NAMES[['StubTech']]`...what if the contents of `LEVEL2_DATA_NAMES[['StubTech']]` changes in the future?

bpbond: I think the better way is to do this?
```R
A25.globaltech_shrwt %>%
     write_to_all_regions(LEVEL2_DATA_NAMES[['Tech']], GCAM_region_names) %>%
     select(one_of(LEVEL2_DATA_NAMES[['StubTech']])) ->
     L225.StubTech_h2
```

fengly20: Emmm...seems not working. `select` picks out the common columns and drops out the columns I wanted to rename. I guess I could use 'rename' to directly rename the desired column since later on similar behavior happened in both re-written code and old code.

R/zchunk_L225.hydrogen.R
fengly20: Sure. I'll add more comments. Basically the section is trying to behave like what 'interpolate_and_melt' does in the old data system.

bpbond:
Waiting on your review @mwisepnnl

fengly20:
Hi @bpbond , thank you for all your notes and code changes. I edited my chunk code and most of your comments were addressed in this update (except ones related to adding proper comments to output files). I haven't encounter any issue with `select` before and now the code can run through at least on my end. 

There is also one failure when testing the chunk and I think it is related to file naming in the old system. An R object called `L225.PrimaryRenewKeyword_h2` got written out with the name '"L225.PrimaryRenewKeyword.csv"' in the old data system (line 141), while in the new system, the same object got written out with the name "L225.PrimaryRenewKeyword_h2.csv". The inconsistency in output filenames caused the testing failure -- when I renamed the file in the comparison folder the chunk passed testing just fine. Would you be having any suggestions on resolving issues like this? Thanks!

bpbond:
Hi @fengly20 - hmm, it looks like there are a number of problems w.r.t. outputs. Hang on, let me test and push up a fix.

bpbond:
@fengly20 re the naming issue this sounds exactly like @kdorheim saw in #649 - a typo in the old data system. Just rename the comparison file (and commit the rename) please.

bpbond:
OK, I don't know how this happened but there were some outputs of `L225.hydrogen` that weren't 'skeletonized' into the new file you started from. (It's done by a script, which is dumb and can miss things sometime.) Anyway, @fengly20 see changes I made, adding four new outputs. Looks like some already exist and some you'll need to add a bit of code for. If anything is unclear let's talk. Thanks!

fengly20:
@bpbond Sure, thanks! I noticed some outputs of the old chunk were missing in the new chunk skeleton but thought they were discarded for some reason. I should have checked with you about this. I'll start looking at the chunk first and make according changes.

fengly20:
@bpbond I tried to fix the chunk but having trouble dealing with conditional outputs. Three newly added output objects (`L225.SubsectorShrwt_h2`, `L225.SubsectorInterp_h2`, `L225.SubsectorInterpTo_h2`) will not be generated since they cannot pass the if statement. Should I just delete all related code to make it clear? I tried conditional return data but seems not working (`if(condition) {return_data(data_objects)}`).

bpbond:
@fengly20 thanks for your work. Let me take a look.

bpbond:
@pralitp @pkyle (and others?) This chunk has a variety of outputs that are generated only when certain inputs are present–e.g. `L225.SubsectorShrwt_h2` and `L225.SubsectorShrwtFllt_h2` (see lines 73-). How should we handle these? Currently all inputs _have_ to be present, so there's no point in checking whether they exist or not. Should the inputs be optional? Should all outputs always be generated? Any thoughts would be welcome, thanks.

pkyle:
The reason for the if statements is that while the input data table is necessarily present, we don't know what individual elements will have data as opposed to missing values. We may want to read in share-weights "filled out" (i.e., copied forward) or not. The file A25.subsector_shrwt.csv has a column called "year" that is just left blank in this case, but in other sectors it does have values. I've written some examples below of the sort of functionality this enables.
First, the way it is right now, where share-weights are simply filled out from 2020 onward to the same value:
`<share-weight fillout="1" year="2020">1</share-weight>`
But we may want to phase in a subsector:
`<share-weight year="2020">0.25</share-weight>`
`<share-weight year="2025">0.5</share-weight>`
`<share-weight fillout="1" year="2030">1</share-weight>`
The reason for having both of these options present in the same CSV table, rather than having separate CSV tables, is that it's helpful to be able to see both of these options together, so that we don't accidentally set internally conflicted values where the one the model uses will be the one that happens to be read in second.
Note that we still do have this issue with interpolation rules, which are set in a different CSV file. So, for each \*.subsector_shrwt.csv file, we need to pay attention to what's in the corresponding \*.subsector_interp.csv files. Still, since share-weights derived from interpolation rules always over-ride share-weights set elsewhere, the file read-in order doesn't matter.

pralitp:
I wonder if this stemmed from the limitation that it used to be the case if the Model Interface got a CSV file with no data it threw an error.  That has been fixed a while back so as long as `write_to_all_regions` can handle zero rows of data gracefully we should be fine to drop the `if` statements I think.

bpbond:
Thanks @pkyle . Re @pralitp 's point - excellent, this was my preferred solution! So @fengly20 please remove all those `if` checks. If `write_to_all_regions` chokes with zero rows, let me know and we'll tweak it.

fengly20:
Hi @bpbond , I removed all the `if` statement and the chunk choked pretty fast...Three outputs are causing troubles.    

(1) `L225.SubsectorShrwt_h2` (line 74)
The `filter` in line 76 could not go through thus stopped the pipeline. 

(2) `L225.SubsectorInterp_h2` (line 88)
The generation of the object relies on input `A25.subsector_interp` which is not read in in both new and old chunk. (The file itself does exist though in the old system). 

(3) L225.SubsectorInterpTo_h2 (line 94) 
Same as (2)

bpbond:
Hmm. OK, thanks @fengly20 - I will try to take a close look at this later today.

pralitp:
> (2) L225.SubsectorInterp_h2 (line 88)
> The generation of the object relies on input A25.subsector_interp which is not read in in both new and old chunk. (The file itself does exist though in the old system).
>
> (3) L225.SubsectorInterpTo_h2 (line 94)
> Same as (2)

Right, we will have to create those files with no rows of data (just with proper comments / column headings).  You can look at `ANN.subsector_interp` for some other sector to see what the format should be.

bpbond:
@pkyle @pralitp This block appears in this chunk:
```R
if( exists( "A25.subsector_interp" ) ){
	printlog( "L225.SubsectorInterp_h2 and L225.SubsectorInterpTo_h2: Subsector shareweight interpolation of hydrogen sectors" )
	if( any( is.na( A25.subsector_interp$to.value ) ) ){
		L225.SubsectorInterp_h2 <- write_to_all_regions( A25.subsector_interp[ is.na( A25.subsector_interp$to.value ), ], names_SubsectorInterp )
	}
	if( any( !is.na( A25.subsector_interp$to.value ) ) ){
		L225.SubsectorInterpTo_h2 <- write_to_all_regions( A25.subsector_interp[ !is.na( A25.subsector_interp$to.value ), ], names_SubsectorInterpTo )
	}	
}
```

`A25.subsector_interp` isn't read by the old data code (i.e. there's no `readdata` call for it), and I don't see it anywhere in the old data file energy assumptions. Should this code block be ignored, or how should it be handled?

bpbond:
@fengly20 I've fixed a number of problems (see commits) and this chunk now runs and passes tests. For `L225.SubsectorShrwt_h2`, `L225.SubsectorInterp_h2`, and `L225.SubsectorInterpTo_h2` it generates an empty tibble that is marked FLAG_NO_TEST.

bpbond:
@pralitp I did this:

```R
    tibble(x = NA) %>%
      add_units("None") %>%
      add_comments("Not generated") %>%
      add_flags(FLAG_NO_TEST) ->
      L225.SubsectorShrwt_h2
```

Should `L225.SubsectorShrwt_h2` have 'correct' columns, i.e. should there be a call to `write_to_all_regions` in that pipeline?

bpbond:
@mwisepnnl This is still waiting on your (or someone's) disciplinary review.

pkyle:
Ignoring (i.e., not including) the code for processing the non-existent A25.subsector_interp.csv file certainly doesn't hurt; it looks like I meant to exclude subsector share-weight interpolation from this sector entirely, and intentionally omitted the CSV file and readdata() statement, but messed up by including those lines of code...which didn't end up doing anything because their preceding if() conditions were both false.
Having this functionality but not exercising it seems a waste of effort. To the extent that we want to allow share-weight interpolation for hydrogen production, it ought to be region-specific. This data/code method writes share-weight interpolation rules equally to all regions, a capability that doesn't strike me as all that necessary. Right now in the model we don't regionally differentiate energy prices; instead, calibrated share-weights carried forward bring price effects into the future decisions. In the current approach, because hydrogen production is not calibrated, and each fuel costs just about the same in each region, all regions have (more or less) equal shares of the different fuel options in future periods. That's not ideal. If we ever start focusing on hydrogen in the future, we will need inter-regional differences in fuel (feedstock) market shares, and one easy way to get that will be to read in exogenous share-weights that are specific to each region and fuel.

pralitp:
Yup, I was going to comment along the lines of what Page said.  If we really cared to have this functionality we would create the `A25.subsector_interp.csv` file which would be an empty file (in terms of rows of data) and process it normally (i.e. call `write_to_all_regions` etc).

But if we suspect users wouldn't really want to use that feature, and it sounds like from Page that it wouldn't even make sense to do so, then just putting a place holder to get us through v1.0 then dropping it moving forward makes sense to me.

bpbond:
Thanks. I've added it as #668 to follow up on later.

bpbond:
@fengly20 This is now waiting only on a review. I'm in Washington (state), but feel free to knock on @mwisepnnl 's door and remind him!

PR #643, opened 2017-08-15
snyder - batch_xml_for_L221_land_input_1

abigailsnyder:
@pralitp Trying to take what you went over yesterday to complete the xmls for my previous chunk. I wasn't sure what function to use in the batch xml file for `L221.LN1_ValueLogit` since it mixes both land value and logit information. 

@rynacui This might be slightly more clearer to learn from than my `land_input_2` PR since it's strictly dealing with the xml pieces of `land_input_1`. Still won't have the context from Pralit, but at least the diff's might be clearer here.


abigailsnyder:
lordy, I used the  phrase "more clearer" above. Leaving it to shame myself.
cue Cersei season 5 finale gif

bpbond:
@abigailsnyder NO SPOILERS!!! Trying finish season 5 right now. Goal is to watch the end of season 7 "with the rest of the world". ⚔️

abigailsnyder:
Oh noooooooooo I figured two years out was safe. No season 5 and later spoilers from here on out

bpbond:
Working hard to fix this! :loser:

pralitp:
Oops, I guess there was one more caveat to the batch xml chunks.  The land input batch xml chunks need to include a call to `add_rename_landnode_xml()` in it's XML pipeline.  You can check `? add_rename_landnode_xml` or the [Wiki page](https://github.com/JGCRI/gcamdata/wiki/XML-Conversion) for details as to why.

Otherwise looks good.

abigailsnyder:
@pralitp thanks. @bpbond that line is going to have to get added to the chunk of mine you just merged, `L222.land_input_2`

bpbond:
>that line is going to have to get added to the chunk of mine you just merged, L222.land_input_2

Groan. Do you want to add that to this PR?

pralitp:
Or we could kick this can down the road as we will take a final pass through the XMLs once each module is done anyway and could just clean it up then.

abigailsnyder:
@bpbond  I added it. Still doing some testing on my machine but it's an easy enough commit to revert if it breaks stuff

Issue #640, opened 2017-08-15
adjust `left_join_error_no_match` for level 2 chunks


abigailsnyder:
One thing I've noticed with level 2 chunks is that we have to carry a column `logit.type` whose values are all NA whenever we're creating a logit-related output.

Therefore, for these logit related tables, we can't use `left_join_error_no_match`, often simply due to the `logit.type` column.  There's obviously ways to work around it, but is there any way `left_join_error_no_match` can be updated to ignore NA in `logit.type` so that the more restrictive join can be more seamlessly used in the level 2 chunks?

rplzzz:
Easiest thing to do would be to add argument like `na.allowed` to `left_join_error_no_match`.

bpbond:
Yeah, or an `ignore_columns = "logit.type"` one.

PR #644, opened 2017-08-15
Implement an ignore_columns option in left_join_error_no_match

bpbond:
Addresses #640

Note that it turned out that `left_join_error_no_match` didn’t have any test code! So this is a good improvement.


pralitp:
What about using something like `anti_join` would that be more robust than having to explicitly specify which columns to ignore?  The downside is we would have to do another join.

bpbond:
@pralitp Don't we _want_ to force users to explicitly say which columns to ignore? It seems a lot more conservative and safer. And yes, a second join seems potentially problematic, performance-wise.

I'll stick with this approach, I guess, unless you feel strongly. Thanks.

pralitp:
I think maybe I misunderstood what the issue was, Sorry!  Don't mind me.

R/pipeline-helpers.R
```diff
@@ -18,19 +18,21 @@ PH_year_value_historical <- function(d) {
...
   d <- left_join(d, ...)
   if(nrow(d) != drows) {
     stop("left_join_no_match: number of rows in data changed")
   }
-  if(any(is.na(d[setdiff(names(d), dnames)]))) {
+  names_to_check <- setdiff(names(d), dnames) %>% setdiff(ignore_columns)
+  if(any(is.na(d[names_to_check]))) {
     stop("left_join_no_match: NA values in new data columns")
   }
   d
```
abigailsnyder: super clear, great

tests/testthat/test_pipeline-helpers.R
```diff
@@ -67,6 +67,28 @@ test_that("protect and unprotect integer cols work", {
...
+  y <- tibble(year = 2000:2002, value = 1:3)
+  expect_silent(left_join_error_no_match(x, y, by = "year"))
+  expect_identical(left_join_error_no_match(x, y, by = "year"),
+                   left_join(x, y, by = "year"))
+
+  # Catches duplicate matches
+  y <- tibble(year = c(2000, 2001, 2001, 2002), value = 1:4)
+  expect_error(left_join_error_no_match(x, y, by = "year"))
+
+  # Catch no match (NAs)
```
abigailsnyder: The only other test I can maybe think of is an explicit one for the number of rows changing, similar to this test. Since those are the two if conditions that _should_ throw errors based on the function.

bpbond: Not sure I understand, sorry. Line 81 catches an b/c `x` has three rows, but after the join has 4, so the row-number-change is what's being tested here. Do you mean a test to confirm that `left_join(x, y)` changes the number of rows as expected?

Issue #636, opened 2017-08-14
Do GCM3 GDP and population outputs


bpbond:
This is not ideal, but after discussion with @pralitp and @d3y419 it seems better to implement GCAM3 outputs for now. (See #337 .) The problem is that it's not a simple removal of unneeded outputs; rather doing so changes a number of downstream outputs, and so removing all the GCAM3 stuff complicates our V1 comparison. @russellhz estimates we're talking 4-8 hours of work to put them in for

* L102.GDP.R
* L201.Pop_GDP_scenarios.R
* L242.Bld_Inc_Elas_scenarios.R
* L252.Trn_Inc_Elas_scenarios.R

Yes, this work will probably be ripped back out in the 6-12 month timeframe, but it's a small investment and allows a cleaner, simpler trip to V1.

Russell, you good with this? At your convenience.

russellhz:
Sounds good!

rplzzz:
Just a thought, would it be easier to cut these outputs out of the comparison output, instead of adding them to the new output?

bpbond:
Thanks, Robert. Yes, we can easily exclude things, but that's not the problem. Rather, it's that the GCAM3 outputs flow into other downstream data products that we _do_ want to verify for version 1.

rplzzz:
_Do_ we want to verify them?  If they used GCAM3 data to calculate, then they are junk.  we could just remove all of the downstream GCAM3 products.  Or are you saying that GCAM3 values are being used to calculate values that are ostensibly GCAM4 data?  If that's what's happening, then the old data system is simply _wrong_ and should be fixed.  (This, btw, is one of the best arguments against keeping obsolete data around "just in case".)

bpbond:
My understanding is that GCAM3 data flow into ostensibly non-GCAM3 outputs, but roping in @pralitp and @d3y419 for a more precise answer.

pralitp:
@pkyle was the one who initially looked into where it was being used so he can better comment on that but in the meantime you can check the comments in #337. 

In any case it isn't clear why the downstream products should be _wrong_.  Obviously most of the data products in the system were created long before the SSPs were finalized (and they are still not in the Core version of the model).  Which socioeconomics should they have used?

rplzzz:
If half the model is using GCAM4 socioeconomics, and the other half is using GCAM3 socioeconomics, then the socioeconomics aren't consistent across the model, which is wrong in my book.

pkyle:
[...]

PR #652, opened 2017-08-17
Russell-GCAM3

russellhz:
#636


R/zchunk_L102.GDP.R
```diff
@@ -285,6 +299,100 @@ module_socioeconomics_L102.GDP <- function(command, ...) {
...
+    # rebasing GDP to 2010 USD at country level
+    # Calculate the GDP ratios from the first year in the projections. Use this ratio to project GDP from historical dataset in final historical period
+    gdpRatio_GCAM3_ctry_Yfut <- gdp_mil90usd_GCAM3_ctry_Y %>%
+      group_by(iso) %>%
+      mutate(base = value[year == socioeconomics.FINAL_HIST_YEAR]) %>%
+      ungroup() %>%
+      filter(year %in% FUTURE_YEARS) %>%
+      transmute(iso, year, ratio = value / base)
+
+    #Use these ratios to build the GDP trajectories by old GCAM3
```
bpbond: `# Use`

R/zchunk_L242.Bld_Inc_Elas_scenarios.R
```diff
@@ -73,6 +78,20 @@ module_socioeconomics_L242.Bld_Inc_Elas_scenarios <- function(command, ...) {
...
                          "L102.pcgdp_thous90USD_Scen_R_Y") })
 
+    # L242.IncomeElasticity_bld_GCAM3: building sector income elasticity for GCAM 3.0 socioeconomics
+    # For the GCAM 3.0 scenario, calculate the per-capita GDP
+    L242.IncomeElasticity_bld_GCAM3 <- L102.gdp_mil90usd_GCAM3_R_Y %>%
+      left_join_error_no_match(L101.Pop_thous_GCAM3_R_Y, by = c("GCAM_region_ID", "year")) %>%
+      filter(year %in% FUTURE_YEARS) %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID") %>%
+      transmute(region, year, pcgdp_90thousUSD = value.x / value.y) %>%
+      mutate(income.elasticity = approx(x = A42.inc_elas$pcgdp_90thousUSD, y = A42.inc_elas$inc_elas,
```
bpbond: Why using `approx` here? Can we tweak `approx_fun` to let you use that?

bpbond:
@pralitp Do you mind reviewing, since Haewon is away I think?

R/zchunk_L242.Bld_Inc_Elas_scenarios.R
russellhz: We are basing the interpolation on an assumption file. To use approx_fun, I think I would need to bind the rows of the assumption files, probably after doing a `repeat_add_columns` for all regions, then `approx_fun`, then remove the assumption file data? Seems messy, but let me know if you have a cleaner method. Maybe adding all unique `pcgdp_90thousUSD` to the assumption file then doing approx_fun there would work.

This is also done in the file previously and in L252

bpbond: Ugh, does seem messy. OK maybe just add a comment explaining better why we're doing it this way. Thanks.

bpbond:
Holding on #656

bpbond:
Runs fine on my system, so merging even though Travis is borked ( #656 ).

Issue #517, opened 2017-07-05
L223.electricity


bpbond:
Needed for `L2233.electricity_water` #410 

cc @mihejazi @mwisepnnl

bpbond:
@pralitp this chunk returns both logit tables and data frames; following our model with #476 I'm only going to deal with the latter, OK?

bpbond:
@pkyle this chunk uses _lots_ of inputs and I could use help in assigning `Units` and `Source` for them...
<img width="229" alt="screen shot 2017-07-05 at 10 08 49 pm" src="https://user-images.githubusercontent.com/1956468/27892596-9015e0ce-61ce-11e7-92c3-2c6d0c55dc12.png">

pkyle:
@bpbond shall I make the changes directly to your branch, since there are so many? Also happy to post the changes as comments here if that's easier.

bpbond:
No, sure, if you're willing that sounds good.  I've just published the branch (`L223.electricity`). Thanks for your help!

pkyle:
Yup, I got it! Will try to get this done today

pkyle:
I just pushed my edits up to the branch (in case you don't get notifications)

bpbond:
Hugely helpful--thanks so much.

bpbond:
@cwroney Would you be willing to take this over for a while? I started on it, but got pulled into other things (including current vacation). The bad thing is it's a monster--500+ lines.

cwroney:
@bpbond Happy to take this over!

bpbond:
That would be great, even if you end up passing it back to me later. The branch (`L223.electricity`) is pushed to GitHub.

cwroney:
[...]

PR #653, opened 2017-08-17
Add a module helper fill_exp_decay_extrapolate

pralitp:
Which fills the need of the old data system's gcam_interp(rule=3).  This is needed for #517.  It replicates all of the functionality of the old data system including:
* Linearly interpolate between given values
* Use an exponential decay function to extrapolate (for rows in which the parameters are given)
* Allow "shadowing" another technology for when an improvement rate should be related across technologies.


R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+             if(has_name(., "improvement.shadow.technology")) improvement.shadow.technology else as.character(NA)) ->
+    d
+
+  # The first step is to linearly interpolate missing values that are in between
+  # values which are specified (approx_fun rule=1)
+  d %>%
+    gather(year, value, matches(YEAR_PATTERN)) %>%
+    mutate(year = as.integer(year)) %>%
+    complete(nesting_(select(., -year, -value)), year = union(year, out_years)) %>%
+    group_by_(.dots=paste0('`',names(.)[!(names(.) %in% c("year", "value"))], '`')) %>%
```
bpbond: Since this is the only use of `group_by_` in the entire package, perhaps `dplyr::group_by_` here and remove from line 382? Similar comment may apply to `nesting_`.

pralitp:
If the overall strategy looks ok I'll go ahead an update the wiki as well.

On bothersome thing is if data passed to this function isn't using the extrapolation and/or "shadowing" feature it generates warning messages that I'm not sure how to suppress apart from setting options which seems like overkill:

```
Warning message:
In max(numeric(0)[!is.na(numeric(0))]) :
  no non-missing arguments to max; returning -Inf
```

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+  assert_that(has_name(d, "improvement.max"))
+  # either improvement.rate/max are both NA or neither are
+  assert_that(all(is.na(d$improvement.rate) == is.na(d$improvement.max)))
+
+  # Note we want to allow use of the improvement.shadow.technology feature to be
+  # optional so we will check if they have not provided the column and fill NAs
+  # if not to avoid error
+  d %>%
+    mutate(improvement.shadow.technology =
+             if(has_name(., "improvement.shadow.technology")) improvement.shadow.technology else as.character(NA)) ->
```
bpbond: Glad this works, but perhaps `if_else` more appropriate to use here?

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+    d
+
+  # The first step is to linearly interpolate missing values that are in between
+  # values which are specified (approx_fun rule=1)
+  d %>%
+    gather(year, value, matches(YEAR_PATTERN)) %>%
+    mutate(year = as.integer(year)) %>%
+    complete(nesting_(select(., -year, -value)), year = union(year, out_years)) %>%
```
bpbond: Lines 410-411 are a bit hairy...please add a comment clearly explaining what's going on.

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+    group_by_(.dots=paste0('`',names(.)[!(names(.) %in% c("year", "value"))], '`')) %>%
+    mutate(value = approx_fun(year, value, rule=1)) ->
```
bpbond: Here and throughout, spaces around `=` please

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+    d
+
+  # Rows which did not specifiy improvement.max/rate did not intend to exptrapolate
+  # with the exponential decay so split those out for now
```
bpbond: "Rows in which improvement.max/rate is not specified should not be extrapolated, so split those out for now"

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+  d %>%
+    filter(is.na(improvement.max) | is.na(improvement.rate)) ->
+    d_no_extrap
+
+  d %>%
+    setdiff(d_no_extrap) ->
```
bpbond: Huh, did not know you could do this.

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+  # First partition the technologies that are not "shadowing" another technology
+  # and apply the exponential decay extrapolation to them
+  d_extrap %>%
+    filter(is.na(improvement.shadow.technology)) %>%
+    # figure out the last specified year from which we will be extrapolating
+    mutate(year_base = max(year[!is.na(value)])) %>%
+    # fill out the last specified value from which we will be extrapolating
+    mutate(value_base = value[year == year_base]) %>%
+    # calculate the exponential decay to extrapolate a new value from the last
+    # specified value
```
bpbond: Good comments here, thanks

R/module-helpers.R
```diff
@@ -353,3 +353,120 @@ get_ssp_regions <- function(pcGDP, reg_names, income_group,
...
+    mutate(value = if_else(is.na(value),
+                           value_base*improvement.max+(value_base-value_base*improvement.max)*(1.0-improvement.rate)^(year-year_base),
```
bpbond: Spaces around math operators pls

bpbond:
Is this actually hooked into `approx_fun(..., rule = 3)` yet?

bpbond:
@pralitp Re warning, I think you can do this at line 451?
```R
    # (adding a -Inf in case there are no extrapolation years, to avoid a warning)
    mutate(year_base = max(c(-Inf, year[!is.na(value)]))) %>%
```

R/module-helpers.R
pralitp: Done.  But do you have any suggestions for doing the `group_by` better?  I feel like this comes up often: I would like to group by everything except two columns such as a `year` and `value` but there doesn't appear to be any good shorthand for doing this (it seems maybe they have added something in dplyr 0.7 but I'm not sure if folks are even using that, I certainly am not).

bpbond: It's the crappy side of `dplyr`, for sure. @rplzzz may have a suggestion.

R/module-helpers.R
pralitp: I was / still am struggling `if_else` to work here.  Ultimately I'm not sure why I was doing it this way and just moved the `if` block to the outside of the pipeline which I think cleans things up.

pralitp:
Thanks for the hint for dealing with the warning.  It worked like a charm.

I didn't add this to the `approx_fun(..., rule = 3)` since that was written as a pipeline helper and there is a bunch more going on here.  It would really need access to many more columns than just `year` and `value` so I thought it would be best to leave it out and instead write in the wiki if you see `rule=3` then use `fill_exp_decay_extrapolate` instead.

We could probably get `approx_fun(..., rule = 3)` working if we really wanted but we would have to add some more arguments to it to be able to shoehorn it in.

bpbond:
>We could probably get approx_fun(..., rule = 3) working if we really wanted but we would have to add some more arguments to it to be able to shoehorn it in.

No, I think you're absolutely right.

pralitp:
@bpbond, I've updated the wiki in `Name-That-Function` for guidance on how to deal with `rule=3` (as well as 1/2 since I didn't see it there).

But then I came back to notice the CI failed.  The test package, check package all work locally but on travis there are 4 failures some of which don't even have to do with my changes.  Do you see any clues?

bpbond:
Thanks for the wiki update. Re test failures--it doesn't have any problem on my system. My guess would be that some of the slightly exotic dplyr you've used is dying on Linux; we have seen inconsistencies like this before.

I'd suggest either wait for @rplzzz to get back and test on his Linux box, or rewrite using simpler constructs. Some good 'ol print statements may help isolate exactly where this is.

bpbond:
...hmm, maybe not. Ryna's #654 is failing too.

pralitp:
Yea, it seems every PR has got a red X at the moment.  I'll try to update my environment to match travis and see if I can't replicate the failure then.

PR #657, opened 2017-08-18
Fix line endings in csv files.

rplzzz:
Clean up a bunch of broken line endings in CSV input data.  All currently existing CSV files now have unix line endings.


bpbond:
Thanks @rplzzz ! Save that script.

PR #660, opened 2017-08-18
Calculate the minimum sequence of chunks for stop_before/after

pralitp:
Come up with and constraint to only running the minimum chunks to fulfill a stop_before/after call to the driver.

### DIFF
--- a/R/driver.R
+++ b/R/driver.R
@@ -111,7 +111,7 @@ driver <- function(all_data = empty_data(),
                    return_data_names = union(inputs_of(return_inputs_of),
                                              outputs_of(return_outputs_of)),
                    write_outputs = TRUE, outdir = OUTPUTS_DIR, xmldir = XML_DIR,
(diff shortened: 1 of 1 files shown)


pralitp:
@bpbond, @rplzzz this isn't ready to get pulled but it was a first cut to come up with a minimum sequence of chunk calls in the driver when doing a stop_before/after which is becoming quite helpful for me.  I figured I'd just open this to share.  The biggest thing is it would require the package `igraph` which currently is only suggested I believe.

bpbond:
That's a cool first cut at this. I'm not crazy about adding another package dependency, but given that it's already suggested–and we'll have lots more similar problems when we beef up the tracing facility, for example–it may make sense. Unless we want to roll our own `upstream_dependencies(x)` and similar functions.

bpbond:
Closing in preparation for running git bfg. Opened #964 so we can come to it.

Issue #445, opened 2017-06-05
L241.en_newtech_nonco2


bpbond:
OK @cahartin ?

cahartin:
@bpbond i have a few more level 1's left.  but if this one looks good, go ahead!

bpbond:
@kdorheim Is this really in progress?

kdorheim:
@bpbond Yes, I started working but then you posted to the dsr channel to w.r.t. for `L112.ghg_tgej_R_en_S_F_Yh.csv` so then I moved on to my next assigned chunk.

bpbond:
Huh?  😕

kdorheim:
> current `master` is throwing an error -- I didn't check fully before merging. Sorry about that. We'll fix it, but if you see an error w.r.t. `L112.ghg_tgej_R_en_S_F_Yh.csv`, sorry about that, ignore.

kdorheim:
`L241.en_newtech_nonco2` uses that .csv and I thought that's what it meant to do. opps sorry. I will finish it soon.

CLynchy:
@bpbond When running the driver, I get the following error: 
` Error in left_join_error_no_match(., select(L241.co_tgej_R_en_S_F_fy,  : 
  left_join_no_match: NA values in new data columns `

bpbond:
Hi Cary, it looks like someone put a couple of wrong files into `inst/extdata`. I'll fix.

CLynchy:
Was it me?  I blame CH.

bpbond:
I am determinedly NOT running `git blame` here. Let's just say the files appeared mysteriously.

bpbond:
OK, pushed commits. I have (i) removed the `temp-data-inject/` files entirely, so `L241` now uses the output from your `L211`.

Issue #650, opened 2017-08-16
L241.en_newtech_nonco2 TODO 


kdorheim:
`L241.en_newtech_nonco2.R` in the old ds contains the following

> #TODO: better way to handle this, probably these technologies should pull from historical data

(The code currently replaces 1975 technology coefficients with `may.be.historic` values from an assumption file.)

Is this something we want address later on?

russellhz:
@ssmithClimate any thoughts on this issue? For some further context, this chunk is outputting emission coefficient tables for new energy technologies. The majority of these emissions come from `emissions/A41.tech_coeff`, although some coefficients for CO and CH4 come from historical data calculated by previous chunks.

Then some of the years (which are by default 2020) are replaced with 1975 if the "newtech" was actually used historically. 

The todo is before this year replacement on line 175 (note that I don't think the second part of the comment is true, as it looks like all regions get overwritten identically):

    # TODO: better way to handle this, probably these technologies should pull from historical data
    # Ensure only regions that have first gen biofuels get read in
    # See GitHub issue #650


    # Not all of the regions started using the new technologies, "newtech", at the same time,
    # some regions started using them in historical years whereas other regions started
    # using them in future years. For the regions that may have been using these
    # technologies in historical years replace the first model base year with the maybe
    # historical value to include the historical "newtech" nonco2 parameters in the historical data.

PR #613, opened 2017-08-05
Cary - NEW L111.nonghg_en_R_S_T_Y

bpbond:
#482 

This replaces PR #610

PR #661, opened 2017-08-18
Kd - L241.en_newtech_nonco2

kdorheim:
#445


kdorheim:
I am working on resolving the merge conflicts right now

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
 #' module_emissions_L241.en_newtech_nonco2
 #'
-#' Briefly describe what this chunk does.
+#' Non-CO2 emission coefficients.
```
bpbond: Need something more descriptive and clearer here, please.

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
...
     # Load required inputs
     GCAM_region_names <- get_data(all_data, "common/GCAM_region_names")
-    A_regions <- get_data(all_data, "emissions/A_regions")
-    A_regions <- get_data(all_data, "energy/A_regions")
+    Emiss_A_regions <- get_data(all_data, "emissions/A_regions")
+    En_A_regions <- get_data(all_data, "energy/A_regions")
     A41.tech_coeff <- get_data(all_data, "emissions/A41.tech_coeff")
     A51.max_reduction <- get_data(all_data, "emissions/A51.max_reduction")
     A51.steepness <- get_data(all_data, "emissions/A51.steepness")
-    L111.nonghg_tgej_R_en_S_F_Yh <- get_data(all_data, "L111.nonghg_tgej_R_en_S_F_Yh")
```
bpbond: This will be available with #613 ...

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
...
+      select(region, supplysector, subsector, stub.technology, Non.CO2, year) %>%
+      filter(year == min(FUTURE_YEARS)) %>%
+      mutate(ctrl.name = "GDP_control") ->
+      L241.nonco2_steepness
+
+    # Combine closet future max reduction emission coefficients with the assumed steepness
+    # of emission reductions into one data frame.
+    L241.steepness %>%
+      select(steepness = value, region, supplysector, subsector, stub.technology, Non.CO2) %>%
+      left_join(L241.nonco2_steepness, by = c("region", "supplysector", "subsector",  "stub.technology", "Non.CO2")) %>%
```
bpbond: Can't use `left_join_error_no_match` here?

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
...
+    # Rename SO2 to regional SO2.
+    L241.nonco2_tech_coeff <- rename_SO2(L241.nonco2_tech_coeff, Emiss_A_regions, FALSE)
+    L241.nonco2_max_reduction <- rename_SO2(L241.nonco2_max_reduction, Emiss_A_regions, FALSE)
+    L241.nonco2_steepness <- rename_SO2(L241.nonco2_steepness, Emiss_A_regions, FALSE)
+
+
+    # TODO: better way to handle this, probably these technologies should pull from historical data
+    # "Ensure only regions that have first gen biofuels get read in"
+    # (This is where the old data system had the note - TODO: better way to handle this,
+    # probably these technologies should pull from historical data. I opened an issue about it.)
```
bpbond: Reference specific issue number.

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
...
+
+    # Create a vector of the first generation biofuel technologies this vector will be used to
+    # remove non-applicable first-genbio technologies from the data frames.
+    L241.firstgenbio_techs <- c("corn ethanol", "sugarbeet ethanol", "sugar cane ethanol", "biodiesel")
+
+    # Create region ethanol type and region biodiesel type identifiers from the
+    # region-specific assumptions in the energy system. These identifiers will be
+    # used to remove non-applicable first-genbio technologies.
+    En_A_regions %>%
+      unite(col = region_eth, region, ethanol, sep = "~", remove = FALSE) %>%
```
bpbond: What's the significance of `~`? Curious separator.

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,286 @@
...
-    # If no precursors (very rare) don't call `add_precursor` at all
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L241.nonco2_tech_coeff %>%
+      add_title("Non-CO2 new technology emission coefficients by supply sector") %>%
+      add_units("NA") %>%
+      add_comments("Pass through may be historic values into the non CO2 new technology emission coefficients") %>%
```
bpbond: Confusing comment.

inst/extdata/emissions/A41.tech_coeff.csv
```diff
@@ -0,0 +1,36 @@
+# File: A41.tech_coeff.csv
+# Title: NA
+# Units: NA
+# Source: NA 
+# Description: NA,,,,,,,,,,,,,
```
bpbond: Remove trailing commas please

inst/extdata/emissions/A41.tech_coeff.csv
```diff
@@ -0,0 +1,36 @@
+# File: A41.tech_coeff.csv
+# Title: NA
```
bpbond: Please provide a descriptive title

R/zchunk_L241.en_newtech_nonco2.R
kdorheim: Because this will become available with #613 should I remove the temp-data-inject call?

R/zchunk_L241.en_newtech_nonco2.R
kdorheim: Um I always use the `~` as separator if I have to do a unite and separate. In the past I have run into issues when one of the variable being united/separated contains as space , `-` , or `_` already but I have yet to run into `~` existing in a variable name.

 If you want I can use something else.

R/zchunk_L241.en_newtech_nonco2.R
bpbond: No, thanks. (Unless that one merges first.)

R/zchunk_L241.en_newtech_nonco2.R
bpbond: If it works for you and here, ok!

bpbond:
This is holding on a review by @cahartin or someone else.

R/zchunk_L241.en_newtech_nonco2.R
```diff
@@ -1,107 +1,282 @@
...
+    # coefficients in the min model base year by a supplysecotr, subsector, stub.technology identifier
+    # called sector_tech_id.
+    A41.tech_coeff %>%
+      select(supplysector, subsector, stub.technology, may.be.historic) %>%
+      na.omit %>%
+      unite(sector_tech_id, supplysector, subsector, stub.technology, remove = FALSE) ->
+      L241.maybe_historic
+
+    # Replace the new technology emission coefficients in the model base year with the may be historic
+    # values from the technology assumption file by the unique supplysector, subsector,
```
cahartin: "with the may be historic values"?

bpbond: I was going to just fix and merge this but have to agree–it's a confusing comment, and the `may.be.historic` flag isn't documented in the `A41.tech_coeff.csv` header either. Otherwise, this PR is good to go.

kdorheim: Sorry about the vague commenting, I have no idea what the `may.be.historic` values are or how to better describe what is going on here.  The old ds had a the following comment `Allow some future techs that may be in the historical years` would that make more sense?

bpbond: I'd say ask, probably @pkyle but perhaps someone else, for clarification.

pkyle: OK I think I understand this now...these are technologies that are available in most or all regions in future years, but only in the historical years in some cases. For example, coal to liquids exists in the base years in South Africa, whereas most everywhere else it's only allowed in future years. By default, these "newtech" nonco2 parameters start in the first future model time period (2015), but because some of these "newtechs" can be in the historical data in some regions and years, their nonco2 data are instead started (generally, in all regions) in the first model base year (1975).
That said, I don't see any purpose to these steps; all of the nonco2 data could have just been assigned to 1975 for all technologies, with no consequences b/c where output is zero, emissions are also zero. This step also doesn't reduce the file size of the XML, b/c each tech's nonco2 data is only read into one year, and the model internally copies it forward to all subsequent years. That said, my first guess is that the motivation for this step was reducing the size of the XML files. My second guess is that these technologies were intended to be excluded from getting emissions coefs in the base years, because they weren't considered in the mappings from inventory emissions data to GCAM technologies. So, having non-zero historical coefs means that they will produce emissions, so then the model output won't match the emissions inventory data. But again, what's done here would achieve the opposite of that outcome, so really I don't know!

bpbond: Thanks, Page! Kalyn, can you distill that down to a useful comment? If not ask @pkyle to say the same thing but using fewer words :)

kdorheim: I think so. This is the part of the chunk that I think the comment 
``` {r} 
# TODO: better way to handle this, probably these technologies should pull from historical data
# Ensure only regions that have first gen biofuels get read in
# See GitHub issue #650
 ```
is referring to. @pkyle & @bpbond do we want to figure out a better way to do this now or wait?

bpbond: No, no need now. Let's wrap this up–making sure the comment here is clear and in A41 too–and then get 'er merged.

bpbond:
This is failing because of #677 but otherwise looks and builds fine on my system. Merging. Thanks @kdorheim .

Issue #301, opened 2017-04-28
Test gcamdata against dplyr 0.7.0


bpbond:
https://blog.rstudio.org/2017/04/13/dplyr-0-6-0-coming-soon/

Install the pre-release version with `devtools::install_github("tidyverse/dplyr")`.

Reading the notes, it doesn't _look_ like we'll necessarily break, but it would be great to test before mid-May.

I will try to get to this--but if anyone would like to experiment, see `devtools::install_version`, have fun, and please report back what you find.

bpbond:
https://blog.rstudio.org/2017/06/13/dplyr-0-7-0/

PR #664, opened 2017-08-22
Fix dplyr 0.7

### LINKED ISSUE #656: Chunks fail to run with latest version of dplyr
It looks like the CI failures are coming from updating some package and is causing failures in chunks that previously ran just fine.  I'm not totally sure which package but most likely candidate is dplyr.  This is what I saw when running the driver:

```
[1] "module_emissions_L113.ghg_an_R_S_T_Y"
Adding missing grouping variables: `EDGAR_agg_sector`
[1] "- make 1.32"
[1] "module_emissions_L115.nh3_an_R_S_T_Y"
[1] "- make 1.13"
[1] "module_emissions_L121.nonco2_awb_R_S_T_Y"
[1] "- make 5.38"
[1] "module_emissions_L1211.nonco2_awb_R_S_T_Y_IRR"
[1] "- make 1.73"
[1] "module_emissions_L122.ghg_agr_R_S_T_Y"
[1] "- make 2.40"
[1] "module_emissions_L123.bcoc_awb_R_S_T_Y"
[1] "- make 0.15"
[1] "module_emissions_L124.nonco2_unmgd_R_S_T_Y"
[1] "- make 1.74"
[1] "module_emissions_L125.bcoc_unmgd_R_S_T_Y"
`summarise_each()` is deprecated.
Use `summarise_all()`, `summarise_at()` or `summarise_if()` instead.
To map `funs` over all variables, use `summarise_all()`
[1] "- make 0.62"
[1] "module_emissions_L131.nonco2_proc_R_S_T_Y"
[1] "- make 3.36"
[1] "module_emissions_L152.MACC"
[1] "- make 0.58"
[1] "module_emissions_L211.ag_nonco2"
Adding missing grouping variables: `GCAM_region_ID`, `EDGAR_agg_sector`
[1] "- make 1.25"
[1] "module_emissions_L2111.ag_nonco2_IRR"
[1] "- make 2.01"
[1] "module_emissions_L232.prc_nonco2"
 Show Traceback
 
 Rerun with Debug
 Error in mutate_impl(.data, dots) : 
  Column `Non.CO2` can't be modified because it's a grouping variable 
```

pralitp:
Fixes to be able to run with dplyr 0.7.  Mostly revolving around `group_by` that linger longer than they should.  In the latest version allowing them to linger can result in new `NA` filled columns as the result of merges and `bind_rows`.

To help keep these under wraps `return_data` will now enforce that chunks can not return grouped tibbles.

This closes #656


R/utils-data.R
```diff
@@ -217,6 +217,16 @@ get_data <- function(all_data, name) {
 return_data <- function(...) {
   dots <- list(...)
   names(dots) <- as.list(substitute(list(...)))[-1L]
+  # disallow any data which is "grouped" as it may lead to unexpected
+  # behavior, especially for unsuspecting chunks which may use it down
+  # the line not expecting any groupings.
+  lapply(names(dots), function(dname) {
+    # note we wrap the assert in a tryCatch since groups will throw an
```
bpbond: Would it be clearer to do
```R
if(is_tibble(dots[[dname]])) {
  assert_that(is.null(groups(dots[[dname]])))
}
```
?

bpbond:
Note this will also close #301 .

R/utils-data.R
pralitp: Yea, I think so.  When I was looking at the documentation for `group_by` it sounded like it supported many different types of tables and when I check `is_tibble(data.table::data.table())` it returned false.  Thus I wasn't sure how to check all possible types and so put in the `tryCatch`. 
But as I try `is_tibble(data.table::data.table(...) %>% group_by(x))` it sure enough does report `true`.  And of course if it didn't get grouped then we don't really need to do the check anyway.
I'll switch that and update the comments.

pralitp:
FYI, I just tried this on my laptop and got the error:
```
Error: assert_that: assertion must return a logical value 
```

Updating `assertthat` from v0.1 to v0.2 seems to fix that.  I'm not sure what was wrong with the `assert_that` call otherwise.  Any ideas, or should we just advise (and update package requirements) users to upgrade?

bpbond:
`assert_that(is.null(...))` gave that error? Bizarre.

I'd suggest we force upgrading to `assertthat` 0.2 then.

bpbond:
Note two chunks were still returning grouped tibbles and thus erroring the driver. I fixed these in b773845, just committed.

Issue #637, opened 2017-08-14
Need help with L231.proc_sector assumption file units


russellhz:
`module_emissions_L231.proc_sector` reads in a bunch of assumption files with no units or sources. The ones I'm not sure about:

- Cost: `A31.globaltech_cost`
- Coefficients: `A31.globaltech_coef`
- Efficiency: `A31.globaltech_eff` and in energy `A32.globaltech_eff`
- Other: `A31.rsrc_info`

- Logit and shareweight: I've been listing these as NA/Unitless

ssmithClimate:
I don't have anything to add on these files. Logit and share weight are, indeed, unitless (so use that instead of NA).

bpbond:
Thanks, Steve.

PR #665, opened 2017-08-23
Russell-L231.proc_sector

russellhz:
#579

Could still use some units help, per #637. Some of the documentation needs work, as I'm not totally sure what all these outputs actually represent


R/zchunk_L231.proc_sector.R
```diff
@@ -0,0 +1,460 @@
...
+
+    # ===================================================
+    # L231.FinalDemand_urb: Final demand information for urban processes sector
+    L231.FinalDemand_urb <- tibble(
+      region = A_regions$region,
+      energy.final.demand = "urban processes",
+      perCapitaBased = 1,
+      income.elasticity = 0,
+      base.service = 0.004,
+      aeei = 0
```
bpbond: What is `aeei`? Where units of these values?

R/zchunk_L231.proc_sector.R
```diff
@@ -0,0 +1,460 @@
...
+    ) %>%
+      repeat_add_columns(tibble(year = BASE_YEARS))
+
+    # L231.Supplysector_ind: Supply sector information for urban & industrial processes sectors
+    L231.Supplysector_urb_ind <- A31.sector %>%
+      write_to_all_regions(c("region", "supplysector", "output.unit", "input.unit", "price.unit",
+                             "logit.year.fillout", "logit.exponent",  "logit.type"), GCAM_region_names = GCAM_region_names )
```
bpbond: Is there a `LEVEL2_DATA_NAMES` entry that can (should) be used here and below?

R/zchunk_L231.proc_sector.R
```diff
@@ -0,0 +1,460 @@
...
+
+    L231.SubsectorLogit_urb_ind %>%
+    add_title("Urban and Industrial Processes Subsector Logit Info") %>%
+    add_units("Unitless") %>%
+    add_comments("A31.subsector_logit written to all regions") %>%
+    add_legacy_name("L231.SubsectorLogit_urb_ind") %>%
+    add_precursors("emissions/A31.subsector_logit", "common/GCAM_region_names") ->
+    L231.SubsectorLogit_urb_ind
+
+    if (exists("L231.SubsectorShrwt_urb_ind")){
```
bpbond: When would this not exist? Usually by definition if `get_data` returns, it exists.

bpbond:
Probably. We can always re-do the line endings if necessary, I guess.

R/zchunk_L231.proc_sector.R
```diff
@@ -0,0 +1,460 @@
...
+                                     L1322.in_EJ_R_indfeed_F_Yh %>%
+                                       mutate(sector = "industrial feedstocks")) %>%
+      left_join_keep_first_only(Ind.globaltech_eff, by = c("sector", "fuel", "year")) %>%
+      # Calculate service as energy * efficiency
+      mutate(service = value * efficiency) %>%
+      na.omit() %>%
+      group_by(GCAM_region_ID, year) %>%
+      summarise(ind_output = sum(service)) %>%
+      ungroup() %>%
+      mutate(ind_proc_input = 0.008,
```
cahartin: @bpbond there are a lot of hard cores #s in this file.  Do you know if they are in multiple files and would best to be pulled out in the constants?

R/zchunk_L231.proc_sector.R
cahartin: # This was a mistake in the old data system - the wrong name was assigned, but because it didn't exist, no error was thrown

russellhz: There are 4 objects here that only exist and only get printed out depending on whether NA values exist in certain assumption files. 2 of them actually were made in the old data system and 2 were not (did not exist and were not made in this chunk)

R/zchunk_L231.proc_sector.R
bpbond: I'd suggest open an issue ("Lots of hardcoded numbers in L231.proc_sector") to remind us to follow up on this later.

R/zchunk_L231.proc_sector.R
russellhz: I'm not sure, but Google tells me "Autonomous Energy Efficiency Improvement (AEEI) parameter"

A lot of the values in this chunk are pretty meaningless to me, as there are mostly just a whole lot of 0's and 1's.

bpbond:
Waiting on @russellhz revisions and update to latest `master`.

russellhz:
@bpbond I added an extra comment and updated to latest master. There are still some missing units right now (answers to #637 would help greatly).

bpbond:
What's up with the "Batch file disabled" commit b648435 ?

...anyway, I'm going to test this out and probably merge, OK? We'll follow up when #637 when that information comes in.

russellhz:
Oh, that doesn't need to be disabled anymore. I had two PRs that both needed to write to the same batch file, so I disabled it since the outputs from the other chunk weren't disabled

Issue #634, opened 2017-08-14
L226.en_distribution


bpbond:

cwroney:
@bpbond This chunk has a few times when it uses `interpolate_and_melt` to filter a historical data set down to the `BASE_YEARS`. I could replicate this just using a one line `filter` or I could use the full code to complete, arrange, group, and approx_fun on the data, even though the approx_fun isn't doing anything. If I just filter it, it will only able to accept data sets where the historical data includes the base years we are using (the ones we have right now do). Is this a problem if I simplify the code, or should I keep that flexibility to interpolate if necessary?

cwroney:
to be more exact, `    L126.IO_R_elecownuse_F_Yh %>% 
      complete(nesting(GCAM_region_ID, sector, fuel), year = c(year, BASE_YEARS)) %>% 
      arrange(GCAM_region_ID, sector, fuel, year) %>%
      group_by(GCAM_region_ID, sector, fuel) %>%
      mutate (value = approx_fun(as.numeric(year), value)) %>%
      filter(year %in% c(BASE_YEARS))` would output the same result as `L126.IO_R_elecownuse_F_Yh %>% filter(year %in% c(BASE_YEARS))` given that the input data has all the relevant years listed.

bpbond:
Good lord. If that `filter` does it, and if it passes timecheck, let's do that. So much simpler and clearer.

cwroney:
@bpbond Excellent, thanks, done. Follow up: This chunk only conditionally produces four of its outputs, two of which don't exist by default. How should I handle this? Is it okay to wrap up the produce outputs section and `return_data` in `if` statements? e.g. `if (exists("L226.SubsectorShrwt_en")) {
    L226.SubsectorShrwt_en %>% add_title() ... }` and `if (exists("L226.SubsectorShrwt_en")) {return_data(L226.SubsectorShrwt_en)}` I'm assuming it's going to break other places as well.

bpbond:
Yeah this question of optional outputs is coming up again and again, e.g. #633 #655 and others. I _think_ we want to follow what @russellhz did in #665 - create a dummy table and return it. E.g.:

```R
    if (exists("L231.SubsectorShrwt_urb_ind")){
... 
   } else {
       tibble(x = NA) %>%
         add_title("Data not created") %>%
         add_units("Unitless") %>%
         add_comments("Data not created") %>%
         add_legacy_name("L231.SubsectorShrwt_ind") ->
[...]

PR #671, opened 2017-08-24
Chris-L226.en_distribution

cwroney:
Notes: There's a mysterious cost adder in A26globaltech_cost and I have no idea what it represents aside from, broadly, the cost of distribution. Also one grouped input file I commented on (assuming that's supposed to be coming in ungrouped from the other chunk). 

There was a conflict in namespace when I updated from the master branch before committing but I couldn't tell what it was on about so I went ahead with it. 

#634


bpbond:
🥇 to @cwroney for amusing commit names

R/zchunk_L226.en_distribution.R
```diff
@@ -12,12 +13,10 @@
 #' @importFrom assertthat assert_that
```
bpbond: `@details` line filled out?

R/zchunk_L226.en_distribution.R
```diff
@@ -46,11 +45,16 @@ module_energy_L226.en_distribution_DISABLED <- function(command, ...) {
              "L226.StubTechCoef_gaspipe"))
   } else if(command == driver.MAKE) {
 
+    #Silence global variable package check
```
bpbond: `# Silence`

R/zchunk_L226.en_distribution.R
```diff
@@ -65,172 +69,358 @@ module_energy_L226.en_distribution_DISABLED <- function(command, ...) {
...
+      select(-calibration, -secondary.output) ->
+      L226.IO_R_elecownuse_F_Yh
+
+    #repeat final year's ownuse ratio into future years and append future years to base years (could perhaps be tied to industrial CHP...but also AUTOELEC)
+    L226.IO_R_elecownuse_F_Yh %>%
+      filter(year == max(BASE_YEARS)) %>%
+      repeat_add_columns(tibble("year" = FUTURE_YEARS)) %>%
+      select(-year.x) %>%
+      rename(year = year.y) %>%
+      bind_rows(L226.IO_R_elecownuse_F_Yh, .) ->
```
bpbond: Why `, .`? Does the order of binding matter?

R/zchunk_L226.en_distribution.R
```diff
@@ -65,172 +69,358 @@ module_energy_L226.en_distribution_DISABLED <- function(command, ...) {
...
+      L226.IO_R_elecownuse_F_Y
+
+    #rename columns and round coefficients - L226.StubTechCoef_elecownuse: calibrated coefficients on electricity net ownuse
```
bpbond: Here and throughout, please ensure a space after `#`

R/zchunk_L226.en_distribution.R
```diff
@@ -65,172 +69,358 @@ module_energy_L226.en_distribution_DISABLED <- function(command, ...) {
...
+    L226.IO_R_electd_F_Yh %>%
+      filter(year == max(BASE_YEARS)) %>%
+      repeat_add_columns(tibble("year" = FUTURE_YEARS)) %>%
+      select(-year.x) %>%
+      rename(year = year.y) ->
+      L226.IO_R_electd_F_Yfut
+
+    # append assumed techchange value and calculate decrease in future energy use
+    L226.IO_R_electd_F_Yfut %>%
+      left_join(A_regions, by = "GCAM_region_ID") %>%
```
bpbond: Why not `left_join_error_no_match`? Are there region IDs not present in `A_regions`?

inst/extdata/energy/A26.globaltech_cost.csv
```diff
@@ -0,0 +1,17 @@
+# File: A26.globaltech_cost.csv
+# Title: Final energy delivery cost adders
+# Units: ?
```
bpbond: Please check with @pkyle or @mwisepnnl or someone about this.

bpbond:
>Notes: There's a mysterious cost adder in A26globaltech_cost and I have no idea what it represents aside from, broadly, the cost of distribution.

Also - let's clarify this if possible! Thanks.

inst/extdata/energy/A26.globaltech_cost.csv
pkyle: In general, these are the energy delivery cost adders, or the difference between producer prices and delivered energy prices to different consumer groups. The unit is 1975$/GJ, and the source for most of the items is the pre-2006 MiniCAM model inputs. Values of zero indicate that delivery costs are included elsewhere in the chain.
A26 also includes some grid reliability (backup) related costs. The backup_electricity cost represents the variable O&M cost of a gas turbine (levelized capital and fixed O&M costs are accounted elsewhere). No such cost is included here for csp_backup, because the turbine costs are already accounted in the primary electric technology costs.

R/zchunk_L226.en_distribution.R
cwroney: no, but there's one entry in biomassoil_Tech that's NA and so triggers the error when it copies it (at least I assume that's what's happening).

inst/extdata/energy/A26.globaltech_cost.csv
cwroney: Thanks, @pkyle!

R/zchunk_L226.en_distribution.R
cwroney: I don't believe so, it just seemed intuitive to me to have the future years come after the base years.

cwroney:
@bpbond There's no XML generating file with this yet. I've never done a level 2 before though -- are we handling the xml code in the same branch as the level 2 code? Certainly the outputs for this would next be used to generate en_distribution.xml.

bpbond:
I'll add the xml-generating file. Is this ready for a second review? Thanks.

cwroney:
@bpbond it is, it kept throwing merge conflicts in the namespace because the L225 export was removed in the master in the same line L226 was removed here, so I had to manually remove both, but it should be good to go now.

bpbond:
This looks good and is just waiting on @mwisepnnl comments/approval.

mwisepnnl:
Is the A26 cost adder still mysterious?  It is just as your comment states a cost adder to get from a centralized cost to a distributed cost.  I don't know if the original data system has the reference for these.  If not, maybe raise it as an issue?

cwroney:
realized I forgot to add the @details so put them in, also deleted the two misspelled variables that were generating the notes...now redundant with the correctly spelled ones you added. Also, I added the source from Page for the A26 globaltechcost. Done meddling!

bpbond:
>Is the A26 cost adder still mysterious? It is just as your comment states a cost adder to get from a centralized cost to a distributed cost. I don't know if the original data system has the reference for these. If not, maybe raise it as an issue?

Good point–thanks. @cwroney please open an issue to follow up on this if you think useful.

bpbond:
>now redundant with the correctly spelled ones you added

Haha, I missed that. Anyway thanks for all your work, looks good, merging as soon as tests pass.

cwroney:
@bpbond Page said the A26 cost adders were from "pre-2006 MiniCAM model inputs", so I added that as the source. I'm not sure if we need more specific than that, since that's kind of saying that our data source is ourselves in the past?

bpbond:
I'm willing to bet that means, practically speaking, "lost in the mists of time". Please do open an issue.

Issue #677, opened 2017-08-28
Import error in module-helpers


bpbond:
Line 426 in `module-helpers.R`, recently added by you, is generating a package check error:

```R
    nesting_vars <- rlang::syms(names(d)[!(names(d) %in% c("year", "value"))])
```
>checking dependencies in R code ... WARNING
>'::' or ':::' import not declared from: ‘rlang’

I don't even _have_ an `rlang` package on my system; what is this?

pralitp:
It is added with dplyr 0.7 to do.. what I used it for.  I guess it is back to the drawing board for #656.  I could go back to using `nested_` even though it is deprecated.

bpbond:
I figured. Sorry to bug you but since this is causing every build to error it's a high-priority fix if possible. Thanks.

pralitp:
In that case I'm going to commandeer this issue and leave it open to come up with a better long term solution.  In the short term I'll push up a fix to just use the deprecated function.

bpbond:
See also #747 .

PR #678, opened 2017-08-28
Revert to using the deprecated nesting_ for dplyr 0.7

pralitp:
since otherwise we get package check errors with dplyr 0.5 about not package rlang.  In the long term we can come up with a better solution. See #677


bpbond:
I think the long term will be to force everyone to dplyr 0.7, but for now, thanks!

bpbond:
Hmm. This checks out fine on my system, but is failing on Travis.

PR #685, opened 2017-08-31
Ryna - L2242.land_input_4_irr_mgmt

rynacui:
#685


rynacui:
I've tried making the batch chunk, but it fails the driver and I couldn't figure it out why. It is disabled now.

R/zchunk_L2242.land_input_4_irr_mgmt.R
```diff
@@ -1,90 +1,120 @@
 #' module_aglu_L2242.land_input_4_irr_mgmt
 #'
-#' Briefly describe what this chunk does.
+#' This chunk generates files of the fourth land nest by region.
```
bpbond: What does "fourth land nest" mean? Can we provide something clearer or more specific?

R/zchunk_L2242.land_input_4_irr_mgmt.R
```diff
@@ -1,90 +1,120 @@
...
-    L2241.LN4_Logit <- get_data(all_data, "L2241.LN4_Logit")
-    L2241.LN4_NodeGhostShare <- get_data(all_data, "L2241.LN4_NodeGhostShare")
-    L2241.LN4_NodeIsGhostShareRel <- get_data(all_data, "L2241.LN4_NodeIsGhostShareRel")
+    A_LandNode_logit_irr <- get_data(all_data, "aglu/A_LandNode_logit_irr")
+    L223.LN3_MgdAllocation_bio <- get_data(all_data, "temp-data-inject/L223.LN3_MgdAllocation_bio")
+    L223.LN3_MgdAllocation_crop <- get_data(all_data, "temp-data-inject/L223.LN3_MgdAllocation_crop")
+    L223.LN3_LeafGhostShare <- get_data(all_data, "temp-data-inject/L223.LN3_LeafGhostShare")
+    L223.LN3_LeafIsGhostShareRel <- get_data(all_data, "temp-data-inject/L223.LN3_LeafIsGhostShareRel")
+
+    names_LN3_Leaf <- c("region", "LandAllocatorRoot", "LandNode1", "LandNode2", "LandNode3", "LandLeaf")
```
bpbond: Not using `LEVEL2_DATA_NAMES`?

R/zchunk_batch_land_input_4_IRR_MGMT.xml.R
```diff
@@ -28,9 +28,10 @@ module_aglu_batch_land_input_4_IRR_MGMT.xml_DISABLED <- function(command, ...) {
 
     # Produce outputs
     create_xml("land_input_4_IRR_MGMT.xml") %>%
-      add_xml_data(L2242.LN4_Logit,"LN4_Logit") %>%
+      add_logit_tables_xml(L2242.LN4_Logit,"LN4_Logit") %>%
```
bpbond: Looks like the error is occurring in this call: `add_logit_tables_xml` is expecting a `logit.type` column in `L2242.LN4_Logit` but there's not one.

@pralitp we should probably add a 
```R
  assert_that("logit.type" %in% names(data))
```
at the beginning of `add_logit_tables_xml`, yes? And what's the best/correct way for @rynacui to fix this? Thanks.

R/zchunk_L2242.land_input_4_irr_mgmt.R
rynacui: `LN3_Leaf` currently is not in the `LEVEL2_DATA_NAMES` function. I tried adding it to the function file myself, but it seems not doing anything after rebuilding `gcamdata` and running the driver.

bpbond: Another good question for @pralitp

pralitp: Did you source the data file after update: `source("data-raw/level2_data_names.R")`

The call to `devtools::use_data(LEVEL2_DATA_NAMES, overwrite = TRUE)` at the bottom is important.  We set it up this way to follow the guidelines for handling data in R packages here: http://r-pkgs.had.co.nz/data.html

R/zchunk_batch_land_input_4_IRR_MGMT.xml.R
pralitp: @bpbond, good idea about the assert.  I'll open a PR.

About the missing column I think in we want to change line 67 in module_aglu_L2242.land_input_4_irr_mgmt  to (with the caveat that I haven't actually run it but I'm pretty sure):
```R
      select(one_of(c(LEVEL2_DATA_NAMES[["LN4_Logit"]], "logit.type"))) %>%
```

R/zchunk_L2242.land_input_4_irr_mgmt.R
rynacui: Got it. Thanks @pralitp

R/zchunk_L2242.land_input_4_irr_mgmt.R
kvcalvin: Maybe note what sort of data is in the file (land allocation, carbon) and what type of land (nodes/leafs, crops)

R/zchunk_L2242.land_input_4_irr_mgmt.R
```diff
@@ -1,90 +1,120 @@
...
 #' @param command API command to execute
 #' @param ... other optional parameters, depending on command
 #' @return Depends on \code{command}: either a vector of required inputs,
 #' a vector of output names, or (if \code{command} is "MAKE") all
-#' the generated outputs: \code{L2241.LN4_LogitTables[[ curr_table_name ]]}, \code{L2242.LN4_Logit}, \code{L2242.LN4_NodeGhostShare}, \code{L2242.LN4_NodeIsGhostShareRel}. The corresponding file in the
+#' the generated outputs: \code{L2242.LN4_Logit}, \code{L2242.LN4_NodeGhostShare}, \code{L2242.LN4_NodeIsGhostShareRel}. The corresponding file in the
 #' original data system was \code{L2242.land_input_4_irr_mgmt.R} (aglu level2).
-#' @details Describe in detail what this chunk does.
+#' @details This chunk generates the logit exponent of the fourth land nest by region, the ghost node share for bionenergy node in future years,
+#' and specifies whether the bionenergy ghost node share is relative.
```
kvcalvin: Is that really it? Or does it have land allocation?

kvcalvin: Never mind. I guess that is all it is doing.

R/zchunk_L2242.land_input_4_irr_mgmt.R
bpbond: @pralitp If you get a chance please add this info to the wiki.

R/zchunk_batch_land_input_4_IRR_MGMT.xml.R
rynacui: Shall we use old data system behavior for
`select(one_of(c(LEVEL2_DATA_NAMES[["LN4_Logit"]], "logit.type"))) %>%`
since `logit.type` is not in the old output?

pralitp: Nope, this was a conscious decision on our part not to match the old data system about keeping the `logit.type` column.  The old / new tests will ignore that discrepancy.

R/zchunk_L2242.land_input_4_irr_mgmt.R
pralitp: I'm on it

pralitp: @bpbond, oops.  It looks like I already did but it is part of the XML conversion page.  Should it go else where?

R/zchunk_L2242.land_input_4_irr_mgmt.R
```diff
@@ -1,90 +1,119 @@
...
+             LandNode4 = sub("RootTuber", "Root_Tuber", LandNode4),
+             LandNode4 = sub("biomasstree", "biomass_tree", LandNode4),
+             LandNode4 = sub("biomassgrass", "biomass_grass", LandNode4)) %>%
+      # Match in logit exponent values, use left_join instead because the logit.type variable are NAs, drop later
+      left_join(A_LandNode_logit_irr, by = c("LandNode4" = "LandNode")) %>%
+      mutate(LandNode4 = paste(LandNode4, GLU_name, sep = "_")) %>%
+      select(one_of(LEVEL2_DATA_NAMES[["LN4_Logit"]])) %>%
+      # Note: The na.omit is intended to get rid of a tailing node-rename table
+      # that may be appended to level2 land alloator tables
+      na.omit ->
```
rynacui: By adding the `logit.type` which are NAs, `na.omit` drops every single row in the tibble. 
It currently does not make any difference. Is it safe to drop this line? @bpbond @pralitp

pralitp: Judging by that comment and that we no longer to worry about any `node-rename` issues in DSR I would say yes go ahead and drop it and the comment which is no longer applicable.

PR #694, opened 2017-09-01
Xml verify

rplzzz:
Add a utility to compare output XML files to the corresponding files from the original data system.  The comparison utility is written in python, but there is a test case that runs the comparison on each output file produced by the system as part of the overall test suite.  Because the comparison data is too large to include in the repository, the system looks for the location of the comparison in an option called `gcamdata.xml_cmpdir`.  If the option is not set, then the XML comparison test is skipped.

The comparison can also be run stand-alone from the command line.  This is useful when you know a file is failing its comparison, and you want to capture the differences to diagnose the problem.  There is also a shell script to run the comparison on all of the data system output files, for cases where you want to run all the comparisons, but not the rest of the package tests.  The script takes the new data directory and the old data directory as arguments and loops over all the files it finds in the new data directory, finding the corresponding comparison file and running the comparison.


bpbond:
Thanks for writing tests @rplzzz .

bpbond:
@rplzzz I'm getting a whole bunch of test errors

```
Failed -------------------------------------------------------------------------
1. Error: Identical XML files are equivalent (@test_xml_diff.R#8) --------------
error in running command
1: expect_silent(cmp_xml_files(basefile, basefile, raw = TRUE)) at /Users/d3x290/Documents/Work/Code/gcamdata/tests/testthat/test_xml_diff.R:8
2: evaluate_promise(object)
3: with_sink(temp, withCallingHandlers(withVisible(code), warning = handle_warning, 
       message = handle_message))
4: withCallingHandlers(withVisible(code), warning = handle_warning, message = handle_message)
5: withVisible(code)
6: cmp_xml_files(basefile, basefile, raw = TRUE)
7: system2(cmd, args, stdout = TRUE) at /Users/d3x290/Documents/Work/Code/gcamdata/R/xml.R:250
```

Did you forget to add `exec/xml_verify.py` to the PR?

rplzzz:
@bpbond `xml_verify.py` is in the repository:

https://github.com/JGCRI/gcamdata/blob/xml-verify/inst/exec/xml_verify.py

Also, it's passing on Travis, so there must be an issue with your local setup.  "Error in running command" is a little vague.  Does that indicate that it couldn't find the python program?

rplzzz:
Update for anyone following this PR:

I have a conjecture as to why the tests are failing on appveyor despite passing on Travis.  I've got some tests running (well, queued to run, anyhow) on another branch.  If they work out I'll pull them into this branch.

Also, I did a little minor restructuring to silence a warning we were getting regarding the `exec` subdir.  I'm holding that change until I can make a decision on the windows fix, so as to cut down on redundant CI builds.

rplzzz:
Ok, I disabled the tests that called the python comparison tools under Windows.  @bpbond did we ever figure out why it was failing on your system?

bpbond:
No. Sorry for the delay. I should have time to look at this tomorrow.

rplzzz:
Fun fact:  if you put `testthat` functions like `expect` or `skip` in your package code, your tests will all pass, but the package functions will fail in live use.  

I'll have that fixed in a jiffy.

bpbond:
Note although tests don't _fail_ on my machine, some funky output is produced:
```
xml: .SSSSSS...SS......S......S.
XML diff: ............................At: [('scenario', ''), ('modeltime', '')]
	Left:
		final-calibration-year
		
		[]
		[]
		2010
		
	Right:
		final-calibration-year
		
		[]
		[]
		2015
		

At: [('scenario', ''), ('modeltime', '')]
	Left:
		inter-year
		
		['time-step']
		['5']
		2005
		
	Right:
		inter-year
		
		['time-step']
		['1']
		2005
		

..
```
If it's possible to clean this up, please do. Thanks.

rplzzz:
The console output in the standard mode of `cmp_xml_files` has been removed.  Standard mode now just returns a logical value.  If you want the diagnostic output from the comparison, you now have to use raw mode and examine the character vector returned from it.

PR #701, opened 2017-09-07
Kd_LA_142_building

kdorheim:
#700


kdorheim:
@bpbond  I think so but this is my first time doing the temp-data-inject stitching. I removed the only output from my chunk, `L142.in_EJ_state_bld_F.csv` from `inst/extdata/temp-data-inject`. Are there any other files I would need to delete?

bpbond:
No, that should be it. Thank you!

bpbond:
Waiting on @gokuliyer review.

PR #702, opened 2017-09-07
Snyder l223 land input 3

abigailsnyder:
actually for l223 land input 3 irr.

haven't done the corresponding xml chunk yet.

also, has errors on the function `fill_decay_extrapolate` which I neither used nor touched.


R/constants.R
```diff
@@ -56,6 +56,7 @@ PREAGLU_YEARS <- c(1700, 1750,1800, 1850, 1900, 1950)
 aglu.LAND_COVER_YEARS <- sort(unique(c(LAND_HISTORY_YEARS, AGLU_HISTORICAL_YEARS)))
 GTAP_HISTORICAL_YEAR <- 2000
```
kvcalvin: This isn't for this PR, but should we add `aglu` to the other constants too?

R/zchunk_L223.land_input_3.R
```diff
@@ -29,8 +29,9 @@ module_aglu_L223.land_input_3_DISABLED <- function(command, ...) {
              "L121.CarbonContent_kgm2_R_LT_GLU",
```
kvcalvin: I thought we didn't want this file?

abigailsnyder: I think this one is fine (and has been used in a ton of other chunks I've done...); I think it's L201 that we don't want? 
Depends on what context you're thinking of for not wanting I guess

R/zchunk_L2231.land_input_3_irr.R
```diff
@@ -0,0 +1,356 @@
...
+      select(one_of(LEVEL2_DATA_NAMES[["LN3_HistUnmgdAllocation"]])) ->
+      L223.LN3_HistUnmgdAllocation
+
+    L223.LC_bm2_R_Unmgd3_Yh_GLU %>%
+      filter(year %in% BASE_YEARS) %>%
+      select(one_of(LEVEL2_DATA_NAMES[["LN3_UnmgdAllocation"]])) ->
+      L223.LN3_UnmgdAllocation
+
+
+    # Managed land - non-crop (forest)
```
kvcalvin: Currently, this is just `non-crop (forest)`, but really this code would do anything we make a `LandLeaf` at level 3. Maybe change the comment to reflect?

abigailsnyder:
@bpbond what's the deal with codecov at this point? is everything still failing or is my failure an actual problem?

R/zchunk_L223.land_input_3.R
kvcalvin: I was thinking we don't want `L223.land_input_3.R`. It doesn't look like the code is producing meaningful tables...just adds precursors to a `tibble`

R/zchunk_L2231.land_input_3_irr.R
```diff
@@ -0,0 +1,356 @@
...
+      filter(grepl("Forest", UnmanagedLandLeaf)) %>%
+      select(region, LandAllocatorRoot, LandNode1, LandNode2, LandNode3, UnmanagedLandLeaf) %>%
+      rename(LandLeaf = UnmanagedLandLeaf) ->
+      L223.LN3_NoEmissCarbon
+
+    L223.LN3_MgdCarbon_noncrop %>%
+      select(region, LandAllocatorRoot, LandNode1, LandNode2, LandNode3, LandLeaf) %>%
+      bind_rows(L223.LN3_NoEmissCarbon, .) %>%
+      mutate(no.emiss.carbon.calc = " ",
+             extra = "junk") ->
```
kvcalvin: ?

R/zchunk_L223.land_input_3.R
abigailsnyder: oh duh I was looking at the wrong chunk. I started out on this one and then switched to the irr and added DISABLED back to this; so it doesn't get run but you're right, I should just delete the chunk entirely

R/zchunk_L2231.land_input_3_irr.R
abigailsnyder: No clue, I kept that from the original code; it seems so stupid that I assumed it had to be important for running gcam. At the very least, it has to stay in an `OLD_DATA_SYSTEM_BEHAVIOR` flag to pass tests.

abigailsnyder: this is the comment from the old code, which I can (and should have) bring over:
```
# Just need to create an empty tag, to do so we have to have a whitespace column and an
# extra column with extraneous data
```

bpbond:
@abigailsnyder re codecov...

1) We enable codecov for `gcamdata` and it works great.
2) But test coverage means running everything, and it turns out, as the number of chunks increases, this takes a fair amount of time...and codecov starts timing out and generating spurious errors. Suck.
3) So we skip the timeshift test when doing code coverage checks. This is the one test that runs all chunks, all the way through. Rationale: it only really makes sense to check intensively the support utilities, not the chunk code, as we can't really 'test' it in most ways.
4) Now, however, codecov reports 17% coverage or whatever...and every additional PR lowers this number, as it brings yet more chunk code that isn't tested.

Kind of sucky. We really should exclude all chunk code for coverage consideration entirely, I guess, or find a way to keep codecov for timing out and dying when running the entire system.

R/module-helpers.R
```diff
@@ -301,7 +301,7 @@ add_carbon_info <- function( data, carbon_info_table, matchvars = c("region", "G
   }
 
   data %>%
-    left_join_error_no_match(carbon_info_table, by = matchvars) %>%
+    left_join(carbon_info_table, by = matchvars) %>%
```
bpbond: What's going on with this change?

abigailsnyder: For land input 1 and 2, everything in the carbon info table has a match.  In land input 3, there start to be some mismatches and NAs. So I had to relax the join here and overwrite the NA's on line 191 (also had to be done in the old DS). I think it was like 3 regions that just don't have one specific land type; Pakistan and I don't remember the other two. I think they were maybe South America

bpbond:

PR #708, opened 2017-09-08
Exclude chunks (module_ functions) from coverage testing

bpbond:
The discussion with @abigailsnyder https://github.com/JGCRI/gcamdata/pull/702#issuecomment-328101473 got me thinking - can’t we just exclude all chunk code from coverage testing?


bpbond:
There's nothing like excluding twenty-seven thousand+ lines of code to improve your coverage metrics.

PR #728, opened 2017-09-08
Jill-L224.heat

jhoring:
#597


NAMESPACE
```diff
@@ -29,8 +29,15 @@ export(module_aglu_L2252.land_input_5_irr_mgmt_DISABLED)
...
 export(module_energy_L222.en_transformation_DISABLED)
+<<<<<<< HEAD
+export(module_energy_L224.heat)
+export(module_energy_L225.hydrogen_DISABLED)
+export(module_energy_L226.en_distribution_DISABLED)
+export(module_energy_L252.transportation_DISABLED)
+=======
 export(module_energy_L224.heat_DISABLED)
 export(module_energy_L254.transportation_UCD_DISABLED)
+>>>>>>> master
```
bpbond: @jhoring Did you see this git conflict marker? There's no way this is building without an error. Please, when you merge in `master`, look for conflicts and fix them! I'm happy to go over how on Monday.

R/constants.R
```diff
@@ -247,6 +247,9 @@ ELECTRICITY_INPUT_FUELS<- c("biomass", "coal", "gas", "refined liquids")
 energy.CLIMATE_NORMAL_YEARS <- 1981:2000
 energy.RSRC_FUELS <- c("coal", "gas", "refined liquids")
 
+#Assumed base year heat price, used for calculating adjustment to non-energy costs of electricity technologies with secondary output of heat
```
bpbond: Space after `#`; please add units for both constants, and prefix their names with `energy.`

R/zchunk_L224.heat.R
```diff
@@ -1,23 +1,26 @@
...
+#' subsector logit exponents, subsector shareweight and interpolation, and stubtech info by writing assumption file
+#' information to all model periods and regions that have district heat. It creates global tech coef, costs, and shareweights
+#' by interpolating assumptions. From the level 1 heat data, this chunk computes stub tech calibrated inputs, secondary
+#' outputs from elec and modified costs.
 #' @importFrom assertthat assert_that
 #' @importFrom dplyr filter mutate select
 #' @importFrom tidyr gather spread
-#' @author YourInitials CurrentMonthName 2017
+#' @author JDH August 2017
 #' @export
```
bpbond: Remove `@export` line

R/zchunk_L224.heat.R
```diff
@@ -65,185 +74,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["Supplysector"]], "logit.type"), GCAM_region_names = GCAM_region_names) %>%
+      filter(region %in% heat_region$region) -> L224.Supplysector_heat
+
+    # Subsector logit exponents of district heat sectors
+    A24.subsector_logit %>%
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["SubsectorLogit"]], "logit.type"), GCAM_region_names = GCAM_region_names) %>%
+      filter(region %in% heat_region$region) -> L224.SubsectorLogit_heat
+
+    # L224.SubsectorShrwt_heat and L224.SubsectorShrwtFllt_heat: Subsector shareweights of district heat sectors
+    if( any( !is.na( A24.subsector_shrwt$year ) ) ){
```
bpbond: No GCAM-style parenthesis spacing please

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["Supplysector"]], "logit.type"), GCAM_region_names = GCAM_region_names) %>%
+      filter(region %in% heat_region$region) -> L224.Supplysector_heat
+
+    # Subsector logit exponents of district heat sectors
+    A24.subsector_logit %>%
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["SubsectorLogit"]], "logit.type"), GCAM_region_names = GCAM_region_names) %>%
+      filter(region %in% heat_region$region) -> L224.SubsectorLogit_heat
+
+    # L224.SubsectorShrwt_heat and L224.SubsectorShrwtFllt_heat: Subsector shareweights of district heat sectors
+    if(any(!is.na(A24.subsector_shrwt$year))){
```
kvcalvin: Does this mean if we only have shrwts for some rows we won't process them?

kvcalvin: Never mind. I now see what it is doing.

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+      mutate(secondary.output = round(value, energy.DIGITS_CALOUTPUT)) %>%
+      select(-value) -> L224.StubTechSecOut_elec
+
+    # Calculate cost adjustment, equal to the output of heat multiplied by the heat price (to minimize the distortion of including the secondary output)
+    L224.StubTechSecOut_elec %>%
+      select(LEVEL2_DATA_NAMES[["StubTechYr"]], secondary.output) %>%
+      mutate(minicam.non.energy.input = "heat plant") %>%
+      mutate(input.cost = round(secondary.output*energy.HEAT_PRICE, energy.DIGITS_COST))-> L224.StubTechCost_elec
+
+    # For gas-electric technologies whose efficiencies are below the default assumptions, this cost needs to be reduced
```
kvcalvin: Could you add a little more to this comment about how this adjustment is being calculated?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+    L1231.eff_R_elec_F_tech_Yh %>%
+      filter(year %in% MODEL_YEARS) %>%
+      rename(efficiency = value) %>%
+      left_join(GCAM_region_names, by = "GCAM_region_ID") -> L224.eff_R_elec_F_tech_Y
+
+    L224.eff_R_elec_F_tech_Y %>%
+      filter(region %in% heat_region$region) %>%
+      filter(fuel == "gas") %>%
+      filter(efficiency < DEFAULT_ELECTRIC_EFFICIENCY) %>%
+      mutate(cost_modifier = energy.GAS_PRICE * (1 / DEFAULT_ELECTRIC_EFFICIENCY - 1 / efficiency)) -> L224.eff_Rh_elec_gas_sc_Y
```
kvcalvin: Could we also rename the tibble (`L224.eff_Rh_elec_gas_sc_Y`) to make it clearer that this is the cost adjustment?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
-      # typical flags, but there are others--see `constants.R`
-      add_flags(FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR) ->
+      add_precursors("energy/A24.subsector_shrwt", "energy/A_regions", "common/GCAM_region_names") ->
       L224.SubsectorShrwt_heat
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    } else {
```
kvcalvin: Would this ever get executed? If so, can you put a comment in indicating when?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
       L224.GlobalTechShrwt_heat
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    L224.StubTechCalInput_heat %>%
+      add_title("Calibrated input to district heat") %>%
+      add_units("unitless") %>%
```
kvcalvin: I would guess this is in EJ/yr

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+      filter(year %in% MODEL_YEARS) %>%
+      select(sector.name = supplysector, subsector.name = subsector, technology, minicam.energy.input, year, coefficient = coef) -> L224.GlobalTechCoef_heat
+
+    # Costs of global technologies
+    # L224.GlobalTechCost_heat: Costs of global technologies for district heat
+    A24.globaltech_cost %>%
+      gather(year, input.cost, -supplysector, -subsector, -technology, -minicam.non.energy.input) %>%
+      mutate(year = as.integer(year)) -> A24.globaltech_cost
+
+    A24.globaltech_cost %>%
```
kvcalvin: Could we merge these two pipelines?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+      ungroup() %>%
+      filter(year %in% MODEL_YEARS) %>%
+      select(sector.name = supplysector, subsector.name = subsector, technology, minicam.non.energy.input, year, input.cost) -> L224.GlobalTechCost_heat
+
+    # Shareweights of global technologies
+    # L224.GlobalTechShrwt_heat: Shareweights of global technologies for district heat
+    A24.globaltech_shrwt %>%
+      gather(year, share.weight, -supplysector, -subsector, -technology) %>%
+      mutate(year = as.integer(year)) -> A24.globaltech_shrwt
+
```
kvcalvin: Merge pipelines?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+    # Calibration and region-specific data
+    L124.in_EJ_R_heat_F_Yh %>%
+      left_join(GCAM_region_names, by = "GCAM_region_ID") %>%
+      left_join(calibrated_techs %>%
+                  select(sector, fuel, supplysector, subsector, technology, minicam.energy.input) %>%
+                  distinct, by = c("sector", "fuel")) %>%
+      rename(stub.technology = technology) %>%
+      filter(year %in% MODEL_YEARS) %>%
+      filter(region %in% heat_region$region) -> L224.in_EJ_R_heat_F_Yh
+
```
kvcalvin: Merge pipelines?

R/zchunk_L224.heat.R
```diff
@@ -65,185 +73,341 @@ module_energy_L224.heat_DISABLED <- function(command, ...) {
...
+    # use the secondary output of heat from the power sector only in the model base years.
+    L124.heatoutratio_R_elec_F_tech_Yh %>%
+      filter(year %in% MODEL_YEARS) %>%
+      left_join(GCAM_region_names, by = "GCAM_region_ID") %>%
+      left_join(calibrated_techs %>%
+                  select(sector, fuel, supplysector, subsector, technology) %>%
+                  distinct, by = c("sector", "fuel", "technology")) %>%
+      mutate(stub.technology = technology) %>%
+      mutate(secondary.output.name = A24.sector$supplysector) -> L224.heatoutratio_R_elec_F_tech_Yh
+
```
kvcalvin: Merge pipelines?

R/zchunk_L224.heat.R
jhoring: @kvcalvin I'm adding a description of what calculation is being done, but I'm not sure why they chose to modify it in this way. Do you have any advice?

kvcalvin: I don't know why it is done either. We can ask Page if @bpbond thinks it is important.

kvcalvin: I was mostly interested in a description.

bpbond: I think if no one knows why it's being performed, yes, time to tag @pkyle .

pkyle: OK I'll do a comment and some further explanation. Comment:
The secondary output of heat from CHP in the electric sector can cause the price of the technologies to go very low or negative if the technology cost is not modified to reflect the additional costs of CHP systems (as compared with electricity-only systems). Low technology costs can cause unrealistically low electricity prices in the calibration year, distorting behavior in future years. In this method, costs related to heat production and distribution are backed out from exogenous heat prices and data-derived heat:power ratios.

Further commentary: for a GCAM electric sector technology to have non-zero heat output, this indicates that some portion of the electricity in the region (and fuel type) came from the "MAINCHP" flow: facilities whose primary purpose is to produce heat and power for sale to a third party, as opposed to on-site use. In many cases, these "main activity CHP plants" produce 90% heat and 10% electricity; in other words, heat is the primary saleable commodity, and electricity is secondary. It's a bit odd to have these technologies assigned to the electric power sector to begin with, but that's the nature of having mappings that aren't dependent on the magnitude of the flow volumes (which is simpler than the alternative).

There are several possible methods of dealing with this cost/price issue. One is to read in a p-multiplier of zero on the secondary output, and leave all costs as they are. That's the simplest, but it's not an accurate depiction of what's actually happening; in a lot of the cases, heat is the primary output and therefore primary revenue stream. Another alternative is that this method--assigning heat-related costs in order to keep power generation costs consistent with electric-only technologies--could be implemented more precisely, by calculating heat prices in each region and time period from exogenous global fossil energy prices (gas, oil, coal), cost adders, regional T&D losses, and the share of each fuel. Doing this would require a whole lot more data being read in and processed by this code chunk, and given that biomass prices in the calibration years float, the derived heat prices still wouldn't be guaranteed equal to the prices observed in a model run. The method implemented is a middle ground between zeroing out all heat-related revenues, and performing an ad-hoc estimation of the actual heat prices in each region.

R/zchunk_L224.heat.R
jhoring: @kvcalvin the first one puts A24.globaltech_cost into long format and then in the second pipeline it joins on this again (so needs to be long). I could just move the first pipeline to the read data section, which I think has been a convention with 'gather'ing input files? Same story for the ones below.

jhoring: Actually the one* below. Other two can be merged!

bpbond:

bpbond:
Hmm, @jhoring this is throwing an error at line 183:
```
 Error: All select() inputs must resolve to integer column positions.
The following do not:
*  LEVEL2_DATA_NAMES[["StubTechYr"]] 
```

jhoring:
@bpbond strange! I'll look into that now

bpbond:
Jill, did this _ever_ pass tests on your machine? I don't see how. The code is producing a 40-row `L224.StubTech_heat`, for example, while the old data system produces a 20-row data frame.

The use of `select` + LEVEL2_DATA_NAMES is wrong and throws an error.

There was an error in one of the column names for L2231, and the documentation for LA161 was accidentally duplicated; these were not your mistakes, but mean that (again) there's no way this was passing tests.

If you open a PR that doesn't pass tests, or you merge in `master` on an existing PR and things fail, **say so and tell us what's happening**. The corollary to this is: **test and check the package**.

Issue #719, opened 2017-09-08
L225.hydrogen_USA


gokuliyer:

kdorheim:
@bpbond when you have time can you add the files for this chunk?

PR #738, opened 2017-09-15
Kd_l225_hydrogen_usa

kdorheim:
#719


kdorheim:
@bpbond thanks, the `Interesting way to strip off the data frame attributes.` was @russellhz idea

PR #742, opened 2017-09-18
Sean---LA115.RooftopPV

swd-turner:
#699


swd-turner:
Got a problem with this chunk. I struggled to get all the rows of the output to match exactly. This is related to some arbitrary ordering in the legacy code (thrown by the `merge` function) that I can't easily reproduce. The output is basically supply curve parameters that are fitted against the generation for each US State. This is done by ordering by price, calculating the cumulative sum, then using `optimize` to fit. The issue is that there are often two or three data points with the same price, and the way they get ordered affects the cumulative sums and ultimately the fitted parameters (only slightly). There's no apparent logic behind the ordering thrown by `merge` (it seems to vary depending on State), so I can't `arrange` my data in any way to replicate the results in a way that works for all US States.
Not sure how to proceed but here are two options:
-	I manually re-order the (eleven) problematic States within the code (will look ugly).
-	Accept the code as is and flag as don’t test (current approach). The differences are v.small (< 0.1), but not small enough to pass the test.

Also, not sure if the owner of this part of GCAM will want to review the approach taken here. I’d suggest condensing the data points with the same prices (e.g., Alabama) into a single point so that the supply curves aren’t sensitive to this arbitrary ordering choice.

Let me know if the above doesn't make sense!

bpbond:
Hey @swd-turner thanks. FYI row order does _not_ matter is the oldnew test, so don't sweat that.

swd-turner:
@bpbond it's not to do with ordering of final data, but ordering within the code, which affects the output of a cumulative sum calculation.

R/constants.R
```diff
@@ -404,6 +404,19 @@ emissions.AGR_GASES            <- c("CH4_AGR", "N2O_AGR", "NH3_AGR", "NOx_AGR")
...
+# PV constants ======================================================================
+
+PV.DERATING_FACTOR <- 0.77  # incorporates various factors: inverters, transformers, mismatch, soiling, and others
+PV.RESID_INSTALLED_COST <- 9500  # 2005USD per kw
+PV.COMM_INSTALLED_COST <- 7290  # 2005USD per kw
+PV.RESID_OM <- 100  # 2005USD per kw per year
+PV.COMM_OM <- 40  # 2005USD per kw per year
+PV.LIFETIME <- 30  # years
+PV.DISCOUNT_RATE <- 0.1  # year^-1
+HOURS_PER_YEAR <- 8760
```
bpbond: 24 * 365 would be clearer?

R/constants.R
```diff
@@ -404,6 +404,19 @@ emissions.AGR_GASES            <- c("CH4_AGR", "N2O_AGR", "NH3_AGR", "NOx_AGR")
 emissions.AG_MACC_GHG_NAMES    <- c("CH4_AGR", "N2O_AGR")
 emissions.GHG_NAMES            <- c("CH4", "N2O")
 emissions.USE_GV_MAC           <- 1
+
+# PV constants ======================================================================
+
+PV.DERATING_FACTOR <- 0.77  # incorporates various factors: inverters, transformers, mismatch, soiling, and others
```
bpbond: Can you change these all to `PV_*`?

R/zchunk_LA115.RooftopPV.R
```diff
@@ -30,43 +28,125 @@ module_gcam.usa_LA115.RooftopPV_DISABLED <- function(command, ...) {
...
-    # Note that all precursor names (in `add_precursor`) must be in this chunk's inputs
-    # There's also a `same_precursors_as(x)` you can use
-    # If no precursors (very rare) don't call `add_precursor` at all
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    # Compute PV fixed charge rate
+    PV_FCR <- ( PV.DISCOUNT_RATE * ( 1 + PV.DISCOUNT_RATE ) ^ PV.LIFETIME ) /
```
bpbond: No GCAM-style parenthesis spacing please

R/zchunk_LA115.RooftopPV.R
```diff
@@ -30,43 +28,125 @@ module_gcam.usa_LA115.RooftopPV_DISABLED <- function(command, ...) {
...
+      ( ( PV.DISCOUNT_RATE + 1 ) ^ PV.LIFETIME - 1 )
+
+    # Get maximum residential capacity factor
+    filter(NREL_Res_PV_supply_curve, Relative_Cost == min(Relative_Cost))$MWh /
+      filter(NREL_Res_PV_supply_curve, Relative_Cost == min(Relative_Cost))$MW /
+      HOURS_PER_YEAR -> max_resid_capacity_factor
```
bpbond: This is kind of confusing. What's coming out of this–a vector or a scalar?

R/zchunk_LA115.RooftopPV.R
```diff
@@ -30,43 +28,125 @@ module_gcam.usa_LA115.RooftopPV_DISABLED <- function(command, ...) {
...
+
+    # Combine supply curves, add state abbrv., and rebase p
+    bind_rows(L115.pv_sc_res, L115.pv_sc_com) %>%
+      left_join_error_no_match(select(states_subregions, state_name, state),
+                               by = "state_name") %>%
+      arrange(state_name) %>%
+      select(-state_name, state, p, generation) %>%
+      mutate(p = p - min(p)) %>%
+      arrange(p) ->
+      # ^^ note: ordering of generation for matching p values will affect supply curve parameters (fitted below).
```
bpbond: Ugh

R/zchunk_LA115.RooftopPV.R
```diff
@@ -30,43 +28,125 @@ module_gcam.usa_LA115.RooftopPV_DISABLED <- function(command, ...) {
...
+      L115.pv_sc_State
+
+    # Get mid-price for each state
+    L115.pv_sc_State %>%
+      filter(percent_cumul > 0.5) %>%
+      # mid-price taken as the first price after the 50th percentile
+      summarise(mid_p = first(p)) -> L115.pv_midPrice
+
+    # Create smoothing functions
+    smooth_res_curve_approx <- function( b, midp, p ) {
```
bpbond: Are these used anywhere else? I'm wondering if their definitions belong in `pipeline_helpers.R` or `module_helpers.R`

R/zchunk_LA115.RooftopPV.R
```diff
@@ -30,43 +28,125 @@ module_gcam.usa_LA115.RooftopPV_DISABLED <- function(command, ...) {
...
+    # ... from the actual supply curve, then returns the error between...
+    # ... computed and actual values.
+    get_error_min_b <- function(stateAb) {
+      midPrice <- filter(L115.pv_midPrice, state == stateAb)$mid_p
+      supplyCurve <- filter(ungroup(L115.pv_sc_State), state == stateAb)
+      optimize(f = smooth_res_curve_approx_error, interval = c(1.0, 15.0),
+               midPrice, supplyCurve) %>% unlist
+    }
+    # Get error_min_b for all states using apply (replaces "for loop" structure of legacy code)
+    sapply(unique(L115.pv_sc$state), get_error_min_b) %>% t %>% as.data.frame %>%
```
bpbond: `as_tibble`?

R/zchunk_LA115.RooftopPV.R
swd-turner: Ok I'll catch up with you at some point to explain this issue (you around today?)

bpbond: Sounds good, thanks. I'm out today, unexpectedly, but in the rest of the week.

R/constants.R
bpbond: Also note if we want to account for leap years, 24 * 365.25 = 8766!

R/zchunk_LA115.RooftopPV.R
gokuliyer: I second Ben's question.

R/zchunk_LA115.RooftopPV.R
gokuliyer: I think it will be useful to mention the equation for the supply curve somewhere in the script (preferably upfront) and describe what parameters of that equation are being estimated here.

inst/extdata/gcam-usa/NREL_Com_PV_supply_curve.csv
```diff
@@ -0,0 +1,566 @@
+# File: NREL_Com_PV_supply_curve.csv
+# Title: PV supply curve for commercial buildings
+# Units: Rel_Cost - Unitless
+# Source: www.nrel.gov
```
gokuliyer: Please add a better source than just nrel.gov. The source is:

Denholm, P. & Margolis, R. Supply Curves for Rooftop Solar PV-Generated Electricity for the United States. National Renewable Energy Laboratory, Technical Report NREL / TP-6A0-44073 (2008)

If possible, please track the specific table in the report down.

inst/extdata/gcam-usa/NREL_Res_PV_supply_curve.csv
```diff
@@ -0,0 +1,2040 @@
+# File: NREL_Res_PV_supply_curve.csv
+# Title: PV supply curve for residential buildings
+# Units: Relative_Cost - Unitless
+# Source: www.nrel.gov
```
gokuliyer: Same as above. Add appropriate citation.

R/zchunk_LA115.RooftopPV.R
swd-turner: It's just a constant. I'll clarify in the comments.

R/zchunk_LA115.RooftopPV.R
swd-turner: @bpbond you around today or tomorrow to quickly discuss this??

R/zchunk_LA115.RooftopPV.R
swd-turner: Don't seem to be elsewhere (not sure though). I'd be for keeping them in here so that anyone reading the script can easily see the function being fitted.

R/zchunk_LA115.RooftopPV.R
swd-turner: the `as.data.frame` is useful here for retaining row names that come out of the sapply function (which later get tranferred into tibble columns).

R/zchunk_LA115.RooftopPV.R
bpbond: Tomorrow, sure, sounds good.

swd-turner:
old_new_compare
b_exp parameter in output is slightly different from OLD system, due to ordering issue discussed above. Graph shows that these differences are almost negligible (although some manual adjustment has been added to pass oldnew test).

bpbond:
Looks good; holding on @gokuliyer approval.

Issue #711, opened 2017-09-08
LA132.Industry


gokuliyer:

swd-turner:
@bpbond can you generate the zchunk for this one? There's already a LA132.industry, but I think it's for the energy section rather than gcam_usa.

PR #52, opened 2017-03-17
Update README.md

bpbond:

PR #744, opened 2017-09-18
Sean---LA132.industry_GCAM_USA

swd-turner:
#711


R/zchunk_LA132.Industry-gcamusa.R
```diff
@@ -40,73 +43,165 @@ module_gcam.usa_LA132.Industry_DISABLED <- function(command, ...) {
...
+      # ^^ remove fuels that are inputs to cogen systems, i.e., not electricity
+      mutate(sector = "chp_elec") %>% rename(multiplier = value) %>%
+      # ^^ prepare for smooth join
+      left_join_error_no_match(filter(L123.in_EJ_R_indchp_F_Yh, GCAM_region_ID == gcam.USA_CODE),
+                               by = c("fuel", "year", "sector")) %>%
+      mutate(value = value * multiplier) %>%
+      select(-multiplier, -GCAM_region_ID) ->
+      L132.in_EJ_state_indchp_F
+
+    #Apportion nation-level industrial cogen output to states
```
bpbond: Just because I need to say _something_: `# Apportion`

R/zchunk_LB126.Gas_ElecTD.R
```diff
@@ -308,9 +308,9 @@ module_gcam.usa_LB126.Gas_ElecTD <- function(command, ...) {
       add_legacy_name("L126.out_EJ_state_td_elec") %>%
       add_precursors("L122.in_EJ_state_refining_F",
                      "L123.out_EJ_state_elec_F",
-                     "temp-data-inject/L132.in_EJ_state_indchp_F",
-                     "temp-data-inject/L132.in_EJ_state_indfeed_F",
-                     "temp-data-inject/L132.in_EJ_state_indnochp_F",
```
bpbond: Thanks. Did you delete these files out of `inst/extdata/temp-data-inject/`?

bpbond:
I'm getting all kinds of oldnew errors on this PR.

swd-turner:
hmm weird... it tested clean on my system... will investigate tomorrow.

swd-turner:
Don't think this is a dplyr version issue (I don't use `select("x","y")`). I just ran tests again on my machine and I am getting the oldnew fails. Guess I must have meddled with something after testing. On it now.

R/zchunk_LB126.Gas_ElecTD.R
swd-turner: Done now. I'd also left parts of the temp-inject code in the chunk, which was causing most of the oldnew fails.

swd-turner:
@bpbond FYI had to solve a weird test fail here for output 'L132.in_EJ_state_indeed_F.csv'. Turns out that if the first 1000 entries to the value column zero, the `read_csv` call in the oldnew test (#52) will fail. The reason is that the function looks at the first 1000 entries to determine whether it's integer or double... so there will be parsing fails if it assumes integer and then doubles appear (https://github.com/tidyverse/readr/issues/645). The problem was solved by re-arranging the tibble to bring some non-zero values to the top.

swd-turner:
ok I now see that this is a recognized issue... should have checked the common errors first. Doh!

bpbond:
>ok I now see that this is a recognized issue... should have checked the common errors first. Doh!

haha! Well, kudos for figuring it out.

FYI another solution is to use FLAG_PROTECT_FLOAT. Yeah, it's a crappy fail and not our fault.

bpbond:
Waiting on @gokuliyer review.

Issue #749, opened 2017-09-19
Fix dstrace


bpbond:
See #628

bpbond:
Started on this. Strategy is to have a generator script that, like for LEVEL2_DATA_NAMES, will generate a GCAM_DATA_MAP object that is saved as package internal data. The `dstrace` function can then query this object for info.

To do: When asked to generate a data map, the driver needs to return (in addition to precursor information) the comments, flags, title, and units associated with each data object. Probably want two separate tables?

https://github.com/JGCRI/gcamdata/tree/dstrace-fix

bpbond:
Implemented in #751

Issue #628, opened 2017-08-10
Change when precursor information gets returned?


bpbond:
Currently the driver can query data system chunks for their inputs and outputs, and from the returned information, build [pretty dependency graphs](https://github.com/JGCRI/gcamdata/blob/master/figures/network_current.png) and figure out general dependency information--i.e., the chunk call sequence.

This dependency information is however chunk-level, i.e. coarse. To get _object_ dependencies, we have to wait until chunks run and return this information (via the `add_precursor` mechanism). Then `dstrace` can then run cool object-level traces...but only after we run the entire data system and save _all_ its output.

It seems more sensible to have the precursor information returned as part of the driver's DECLARE_OUTPUTS message to chunks. This would be instantaneous and obviate the need to run the data system to trace data objects. For example, instead of

```R
  } else if(command == driver.DECLARE_OUTPUTS) {
    return(c("L110.For_ALL_bm3_R_Y"))
  } else if(command == driver.MAKE) {
 ...
      add_precursors("common/iso_GCAM_regID",
                     "L100.FAO_For_Prod_m3",
                     "L100.FAO_For_Imp_m3",
                     "L100.FAO_For_Exp_m3") %>%
```

the code would be
```R
  } else if(command == driver.DECLARE_OUTPUTS) {
    return(list("L110.For_ALL_bm3_R_Y" = c("common/iso_GCAM_regID",
                     "L100.FAO_For_Prod_m3",
                     "L100.FAO_For_Imp_m3",
                     "L100.FAO_For_Exp_m3")))
  } else if(command == driver.MAKE) {
```

This would involve changing the driver and all the already-written chunks, obviously, but going forward wouldn't be any extra work for people. The `same_precursors_as` mechanism might be implemented by writing

```R
  } else if(command == driver.DECLARE_OUTPUTS) {
    return(list("X" = c("1", "2", "3"),
                "Y" = "X")))
  } else if(command == driver.MAKE) {
```
Thoughts?  @pralitp @rplzzz

pralitp:
If only we thought of that before.. although it would be pretty easy to write a script to do it.  Maybe we wait until version one and then just convert them all then?

Also I kind of like keeping the `add_precursors` and `same_precursors_as` as basically tags.  It doesn't add anything except in my opinion more make more clear what is going on:

```
  } else if(command == driver.DECLARE_OUTPUTS) {
    return(list("X" %>% add_precursors("1", "2", "3"),
[...]

PR #751, opened 2017-09-20
dstrace fix

bpbond:
This PR reworks the `dstrace` function, addressing #749 as well as #628.

Originally, `dstrace` used the complete data returned by the driver to trace some object back through the maze of dependency data returned by the chunks after they run. However, because of the large size of this store, we completely changed the driver behavior in #530 .

Now `dstrace` depends on a metadata table that is internal to the `gcamdata` package. This metadata table is generated by the driver, when its new `return_data_map_only` parameter is TRUE, and following the LEVEL2_DATA_NAMES approach is saved by `data-raw/generate_data_map.R`. The table has one row for every input and output in the data system, and includes generating chunk (or "INPUT" for file inputs), title, units, comments, flags, and precursor information. It can be viewed/accessed as `gcamdata:::GCAM_DATA_MAP`.

The advantage of this approach is that you can do fullly-detailed traces _without_ having run the data system first. The disadvantage is that the metadata table needs to be updated manually when the precursor structure of the system changes.

Simple example:
```R
> dstrace("L100.FAO_ag_Exp_t")
1 - L100.FAO_ag_Exp_t - produced by module_aglu_LA100.FAO_downscale_ctry 
	FAO agricultural exports by country, item, year (t)
	Downscale countries; calculate 5-yr averages
	Precursor: aglu/FAO/FAO_ag_Exp_t_SUA (#2 below)
	Precursor: aglu/AGLU_ctry (#3 below)
2 - aglu/FAO/FAO_ag_Exp_t_SUA - read from file
	FAO agricultural exports by country, item, year (tons)
	Read from extdata/aglu/FAO/FAO_ag_Exp_t_SUA.csv.zip
	No precursors
3 - aglu/AGLU_ctry - read from file
	Mapping of countries in AGLU databases to ISO codes (NA)
	Read from extdata/aglu/AGLU_ctry.csv
	No precursors
```

Long, complicated example: `dstrace("demand_input.xml")`

It would be cool to have `dstrace` plot the dependency tree...


R/dstrace.R
```diff
@@ -1,17 +1,20 @@
...
 #' dstrace
 #'
-#' @param object_name Name of object to trace
-#' @param all_data Data structure produced by a call to \code{\link{driver}}
+#' @param object_name Name of object to trace (can be either a data object or a code chunk)
+#' @param gcam_data_map A tibble of metadata information; normally a built-in package dataset
 #' @param previous_tracelist Information about previous objects printed
 #' @param recurse Recurse to print information about precursor objects?
-#' @return Boolean (object_name found, or not)
+#' @return A tibble with the trace information (object name and trace number)
```
abigailsnyder: I wonder if there's any kind of option to return a plot as well?  
If I'm trying to figure out a problem, especially with something that results in as many objects as `dstrace("demand_input.xml")`, being able to effortlessly plot a directed graph might be really helpful. Maybe default to not doing it, but have an option like `dstrace("demand_input.xml", plot_directed_graph = TRUE)`

bpbond:
Added more test code, and the ability to trace downstream as well as up ( #752 ):
```R
> dstrace("L100.FAO_ag_Exp_t", downstream = TRUE)
1 - L100.FAO_ag_Exp_t - produced by module_aglu_LA100.FAO_downscale_ctry 
	FAO agricultural exports by country, item, year (t)
	Downscale countries; calculate 5-yr averages
	Dependent: L106.ag_NetExp_Mt_R_C_Y (#2 below)
2 - L106.ag_NetExp_Mt_R_C_Y - produced by module_aglu_LA106.ag_an_NetExp_FAO_R_C_Y 
	Net exports of primary agricultural goods by GCAM region / commodity / year (Mt)
	Aggregate FAO primary agricultural goods gross exports and imports and calculate net exports by GCAM region, commodity and
	year; Gross exports are adjusted so that global net exports add to zero; Re-calculate regional net exports using adjusted
	gross exports minus gross imports
	Dependent: L109.ag_ALL_Mt_R_C_Y (#3 below)
3 - L109.ag_ALL_Mt_R_C_Y - produced by module_aglu_LB109.ag_an_ALL_R_C_Y 
	Primary agricultural good mass balances, by region / commodity / year. (Mt)
	Calculate primary agricultural good mass balances by GCAM region, commodity and year; Adjusts global and regional net
	exports to remove net negative other uses
	Dependent: L203.StubTechProd_nonfood_crop (#4 below)
	Dependent: L203.BaseService (#5 below)
4 - L203.StubTechProd_nonfood_crop - produced by module_aglu_L203.demand_input 
	Crop non-food demand by technology and region (Mt)
	Map in crop non-food demand in calibration years by region / commodity; Remove any regions for which agriculture and land
	use are not modeled
	Dependent: demand_input.xml (#6 below)
6 - demand_input.xml - XML data
	 ()
	
	No dependents
5 - L203.BaseService - produced by module_aglu_L203.demand_input 
	Base service of final demands (Pcal/Mt/bm3)
	Calculate the total final demands by supply sector in each region; Remove any regions for which agriculture and land use
	are not modeled
	Dependent: demand_input.xml (#6 above)
```

R/dstrace.R
bpbond: That would definitely be desirable, agreed. Particularly if the graph were labeled with the corresponding numbers & names that appear in the `dstrace` output!

Opening new issue for this.

bpbond:
Added a basic graphing capability per @abigailsnyder suggestion:

us


ds

kvcalvin:
Ooh, I like it. I'll review the code in a little bit

R/dstrace.R
```diff
@@ -71,10 +105,64 @@ dstrace <- function(object_name, all_data, previous_tracelist = NULL, recurse =
...
+#' @return Adjacency matrix, invisible
+dstrace_plot <- function(object_name, tracelist, downstream, ...) {
+
+  # Make an adjacency matrix, laboriously
+  mat <- matrix(0, nrow = nrow(tracelist), ncol = nrow(tracelist))
+  colnames(mat) <- paste0(tracelist$tracenum, ". ", tracelist$object_name)
+  for(i in seq_len(nrow(tracelist))) {
+    if(!is.na(tracelist$related[i])) {
+      for(j in strsplit(tracelist$related[i], driver.SEPARATOR, fixed = TRUE)[[1]]) {
+        #        cat(tracelist$object_name[i], "->", j, "\n")
```
kvcalvin: Delete?

bpbond:
Any other feedback @kvcalvin @abigailsnyder @rplzzz ? I'd like to get this merged today if possible; but if you tell me better to hold, no worries. Thanks.

kvcalvin:
I am going to test it in a few minutes and then will let you know if I have additional feedback.

R/dstrace.R
```diff
@@ -1,80 +1,197 @@
...
+#' @details What other data products feed into some particular data system object?
+#' Conversely, to what other products does some object feed? These are the kinds
+#' of questions that \code{dstrace} can help answer.
 #' @importFrom assertthat assert_that
+#' @author BBL
 #' @export
-dstrace <- function(object_name, all_data, previous_tracelist = NULL, recurse = TRUE) {
+#' @examples
+#' dstrace("L100.FAO_ag_Exp_t")
+#' dstrace("L100.FAO_ag_Exp_t", downstream = TRUE)
```
kvcalvin: should this example now be `direction = "downstream"`

bpbond: Good point, thanks. I actually don't understand why R CMD CHECK didn't catch this and error.

rplzzz:
@bpbond I just started looking.  Don't let me hold you up if I'm the laggard on this.

R/driver.R
```diff
@@ -179,6 +213,13 @@ driver <- function(all_data = empty_data(),
     all_data <- add_data(csv_data, all_data)
   }
 
+  # Extract metadata from the input data; we'll add output metadata as we run
+  metadata_info <- list()
+  if(return_data_map_only) {
+    if(!quiet) cat("Extracting metadata from inputs...\n")
```
rplzzz: Stuff like this should be done with `message`.  Then you can dispense with the `quiet` argument and just let the user use `suppressMessages` if they don't want to see the output (warning and error text will still come through even with messages suppressed.

bpbond:
@rplzzz thanks–good point! But I think we can do even better.

On Travis, we have to call the `driver` for a full run anyway, because we need to generate the chunk outputs for oldnew testing. We can do so with its new `return_data_map_only` parameter set which means, in addition to writing all the chunk outputs to disk, it will return to the testing code the _actual correct data map_. This can then be compared to `gcamdata:::GCAM_DATA_MAP` and an error thrown if they don't match. Score!

R/driver.R
bpbond: Solid point but since this is the current standard throughout, I'm going to leave it for now. Happy to revisit for v1.1.

rplzzz:
I love it when a plan comes together!

bpbond:
Re my comment above, it turns out that simply comparing what the driver returns with `gcamdata:::GCAM_DATA_MAP` works locally, but doesn't work on Travis. I don't know why. These tests work:
```R
    expect_true(tibble::is_tibble(gcamdata:::GCAM_DATA_MAP))
    expect_true(tibble::is_tibble(gcam_data_map))
    expect_identical(dim(gcamdata:::GCAM_DATA_MAP), dim(gcam_data_map))
    expect_identical(gcamdata:::GCAM_DATA_MAP$name, gcam_data_map$name)
    expect_identical(gcamdata:::GCAM_DATA_MAP$output, gcam_data_map$output)
    expect_identical(gcamdata:::GCAM_DATA_MAP$precursors, gcam_data_map$precursors)
```
But these tests all fail on Travis:
```R
    expect_identical(gcamdata:::GCAM_DATA_MAP$title, gcam_data_map$title)
    expect_identical(gcamdata:::GCAM_DATA_MAP$units, gcam_data_map$units)
    expect_identical(gcamdata:::GCAM_DATA_MAP$comments, gcam_data_map$comments)
    expect_identical(gcamdata:::GCAM_DATA_MAP$flags, gcam_data_map$flags)
    expect_equivalent(gcam_data_map, gcamdata:::GCAM_DATA_MAP,
                       info = "GCAM_DATA_MAP is out of date; rerun data-raw/generate-package-data.R")
```
So not ideal. Still, we can guarantee that the pre-packaged GCAM_DATA_MAP has the same dimensions, object names, output names, and precursors as what's being returned by the driver. That's probably good enough.

PR #754, opened 2017-09-21
Russell-LA144.Commercial

russellhz:
#698


R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+    # a) PREPARATION AND CLEANING OF CBECS DATA (Commercial Buildings Energy Consumption Survey)
+    # The 1979 and 1983 only have floorspace by census region
+    # We can't bind_rows because all CBECS have different columns
+    L144.CBECS_all <- list(CBECS_1986, CBECS_1989, CBECS_1992, CBECS_1995, CBECS_1999, CBECS_2003)
+    names(L144.CBECS_all) <- paste0("CBECS", c(1986, 1989, 1992, 1995, 1999, 2003))
+
+    # In order to be able to work with these data across years, the "edition" number needs to be removed from all
+    # variable names. E.g., re-naming square footage from "SQFT3" in 1986 and "SQFT4" in 1989 to "SQFT" in all.
+    L144.CBECS_all <- lapply(L144.CBECS_all, function(df) {
+      stats::setNames(df, sub("[0-9]{1}", "", names(df)))
```
bpbond: If always at the end, `"[0-9]{1}$"` safer?

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+    L144.CBECS_1979_1983 <- CBECS_1979_1983 %>%
+      left_join_error_no_match(states_subregions_sub4, by = "REGION") %>%
+      # Using left_join because there will be two values for the two years for each subregion
+      left_join(L144.pop_sR4 %>%
+                                 filter(year %in% c(1979, 1983)), by = "subregion4") %>%
+      # SQFT1 represents 1979 square footage, SQFT2 represents 1983 square footage
+      mutate(pcflsp_m2 = if_else(year == 1979, SQFT1 / value * CONV_MILFT2_M2, SQFT2 / value * CONV_MILFT2_M2))
+
+    # Add in a year column to CBECS data so that we can bind rows later
+    for (i in 1:length(L144.CBECS_all)){
```
bpbond: `for(i in seq_along(L144.CBECS_all)) {`

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+    L144.flsp_bm2_sR4 <- L144.CBECS_all %>%
+      # For each tibble, select sqft, weights for summing, year, and subregions
+      lapply(function(df){
+        df %>%
+          select(SQFT, ADJWT, subregion4, subregion9, year)
+      }) %>%
+      # Bind all CBECS tibbles
+      do.call(bind_rows, .)
+
+    # For 1986, calculate per capita floorspace by subregion4
```
bpbond: Why? What's special about 1986?

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+        if (length(cols_to_keep) > 0) {
+          df %>%
+            select(cols_to_keep, ADJWT, subregion9, year) %>%
+            # All cols_to_keep have BTU in name
+            gather(variable, value, contains("BTU")) %>%
+            group_by(subregion9, year, variable) %>%
+            # Sum by census division, multiplying by sampling weights
+            summarise(value = sum(value * ADJWT * CONV_KBTU_EJ)) %>%
+            ungroup()
+        }else{
```
bpbond: `} else {`

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+    L144.in_EJ_sR9_CBECS_Fdist_U_Y <- L144.in_EJ_sR9_CBECS_F_U_Y %>%
+      filter(fuel == "district services") %>%
+      select(-fuel) %>%
+      # Convert from total district service value to value for each fuel
+      repeat_add_columns(tibble(fuel = unique(EIA_distheat$fuel))) %>%
+      left_join_error_no_match(EIA_distheat, by = c("fuel", "service")) %>%
+      mutate(value = value * share / efficiency) %>%
+      select(-share, -efficiency)
+
+    # Merge this back into the initial table, but with the district services fuel removed
```
bpbond: Useful to explain why, exactly?

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+    # Assembling unscaled energy consumption by state, fuel, and service
+    L144.in_EJ_state_comm_F_U_Y_unscaled <- bind_rows(L144.in_EJ_state_comm_F_heating_Y,
+                                                      L144.in_EJ_state_comm_F_cooling_Y,
+                                                      L144.in_EJ_state_comm_F_Uoth_Y)
+
+    # Calculating shares of energy consumption by each service, within each state and fuel
+    L144.pct_state_comm_F_U_Y <- L144.in_EJ_state_comm_F_U_Y_unscaled %>%
+      group_by(state, fuel, year) %>%
+      mutate(value = value / sum(value),
+             sector = "comm") %>%
```
bpbond: If this grouped mutate takes any nontrivial amount of time, it's probably faster to do this `sector = "comm"` assignment after the `ungroup`, i.e. do it once for the entire dataset, not within each individual group. Same comment applies to line 391 below.

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+     L144.in_EJ_state_commint_elec <- L142.in_EJ_state_bld_F %>%
+       filter(sector == "comm",
+              fuel == "electricity") %>%
+       left_join_error_no_match(L144.in_EJ_state_commext_F_U_Y, by = c("state", "year", "sector", "fuel")) %>%
+       mutate(value = value.x - value.y) %>%
+       select(-service, -value.x, - value.y)
+
+     # Bind this back to the initial table of commercial energy use by state and fuel
+     L144.in_EJ_state_commint_F <- L142.in_EJ_state_bld_F %>%
+       filter(sector == "comm" & fuel != "electricity") %>%
```
bpbond: Just for consistency, `filter(sector == "comm", fuel != "electricity")`

R/zchunk_LA144.Commercial.R
```diff
@@ -54,56 +74,453 @@ module_gcam.usa_LA144.Commercial_DISABLED <- function(command, ...) {
...
+
+    # Remove heating and cooling
+    L144.in_EJ_sR9_comm_F_Uoth_Y <- L144.in_EJ_sR9_comm_F_U_Y %>%
+      filter(!(service %in% c("comm heating", "comm cooling")))
+
+    # Convert to state data by multiplying by floorspace share
+    L144.in_EJ_state_comm_F_Uoth_Y <- L144.in_EJ_sR9_comm_F_Uoth_Y %>%
+      select(fuel, service) %>%
+      distinct() %>%
+      repeat_add_columns(tidyr::crossing(state = gcamusa.STATES, year = HISTORICAL_YEARS)) %>%
```
bpbond: Have you gone back and put this (`gcamusa.STATES`) into previously-done gcam-usa chunks?

russellhz: Did this for chunks I had done-should I do a search through all gcam-usa chunks?

rplzzz:
The PR version of this test is failing because it's picking up the new tests for `GCAM_DATA_MAP` from the master branch.  (The CI build for the PR merges into the base branch before running the build, while the branch build builds the branch as is).  So, there are two things that need to happen here.

1. The `GCAM_DATA_MAP` needs to be rebuilt.
2. We need to add some more informative info clauses to the data map tests, since right now when they fail they make it look like the problem is that the old and new data for the chunk don't match.

R/zchunk_LA144.Commercial.R
bpbond: @russellhz No, we can have original authors do that. Thanks.

bpbond:
Good–that is, this was the first real live test of the GCAM_DATA_MAP test and glad it failed. 

Note that I tried adding an informative error message to the test, but it doesn't seem to print on Travis.

rplzzz:
It's also a rare illustration of the difference between the branch build and the PR build.  

Regarding error messages on Travis, for some reason our builds don't produce the failure logs, which have more information in them than the last 13 lines that appear in the console output.  We should take a closer look at the travis config for this repository.  If I remember correctly, we homebrewed it instead of using the version produced by devtools, and that might be the cause.

bpbond:
Agreed.

Issue #706, opened 2017-09-08
L202.Ccoef


bpbond:

jhoring:
@bpbond Could you generate the script for this when you get a chance? Thanks!

jhoring:
@bpbond Thanks!

PR #759, opened 2017-09-22
Jill-L202.Ccoef

jhoring:
#706


jhoring:
@bpbond This chunk has a few switches that are called in if statements that are in constants.R. The current switches are such that not much of this chunk gets run (default option is to use an old table). Is there a good way to test what happens when they are switched, other than doing this in both the old and new data system and rerunning the test and check packages? 

Also, there is a constant called emissions.INVENTORY_MATCH_YEAR that is set in constants (currently to 2009). I imagine this will cause problems in the timeshift test, but since this section of code isn't run with the current switches, it was passing all tests on my computer. 

Let me know if there's a better way to be checking this particular chunk. Thanks!

R/constants.R
```diff
@@ -419,8 +419,15 @@ emissions.AGR_GASES            <- c("CH4_AGR", "N2O_AGR", "NH3_AGR", "NOx_AGR")
 emissions.AG_MACC_GHG_NAMES    <- c("CH4_AGR", "N2O_AGR")
 emissions.GHG_NAMES            <- c("CH4", "N2O")
 emissions.USE_GV_MAC           <- 1
+
+emissions.USE_GCAM3_CCOEFS     <- 1 #Select whether to use GCAM3 fuel carbon coefficients
```
bpbond: Space after `#` please

R/zchunk_L202.Ccoef.R
```diff
@@ -1,24 +1,26 @@
 #' module_energy_L202.Ccoef
 #'
-#' Briefly describe what this chunk does.
+#' This chunk creates the file of CO2 coefficients for primary energy sources.
```
bpbond: What is a CO2 coefficient? Maybe say something like "Create the file of CO2 coefficient (<definition>) for primary energy sources."

R/zchunk_L202.Ccoef.R
```diff
@@ -30,43 +32,63 @@ module_energy_L202.Ccoef_DISABLED <- function(command, ...) {
...
-    # NOTE: there are 'match' calls in this code. You probably want to use left_join_error_no_match
-    # For more information, see https://github.com/JGCRI/gcamdata/wiki/Name-That-Function
+    # Carbon contents of fuels in all regions
+    A_PrimaryFuelCCoef %>%
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["CarbonCoef"]]), GCAM_region_names = GCAM_region_names, has_traded = TRUE) -> L202.CarbonCoef
+
+
+    # For regions with no agricultural and land use sector (Taiwan),
+    # need to remove the passthrough supplysectors for first-gen biofuels
+    ag_en <- c( "regional corn for ethanol", "regional sugar for ethanol", "regional biomassOil" )
```
bpbond: No GCAM-style spacing around parentheses pls

R/zchunk_L202.Ccoef.R
```diff
@@ -30,43 +32,63 @@ module_energy_L202.Ccoef_DISABLED <- function(command, ...) {
...
+        L102.Ccoef_kgCGJ_F_Yh %>%
+          filter(year %in% HISTORICAL_YEARS) %>%
+          filter(year == emissions.INVENTORY_MATCH_YEAR) -> L202.Ccoef_kgCGJ_F_fby
+
+        L202.CarbonCoef %>%
+          left_join(L202.Ccoef_kgCGJ_F_fby, by = "fuel") %>%
+          mutate(PrimaryFuelCO2Coef = if_else(!is.na(fuel), round(value, emissions.DIGITS_CO2COEF), PrimaryFuelCO2Coef)) %>%
+          select(one_of(LEVEL2_DATA_NAMES[["CarbonCoef"]])) -> L202.CarbonCoef
+      }
+      if(emissions.USE_GLOBAL_CCOEFS == 0) {
```
bpbond: Probably better: `} else if(emissions.USE_...`

R/zchunk_L202.Ccoef.R
```diff
@@ -30,43 +32,63 @@ module_energy_L202.Ccoef_DISABLED <- function(command, ...) {
...
+        L102.Ccoef_kgCGJ_R_F_Yh %>%
+          filter(year %in% HISTORICAL_YEARS) %>%
+          filter(year == emissions.INVENTORY_MATCH_YEAR) %>%
+          left_join(GCAM_region_names, by = "GCAM_region_ID") -> L202.Ccoef_kgCGJ_R_F_fby
+
+        L202.CarbonCoef %>%
+          left_join(L202.Ccoef_kgCGJ_R_F_fby, by = c("fuel", "region")) %>%
+          mutate(PrimaryFuelCO2Coef = if_else(!is.na(fuel), round(value, emissions.DIGITS_CO2COEF), PrimaryFuelCO2Coef)) %>%
+          select(one_of(LEVEL2_DATA_NAMES[["CarbonCoef"]])) -> L202.CarbonCoef
+      }
```
bpbond: `} else stop("Unknown emissions.USE_GCAM3_CCOEFS value ", emissions.USE_GCAM3_CCOEFS)`

inst/extdata/emissions/A_PrimaryFuelCCoef.csv
```diff
@@ -0,0 +1,29 @@
+# File: A_PrimaryFuelCCoef.csv,,
+# Title: Global Default Emissions Factors,,
+# Description: assumed global default emissions factors,,
+# Units: kg C per GJ,,
```
bpbond: What's the `Source` of these data? Let's document if possible.

R/zchunk_L202.Ccoef.R
```diff
@@ -30,43 +32,63 @@ module_energy_L202.Ccoef_DISABLED <- function(command, ...) {
...
+
+
+    # For regions with no agricultural and land use sector (Taiwan),
+    # need to remove the passthrough supplysectors for first-gen biofuels
+    ag_en <- c( "regional corn for ethanol", "regional sugar for ethanol", "regional biomassOil" )
+    L202.CarbonCoef %>%
+      filter(!region %in% aglu.NO_AGLU_REGIONS | !PrimaryFuelCO2Coef.name %in% ag_en) -> L202.CarbonCoef
+
+    # if emissions.USE_GCAM3_CCOEFS = 1, using exogenous C coefs from GCAM 3.0
+    if(emissions.USE_GCAM3_CCOEFS == 0) {
```
bpbond: @kvcalvin @d3y419 Should the data system _ever_ have to use GCAM3 coefficients, going forward? Can this code just be deleted?

jhoring: @pkyle maybe has thoughts? To my knowledge (based on my and Marshall's GCAM-core) we are currently using the GCAM3 coefficients. This has some implications for the OTAQ project if this changes.

pkyle: This looks fine to me. Unlike the other instances of "GCAM3" in the data system, these values are still in the core model. For accuracy they could be renamed "MiniCAM", because that's really where they come from.
We should update that at some point, in order to calibrate CO2 emissions volumes in the base years. I.e., GCAM's CO2 emissions in 1990, 2005, and 2010 are different from all inventories, globally, by region, and by fuel. But that switch might make its own issues; the coefs we have always used are high for coal and low for gas.

R/zchunk_L202.Ccoef.R
jhoring: @bpbond quick question on this -- if we do want to replicate the old data system, we wouldn't want this to stop if USE_GCAM3_CCOEFS = 1. Did you want me to add a stop / warning for cases that are not 0 or 1?

bpbond: That’s what I was thinking, yes. Thanks.

R/zchunk_L202.Ccoef.R
```diff
@@ -1,72 +1,104 @@
...
+      L202.CarbonCoef <- L202.CarbonCoef
+    }
+    else if(emissions.USE_GCAM3_CCOEFS == 0) {
+      L202.CarbonCoef %>%
+        left_join(fuel_to_Ccoef, by = "PrimaryFuelCO2Coef.name") -> L202.CarbonCoef
+      if(emissions.USE_GLOBAL_CCOEFS == 1) {
+        # Using global average emissions coefficients based on CDIAC inventory and IEA energy balances
+        L102.Ccoef_kgCGJ_F_Yh %>%
+          filter(year %in% HISTORICAL_YEARS) %>%
+          filter(year == emissions.INVENTORY_MATCH_YEAR) -> L202.Ccoef_kgCGJ_F_fby
```
kvcalvin: Do we need both of these filters?

R/zchunk_L202.Ccoef.R
```diff
@@ -1,72 +1,104 @@
...
+        L202.CarbonCoef %>%
+          left_join(L202.Ccoef_kgCGJ_F_fby, by = "fuel") %>%
+          mutate(PrimaryFuelCO2Coef = if_else(!is.na(fuel), round(value, emissions.DIGITS_CO2COEF), PrimaryFuelCO2Coef)) %>%
+          select(one_of(LEVEL2_DATA_NAMES[["CarbonCoef"]])) -> L202.CarbonCoef
+      }
+      else if(emissions.USE_GLOBAL_CCOEFS == 0) {
+        # Using region-specific average emissions coefficients based on CDIAC inventory and IEA energy balances
+        L102.Ccoef_kgCGJ_R_F_Yh %>%
+          filter(year %in% HISTORICAL_YEARS) %>%
+          filter(year == emissions.INVENTORY_MATCH_YEAR) %>%
```
kvcalvin: Same question as above.

jhoring:
@bpbond I'm failing the travis checks but passing on my computer. I've already updated GCAM_DATA_MAP. Any other ideas what could be going on?

R/zchunk_L202.Ccoef.R
bpbond: Hmm. Either way, I don't think the first test is necessary; if it's not failing now, it won't fail without the HISTORICAL_YEARS line.

bpbond:
@jhoring I'll look.

bpbond:
From the Travis error, it looks like a GCAM_DATA_MAP failure. I'll try rebuilding it (again).

R/zchunk_L202.Ccoef.R
jhoring: @bpbond ok I'll remove that and then it looks like the rebuild worked and we're passing tests on here!

bpbond:
Hurray! 🎉

Issue #598, opened 2017-08-03
L2322.Fert


bpbond:

bpbond:
Jill, if you haven't started working on this I'm going to shift over to @fengly20 - ok?

jhoring:
Yep. I'm finishing up L224.heat now but I'll ping you for an upnext when I submit?

bpbond:
Great, thanks. I just assigned you a new one.

fengly20:
@bpbond It seems there's no script named zchunk_L2322.Fert .R in current repository. Is that right? Or did I miss something somehow?

bpbond:
Nope! Sorry. Hang on, let me add it.

fengly20:
@bpbond Thanks!

fengly20:
@bpbond Just wanted to make sure, the files you added (R/zchunk_L2322.Fert_USA.R and etc.) are for #724  not for #598, right?

bpbond:
Whoops, sorry about that. Fixing.

Issue #737, opened 2017-09-14
Help needed with L2322.Fert assumption file units 


fengly20:
The chunk reads in assumption file `A322.globaltech_retirement` which I'm not sure about the units of the values. 

It has variables with more than one unit in the same table:

`lifetime` -- year? 
`shutdown.rate` -- Unitless? 
`half.life` -- year? 
`steepness` -- Unitless? 
`median.shutdown.point` -- Unitless? 
`profit.shutdown.steepness` -- Unitless? 

Thanks!

pkyle:
All of these parameters should be documented with units defined in other files (see any other input csv file that ends in the string "_retirement.csv"). Lifetime and half-life are years. Everything else is unitless. Textual descriptions should be copied from others rather than re-defined; that said, if the other "_retirement.csv" files are inadequately documented, let me know and we can update 'em all!

fengly20:
Thank you Page! @pkyle 

I quickly checked all '_retirement.csv' files in the ./gcam-data-system-OLD/energy-data/assumptions. There are four files in total and none of them has much documentation and unit information.

pkyle:
Huh, hope it's just the retirement parameters we're missing! Here are descriptions of each parameter which could be copied into the other "_retirement.csv" files as well.
lifetime: maximum lifetime of cohort. If no retirement function is used, the entire cohort is retired in this number of years.
shutdown.rate: portion of cohort retired each year, used by the phased-shutdown-decider retirement function
half.life: number of years at which 50% of the cohort is retired, using the s-curve-shutdown-decider retirement function.
steepness: shape parameter used by the s-curve-shutdown-decider retirement function.
median.shutdown.point: profit-ratio at which 50% of the cohort is shut down for lack of profitability.
profit.shutdown.steepness: shape parameter of profit-shutdown-decider function.

bpbond:
@fengly20 Please close this when appropriate.

fengly20:
Thank you Page! @pkyle

Hi Ben @bpbond , as Page mentioned in his comments, the information about variables could be as well added in other "_retirement.csv" energy input files used by other chunks but those chunks have not been re-written thus the input files do not exist in the ./inst/extdata/energy. Do we have a strategy in handling input files in advance of processing a chunk? Thanks.

bpbond:
Not particularly. Since it's only four files, it would be great if you could do them all at once--grab from the old data system give appropriate headers, and put into new data system. Thanks!

[...]

PR #765, opened 2017-09-25
LF - L2322.FERT

fengly20:
#598 

Finished re-write and passed old-new test.


inst/extdata/energy/A22.globaltech_retirement.csv
```diff
@@ -0,0 +1,38 @@
+# File: A22.globaltech_retirement.csv
+# Title: Primary energy transformation technology lifetimes and shutdown rates
```
bpbond: Any `Source:` for these files? @pkyle ?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+    A322.subsector_logit %>%
+      write_to_all_regions(LEVEL2_DATA_NAMES[["SubsectorLogit"]], GCAM_region_names) ->
+      L2322.SubsectorLogit_Fert
+
+    # L2322.SubsectorShrwt_Fert and L2322.SubsectorShrwtFllt_Fert: Subsector shareweights of fertilizer sector
+    tibble(x = NA) %>%
+      add_units("None") %>%
+      add_comments("Not generated") %>%
+      add_flags(FLAG_NO_TEST) ->
+      L2322.SubsectorShrwt_Fert
```
bpbond: Why are we even outputting this if it's never going to be generated?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      mutate(share.weight = approx_fun(year, share.weight, rule = 1)) %>%
+      ungroup %>%
+      filter(year %in% c(BASE_YEARS, FUTURE_YEARS)) %>%
+      rename(sector.name = supplysector, subsector.name = subsector) ->
+      L2322.GlobalTechShrwt_Fert
+
+    # L2322.GlobalTechCoef_Fert: Energy inputs and coefficients of global fertilizer energy use and feedstocks technologies
+    A322.globaltech_coef %>%
+      gather(year, coefficient, matches(YEAR_PATTERN)) %>%
+      mutate(year = as.numeric(year)) ->
```
bpbond: `as.integer`

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      filter(year %in% c(BASE_YEARS, FUTURE_YEARS)) %>%
+      rename(sector.name = supplysector, subsector.name = subsector) ->
+      L2322.GlobalTechCoef_Fert
+
+    # Costs of global technologies
+    # L2322.GlobalTechCost_Fert: Non-energy costs of global fertilizer manufacturing technologies
+    L2322.GlobalTechCoef_Fert %>%
+      select(LEVEL2_DATA_NAMES[["GlobalTechYr"]]) %>%
+      mutate(minicam.non.energy.input = "non-energy") %>%
+      left_join(L1322.Fert_NEcost_75USDkgN_F, by = c('technology' = 'fuel') ) %>%
```
bpbond: Why no `left_join_error_no_match`? Are we expecting multiple matches or NAs?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      L2322.globaltech_retirement
+
+    # Retirement may consist of any of three types of retirement function (phased, s-curve, or none)
+    # All of these options have different headers, and all are allowed
+
+    # L2322.GlobalTechShutdown_Fert: Global tech lifetime and shutdown rate
+    tibble(x = NA) %>%
+      add_units("None") %>%
+      add_comments("Not generated") %>%
+      add_flags(FLAG_NO_TEST) ->
```
bpbond: Again, if they contain no data and aren't being tested, why are we generating these?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      L2322.GlobalTechShutdown_Fert
+
+    # L2322.GlobalTechSCurve_Fert: Global tech lifetime and s-curve retirement function
+    L2322.globaltech_retirement %>%
+      filter(!is.na(half.life)) %>%
+      select(LEVEL2_DATA_NAMES[["GlobalTechYr"]], "lifetime", "steepness", "half.life") ->
```
bpbond: Probably need to change this `select(one_of(c(...)))` for dplyr 0.5 compatibility

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      filter(year %in% BASE_YEARS) %>%
+      rename(calOutputValue = value) %>%
+      mutate(calOutputValue = round(calOutputValue, energy.DIGITS_CALOUTPUT)) %>%
+      left_join_error_no_match(GCAM_region_names, by = 'GCAM_region_ID') %>%
+      left_join_error_no_match(select(calibrated_techs, sector, fuel, supplysector, subsector, technology), by = c("sector", "fuel") ) %>%
+      rename(stub.technology = technology) %>%
+      mutate(share.weight.year = year) %>%
+      mutate(subs.share.weight = 0) %>%
+      mutate(subs.share.weight = replace(subs.share.weight, which(calOutputValue > 0), 1)) %>%
+      mutate(tech.share.weight = subs.share.weight) %>%
```
bpbond: Combine into a single `mutate` call

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,456 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+    L142.ag_Fert_NetExp_MtN_R_Y %>%
+      filter(year %in% BASE_YEARS) %>%
+      rename(fixedOutput = value) %>%
+      left_join_error_no_match(GCAM_region_names, by = 'GCAM_region_ID') %>%
+      mutate(supplysector = A322.globaltech_renew[["supplysector"]], subsector = A322.globaltech_renew[["subsector"]], stub.technology = A322.globaltech_renew[["technology"]]) %>%
+      mutate(fixedOutput = pmax(0, -1 * fixedOutput)) %>%
+      mutate(fixedOutput = round(fixedOutput, energy.DIGITS_CALOUTPUT)) %>%
+      mutate(share.weight.year = year) %>%
+      mutate(subs.share.weight = 0) %>%
+      mutate(tech.share.weight = 0) %>%
```
bpbond: Ditto

bpbond:
@fengly20 Very solid overall--thanks! Some changes for clarity and efficiency would be good.

inst/extdata/energy/A22.globaltech_retirement.csv
pkyle: Do you mean A322.globaltech_retirement? Not sure why A22 would be called by this code. In either case, the capital stock survival curves are GCAM assumptions for which literature estimates are not available. The A22 assumptions were documented as thoroughly as they ever will be in Proposal 202, but it was not uploaded to Confluence so there's no link.

R/zchunk_L2322.Fert.R
pkyle: If users read in data to the "year" column of A322.subsector_shrwt.csv, these tables will need to be written out. So, agreed that there's no point in writing out these tables as coded, because they'll be non-existent no matter what. They should just be tables whose nrow = 0 right now, but that would exist if years were read in to the year column. To do this, I'd suggest just temporarily switching the data in the "year.fillout" column to the "year" column.

R/zchunk_L2322.Fert.R
pkyle: Might be nbd but this might also be an issue more broadly...the old data system had lots of level2 data files that weren't written out in the core assumptions, but if users modified certain assumptions, then they would be generated. We might be losing some capabilities there, since those files that aren't generated for the core set of assumptions aren't being tested.

inst/extdata/energy/A22.globaltech_retirement.csv
fengly20: `A22.globaltech_retirement.csv` was not called by this chunk. I inserted comments for all "*_retirement.csv" and added all assumption files into the new system as we discussed in #737

R/zchunk_L2322.Fert.R
fengly20: All the empty/no-test data tables generated in this chunk were conditional outputs in the old chunk. Creating an empty/no-test data table for conditional output is how we deal with such situation as discussed in #633

R/zchunk_L2322.Fert.R
fengly20: Yes. NAs are expected. Comments added.

fengly20:
Thanks Ben! @bpbond 

The chunk has been updated to address your comments.

R/zchunk_L2322.Fert.R
pkyle: This is different from #633; in that case, the issue was that the old code was referencing a non-existent CSV, and it only didn't crash because it was within a conditional statement that (in testing and in application) was always FALSE. So the matter discussed in #633 was whether to add functionality that wasn't there to begin with, and we decided not to because it was not useful.
The difference here is that the present approach is dropping existing functionality. Where users can currently set share-weight paths year-by-year (in the "year" column of A322.subsector_shrwt), and choose between different retirement functions (s-curve and phased-shutdown in A322.globaltech_retirement), that capability is no longer supported in this present approach. That might be OK for this particular sector, and that's a different discussion, but I agree with @bpbond that if we do decide to ditch this capability, there is no point in making x=NA tables.

fengly20: Thanks for your note! @pkyle

So how do we decide for those empty tables? Remove them all? @bpbond

R/constants.R
```diff
@@ -254,7 +254,14 @@ energy.RSRC_FUELS <- c("coal", "gas", "refined liquids")
 energy.DIGITS_EFFICIENCY <- 3
 energy.DIGITS_COST <- 4
 energy.DIGITS_REMOVE.FRACTION <- 2
+
+# below are constants needed by chunk L2322.Fert.R
 energy.CO2.STORAGE.MARKET <- "carbon-storage"
+energy.DIGITS_COEFFICIENT <- 7
+energy.DIGITS_REMOVE.FRACTION <- 2
```
kvcalvin: This seems to be duplicated above. Do we need it a second time?

R/zchunk_L2322.Fert.R
kvcalvin: Should we write the code so it will work whether the "year" column exists or not?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+    #                        for interpolation years, and finally performs interpolation using 'rule=1' as in
+    #                        old data system. For more information about 'rule' please see ?approx_fun.
+    #                        Same workflow applies at several places in this chunk.
+    A322.globaltech_shrwt %>%
+      gather(year, share.weight, matches(YEAR_PATTERN)) %>%
+      mutate(year = as.numeric(year)) ->
+      A322.globaltech_shrwt_long
+
+    df_years <- unique(A322.globaltech_shrwt_long$year)
+    int_years <- setdiff(c(BASE_YEARS, FUTURE_YEARS), df_years)
```
kvcalvin: Do we need to do this? The `Name that Function` part of the wiki has code snippets to replicate this behavior: https://github.com/JGCRI/gcamdata/wiki/Name-That-Function

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      L2322.GlobalTechCapture_Fert
+
+    # Retirement information
+    A322.globaltech_retirement %>%
+      set_years %>%
+      mutate(year = as.numeric(year)) %>%
+      rename(sector.name = supplysector, subsector.name = subsector) ->
+      A322.globaltech_retirement_with_years
+
+    # Copy the data in the first future period through to the end year
```
kvcalvin: The pipeline below is getting the max of the base years, not the first future period. Do you man copy the last historical year to the future years?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+    # Calibration and region-specific data
+    # L2322.StubTechProd_Fert: calibrated output of fertilizer technologies
+    L1322.Fert_Prod_MtN_R_F_Y %>%
+      filter(year %in% BASE_YEARS) %>%
+      rename(calOutputValue = value) %>%
+      mutate(calOutputValue = round(calOutputValue, energy.DIGITS_CALOUTPUT)) %>%
+      left_join_error_no_match(GCAM_region_names, by = 'GCAM_region_ID') %>%
+      left_join_error_no_match(select(calibrated_techs, sector, fuel, supplysector, subsector, technology), by = c("sector", "fuel") ) %>%
+      rename(stub.technology = technology) %>%
+      mutate(share.weight.year = year, subs.share.weight = 0, subs.share.weight = replace(subs.share.weight, which(calOutputValue > 0), 1), tech.share.weight = subs.share.weight) %>%
```
kvcalvin: Wouldn't it be easier to do an `if_else` on `subs.share.weight.`

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
+      rename(fixedOutput = value) %>%
+      mutate(fixedOutput = round(fixedOutput, energy.DIGITS_CALOUTPUT)) %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID" ) %>%
+      mutate(supplysector = filter(A322.globaltech_shrwt, grepl( "Exports", supplysector))[["supplysector"]]) %>%
+      mutate(subsector = filter(A322.globaltech_shrwt, grepl( "Exports", supplysector))[["subsector"]]) %>%
+      mutate(stub.technology = filter(A322.globaltech_shrwt, grepl( "Exports", supplysector))[["technology"]]) %>%
+      mutate(fixedOutput = pmax(0, fixedOutput), share.weight.year = year, subs.share.weight = 0, tech.share.weight = 0) ->
+      L2322.StubTechFixOut_Fert_exp_base
+
+    L2322.StubTechFixOut_Fert_exp_base %>%
```
kvcalvin: Can't we combine this pipeline with the previous?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    L2322.Supplysector_Fert %>%
+      add_title("Supply sector information for fertilizer sector") %>%
+      add_units("NA") %>%
+      add_comments("Expanded fertilizer sector supply sector information for all GCAM regions") %>%
```
kvcalvin: Could you include more details? e.g., what is in this information (is it logits, etc.)? And, where does it come from (i.e., an assumption file)?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.SubsectorLogit_Fert %>%
+      add_title("Subsector logit exponents of fertilizer sector") %>%
+      add_units("NA") %>%
+      add_comments("Expanded fertilizer sector subsector logit exponents for all GCAM regions") %>%
```
kvcalvin: See above

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.SubsectorShrwtFllt_Fert %>%
+      add_title("Subsector shareweights of fertilizer") %>%
+      add_units("Unitless") %>%
+      add_comments("Expanded Subsector shareweights of fertilizer to all GCAM regions") %>%
```
kvcalvin: See above. These ones are more specific about what is in them, but still need information on where it came from.

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.StubTech_Fert %>%
+      add_title("Identification of stub technologies of fertilizer sector") %>%
+      add_units("NA") %>%
+      add_comments("Expanded identification of stub technologies of fertilizer sector to all GCAM regions") %>%
```
kvcalvin: What does "identification" mean?

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
-      add_legacy_name("L2322.GlobalTechCoef_Fert") %>%
-      add_precursors("precursor1", "precursor2", "etc") %>%
-      # typical flags, but there are others--see `constants.R`
-      add_flags(FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR) ->
+    L2322.GlobalTechCoef_Fert %>%
+      add_title("Energy inputs and coefficients of global fertilizer energy use and feedstocks technologies") %>%
+      add_units("Unitless") %>%
```
kvcalvin: I doubt these are unitless

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.GlobalTechCost_Fert %>%
+      add_title("Non-energy costs of global fertilizer manufacturing technologies") %>%
+      add_units("1975USDkgN") %>%
+      add_comments("Claculated cost from L1322.Fert_NEcost_75USDkgN_F") %>%
```
kvcalvin: `Calculated`

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
       L2322.GlobalTechCapture_Fert
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.GlobalTechShutdown_Fert %>%
+      add_title("Global tech lifetime and shutdown rate") %>%
+      add_units("NA") %>%
```
kvcalvin: I think lifetime is in years

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    L2322.GlobalTechProfitShutdown_Fert %>%
+      add_title("Global tech profit shutdown decider") %>%
+      add_units("Unitless") %>%
+      add_comments("Extracted from L2322.globaltech_retirement") %>%
```
kvcalvin: What does that mean?

R/zchunk_L2322.Fert.R
pkyle: Should be "GJ per kg N"

R/zchunk_L2322.Fert.R
```diff
@@ -76,272 +76,447 @@ module_energy_L2322.Fert_DISABLED <- function(command, ...) {
...
       L2322.GlobalTechCoef_Fert
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    L2322.GlobalTechCost_Fert %>%
+      add_title("Non-energy costs of global fertilizer manufacturing technologies") %>%
+      add_units("1975USDkgN") %>%
```
bpbond: What are these units? 1975 USD/kg N?

pkyle: Yup!

bpbond: OK, please clarify @fengly20

fengly20:
Sorry about the pause (Monday is my dsr day). The chunk has been updated to address comments from @kvcalvin (and others). The empty tibbles in previous discussion have been removed from the chunk. The documentation has been updated accordingly. The chunk passed old-new test and time shift.

R/constants.R
fengly20: Duplicated entries removed.

R/zchunk_L2322.Fert.R
fengly20: I removed the empty tibbles for now unless we decide to do something else.

R/zchunk_L2322.Fert.R
fengly20: Thanks! That was the old ugly way I did before the wiki got updated with the elegant solution.

R/zchunk_L2322.Fert.R
fengly20: Comment corrected.

R/zchunk_L2322.Fert.R
fengly20: Thanks! I had a false impression that `if_else` is forbidden. Turned out only `ifelse` cannot be used.

R/zchunk_L2322.Fert.R
fengly20: `L2322.StubTechFixOut_Fert_exp_base` will be used later to generate another tibble.

R/zchunk_L2322.Fert.R
fengly20: Copied from old chunk. The table contains Stub-technology (coal, coal CCS, and etc.) for fertilizer sector. Any suggestion to replace the word `identification`?

R/zchunk_L2322.Fert.R
fengly20: Particularity for this one it was an empty tibble so no unit indicated. The tibble now has been removed from the chunk.

R/zchunk_L2322.Fert.R
bpbond: What about just what you said: "Stub-technology (coal, coal CCS, and etc.) for fertilizer sector"?

R/zchunk_L2322.Fert.R
fengly20: So is it in general that the `coefficient` should be having a unit?  

The `coefficient` in `L2322.GlobalTechCoef_Fert` came from `A322.globaltech_coef` which doesn't have much information before I added it into data folder. I also randomly checked a few other `A*_coef.csv` files in the new data system and they all have 'Unitless' in comments. 

Apologies in advance if I totally get it wrong.

R/zchunk_L2322.Fert.R
fengly20: Sounds good. Thanks.

bpbond:
@fengly20 this is erroring:
```R
 Error: All select() inputs must resolve to integer column positions.
The following do not:
*  LEVEL2_DATA_NAMES[["GlobalTechYr"]] 
```

R/zchunk_L2322.Fert.R
pkyle: [Learning opportunity!] With the exception of the infamous "tranTechnology", in GCAM the "coefficient" of any technology and input simply represents the ratio between units of input and units of output. The units of the "output" of any technology are indicated in the tag "output-unit" specified in the corresponding "A\*.sector.csv" file. For the most part, this is also a reliable source of the "input-unit" though in some cases, different inputs may have different units (e.g., cement takes inputs of limestone (Mt) and heat (EJ)). The fail-safe way of determining the model's units of any input is to find the supplysector or resouce of the same name and check its output unit. In this case, fertilizer's output unit is Mt and inputs are EJ, so the (input-output) coefficient is EJ/Mt, which we simplify by power of 9 to GJ/kg. GCAM's standard of simplifying indicators by powers of 9 comes from the order of magnitude of the global population.

bpbond: I thought this was useful enough that I put it [into the wiki](https://github.com/JGCRI/gcamdata/wiki/Units).

bpbond:
@fengly20 Can I ask you to take another look at this please. I've fixed the dplyr 0.5 syntax problems, but this is still giving an error in line 246 I think:

```R
mutate(supplysector = A322.globaltech_renew[["supplysector"]]
```

```
 Error in eval(expr, envir, enclos) : incorrect length (0), expecting: 7 
```

fengly20:
Sure. I'm checking right now.

R/zchunk_L2322.Fert.R
fengly20: Thanks! @pkyle 

@bpbond Turns out in the `L2322.GlobalTechCoef_Fert` there would be two kinds of unit existing: GJ/kg N for supplysector N fertilizer and Unitless for supplysector Exports_fertilizer. 

Does the test accept something like?

```
add_units("GJ/kg N for supplysector N fertilizer; Unitless for supplysector Exports_fertilizer")
```

bpbond: Sure. The test just verifies that you've entered _something_ for units.

fengly20:
@bpbond 

Emmm...the pipeline containing line 246 works fine for me. In line 246 basically I'm just trying to select a column without using `$`. If the use of `$` is not forbidden, I can try that to see whether it works for you. 

The chunk passed old-new test on my end. 

Attached is my `sessionInfo()`

```
> sessionInfo()
R version 3.2.2 (2015-08-14)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] bindrcpp_0.2 gcamdata_0.8

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.9       tidyr_0.6.3       dplyr_0.7.2       assertthat_0.2.0  R6_2.2.0          magrittr_1.5     
 [7] rlang_0.1.1       stringi_1.1.2     lazyeval_0.2.0    data.table_1.10.4 tools_3.2.2       readr_1.1.1      
[13] glue_1.1.1        hms_0.3           pkgconfig_2.0.1   bindr_0.1         tibble_1.3.3   
```

bpbond:
OK, thanks, I'll take another look.

bpbond:
Note warning message was being printed:
```
Inputs energy/calibrated_techs don't appear as precursors for any outputs - chunk module_energy_L2322.Fert
```
I have fixed this.

fengly20:
Thanks! @bpbond

PR #771, opened 2017-09-28
Chris L222.En_Transformation

cwroney:
[#635](https://github.com/JGCRI/gcamdata/issues/635)


inst/extdata/energy/A22.globaltech_co2capture.csv
```diff
@@ -0,0 +1 @@
...
# Title: Primary energy transformation technology capture fractions
# Units: fractions
# Description: Fraction of CO2 captured by transformation technologies
supplysector,subsector,technology,1971,2100
refining,coal to liquids,coal to liquids CCS level 1,0.818,0.818
refining,coal to liquids,coal to liquids CCS level 2,0.981,0.981
refining,biomass liquids,cellulosic ethanol CCS level 1,0.26,0.26
refining,biomass liquids,cellulosic ethanol CCS level 2,0.992,0.992
refining,biomass liquids,FT biofuels CCS level 1,0.818,0.818
refining,biomass liquids,FT biofuels CCS level 2,0.98,0.98
```
bpbond: Please add missing final blank line (here and in any other files that require).

R/zchunk_L222.en_transformation.R
```diff
@@ -1,6 +1,8 @@
 #' module_energy_L222.en_transformation
 #'
-#' Briefly describe what this chunk does.
+#' This chunk prepares the assumptions and calibrated outputs for energy transformation supplysectors, subsectors, and technologies.
```
bpbond: This should be brief (single sentence). Put the rest in `@details` below, which I see you haven't filled out. (Note - I'll do this as am rebuilding docs.)

R/zchunk_L222.en_transformation.R
```diff
@@ -60,9 +60,15 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+    # Silencing global variable package checks
+    GCAM_region_ID <- calOutputValue <- coefficient <- fuel <- half.life <- input.cost <- lifetime <-
+    median.shutdown.point <- minicam.energy.input <- minicam.non.energy.input <- object <-
+    profit.shutdown.steepness <- region <- remove.fraction <- sector <- sector.name <- share.weight <-
+    shutdown.rate <- steepness <- stub.technology <- subsector <- subsector.name <- supplysector <-
+    technology <- to.value <- value <- year <- year.fillout <- year.x <- year.y <- NULL
+
     # Load required inputs
     GCAM_region_names <- get_data(all_data, "common/GCAM_region_names")
-    fuel_energy_input <- get_data(all_data, "energy/fuel_energy_input")
```
bpbond: Was this not used?

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+
+    # 2c. Technology information
+
+    # L222.StubTech_en: Identification of stub technologies of energy transformation
+    # set up filter to keep all non-first gen bio techs in L222.StubTech_en
+    firstgenbio_techs <- c("corn ethanol", "sugarbeet ethanol", "sugar cane ethanol", "biodiesel")
+
+    # create list of regional stub.technologies
+    A22.globaltech_shrwt %>%
+      write_to_all_regions(c("region", "supplysector", "subsector", "technology"), GCAM_region_names) %>%
```
bpbond: Why is there no use of `LEVEL2_DATA_NAMES` in this chunk?

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      rename(stub.technology = technology) %>%
+      # Drops region x technology combinations that are not applicable, i.e. any first gen bio techs not listed for a region in A_regions
+      filter(!(stub.technology %in% firstgenbio_techs) | paste(region, stub.technology) %in%
+               c(paste(A_regions$region, A_regions$ethanol), paste(A_regions$region, A_regions$biodiesel))) ->
```
bpbond: Could this be done with a `semi_join`? Seems like that would be a lot more elegant and clear.

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+    # L222.GlobalTechInterp_en: Technology shareweight interpolation of energy transformation sectors
+    A22.globaltech_interp %>%
+      set_years() %>%
+      rename(sector.name = supplysector, subsector.name = subsector) ->
+      L222.GlobalTechInterp_en
+
+    # L222.GlobalTechCoef_en: Energy inputs and coefficients of global technologies for energy transformation
+
+    A22.globaltech_coef %>%
+      gather(year, coefficient, -supplysector, -subsector, -technology, -minicam.energy.input) %>%
```
bpbond: `gather(year, coefficient, matches(YEAR_PATTERN))` probably clearer? Same comment applies below.

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      complete(nesting(supplysector, subsector, technology, minicam.energy.input), year = c(year, BASE_YEARS, FUTURE_YEARS)) %>%
+      arrange(supplysector, year) %>%
+      group_by(supplysector, subsector, technology, minicam.energy.input) %>%
+      mutate(coefficient = approx_fun(as.numeric(year), coefficient)) %>%
```
bpbond: For clarity and safety explicitly say `rule = x`

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      filter(year == max(BASE_YEARS)) %>%
+      repeat_add_columns(tibble("year" = as.character(FUTURE_YEARS))) %>%
+      select(-year.x) %>%
+      rename(year = year.y) %>%
+      bind_rows(L222.globaltech_retirement_base) ->
+      L222.globaltech_retirement
+
+    # Retirement may consist of any of three types of retirement function (phased, s-curve, or none)
+    # This section checks L222.globaltech_retirement for each of these functions and creates a separate level 2 file for each
+    # All of these options have different headers, and all are allowed
```
bpbond: Good comment–thanks.

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      filter(year %in% BASE_YEARS) %>%
+      # append region names
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID") ->
+      L222.out_EJ_R_gasproc_F_Yh_base
+
+    # append matching calibrated technology sector/subsector/technology to calibrated outputs of gas processing
+    calibrated_techs %>%
+      semi_join(L222.out_EJ_R_gasproc_F_Yh_base, by = c("sector", "fuel")) %>%
+      select(sector, fuel, supplysector, subsector, technology) %>%
+      left_join(L222.out_EJ_R_gasproc_F_Yh_base, by = c("sector", "fuel")) %>%
```
bpbond: Can't use `left_join_error_no_match`?

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+    # L222.StubTechProd_gasproc: calibrated output of gas processing technologies -- writes to all regions, adds calibrated tech info
+    A22.globaltech_coef %>%
+      filter(supplysector == "gas processing") %>%
+      write_to_all_regions(c("region", "supplysector", "subsector", "technology", "minicam.energy.input"), GCAM_region_names) %>%
+      rename(stub.technology = technology) %>%
+      repeat_add_columns(tibble("year" = BASE_YEARS)) %>%
+      left_join_error_no_match(L222.out_EJ_R_gasproc_F_Yh, by = c("region", "supplysector", "subsector", "stub.technology", "year")) %>%
+      # rounds outputs and adds year column for shareweights
+      mutate(calOutputValue = round(value, energy.DIGITS_CALOUTPUT), year.share.weight = year) %>%
+      select(-sector, -GCAM_region_ID, -fuel, -value, -minicam.energy.input) %>%
```
bpbond: In general please if possible `select` positively (inclusion instead of exclusion), as that's clearer.

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      # sets shareweight to 1 if output exists, otherwise 0
+      mutate(share.weight = if_else(calOutputValue > 0, 1, 0)) %>%
+      mutate(subs.share.weight = share.weight) ->
```
bpbond: Combine into single `mutate` call

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
       # typical flags, but there are others--see `constants.R`
       add_flags(FLAG_LONG_YEAR_FORM, FLAG_NO_XYEAR) ->
       L222.SubsectorShrwt_en
+  } else {
+    tibble(x = NA) %>%
+      add_title("Data not created") %>%
+      add_units("Unitless") %>%
+      add_comments("Data not created") %>%
+      add_legacy_name("energy/L222.SubsectorShrwt_en") ->
+      L222.SubsectorShrwt_en
```
bpbond: Won't you need a `FLAG_NO_TEST` on these? ...hmm, no?

R/zchunk_L222.en_transformation.R
bpbond: ...sensibly. Dropping a column using `-x` is fine, obviously.

R/zchunk_L222.en_transformation.R
cwroney: It only makes them if they don't exist in the old system, so it would be testing against a non-existent file? I guess that means it passes okay.

R/zchunk_L222.en_transformation.R
cwroney: I tried it first and it failed.

R/zchunk_L222.en_transformation.R
cwroney: it was not.

R/zchunk_L222.en_transformation.R
cwroney: The short answer is that I didn't know it existed or what it does? I was replacing all of the obtuse "names_StubTech" from the old system by explicitly calling the columns, e.g., `<write_to_all_regions(c("region", "supplysector", "subsector", "logit.year.fillout", "logit.exponent"), GCAM_region_names)>`, which seemed much clearer to me. Was I actually supposed to be referencing this database?

R/zchunk_L222.en_transformation.R
cwroney: oh, that's useful!

R/zchunk_L222.en_transformation.R
cwroney: It may be worth noting that the if statements that evaluate to `TRUE` return a `debug at #2:` when you're using `browser()` but they don't throw it into debug if you're not already in debug mode--it appears to work fine when you run the driver and it passes all the tests. I've quintuple checked the syntax, so it's not a problem with that.

R/zchunk_L222.en_transformation.R
bpbond: Ah, sorry for the missing information. Yes please, use `LEVEL2_DATA_NAMES`! Occasionally there's something missing, but almost all of the level 2 outputs should have an entry.

R/zchunk_L222.en_transformation.R
bpbond: So please add a comment explaining why: usually, because there are NA values in the incoming data frame (which we expect); or, less frequently, there are multiple matches, and we expect that too.

bpbond:
Holding on review by @mwisepnnl .

R/zchunk_L222.en_transformation.R
mwisepnnl: Does another chunk use the fuel_energy_input.csv?  I notice it is still in the files listed in the chunk

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +87,540 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+      # sets shareweight to 1 if output exists, otherwise 0
+      mutate(share.weight = if_else(calOutputValue > 0, 1, 0)) %>%
+      mutate(subs.share.weight = share.weight) ->
+      L222.StubTechProd_gasproc
+
+    # THE FOLLOWING IS IN THE ORIGINAL BUT ALREADY OMITS THOSE REGIONS WITH NO COAL GAS, SO THERE ARE NO MISSING VALUES, JUST NO ROWS
+    # IS THIS A MISTAKE? NOT SURE IF WE NEED THOSE ROWS. PASSES ALL CHECKS.
+    # "#Coal to gas isn't included in all regions so replace missing values
+    # L222.StubTechProd_gasproc$calOutputValue[ is.na( L222.StubTechProd_gasproc$calOutputValue ) ] <- 0"
+
```
mwisepnnl: I guess I am not sure what this comment is saying.  ARe the zero coal to gas regions just not listed in the input?  Or are they stripped out elsewhere?  Or, does code that follows this not require that all regions be included so that assigning the zeroes isn't necessary?  It's clear that I don't know, so explain what you can, and see if you need that step in or can delete it.

bpbond: @mwisepnnl Yes, `fuel_energy_input` is used by quite a few other chunks: L222, L223, L224, L226, L232, L242, L252.

mwisepnnl: Okay.  Then I guess I am waiting on just the coal to gas clarification.

cwroney: The short version is that there are no NAs, so it doesn't do anything. The long version is that it also doesn't do anything in the old data system which made me concerned that I was missing the point. I did a bit more looking and I think what's happening is that the zeroes are being generated upstream in L122, so this line is entirely redundant. Not sure why it existed in the first place though. I'll delete the comment.

R/zchunk_L222.en_transformation.R
cwroney: The best I've come up with is an anti_join followed by two semi_joins and then binding rows back together? Is there a way to use `semi_join` to replace `%in%` that allows you to use the boolean operators? Otherwise I'm creating 2 to 3 separate pipelines.

R/zchunk_L222.en_transformation.R
cwroney: No worries, the perils of starting this project after everyone else had gone through the training. I've made the changes and I'll commit once I've wrapped up the other changes.

R/zchunk_L222.en_transformation.R
bpbond: Do whatever seems most sensible to you. Thanks.

R/zchunk_L222.en_transformation.R
cwroney: @bpbond sorry one more question on this. There are items in LEVEL2_DATA_NAMES where the appropriate list renames columns: e.g. `L222.StubTechProd_refining` would seem to use `LEVEL2_DATA_NAMES[["StubTechProd"]]` but the old data system columns are `"region" "supplysector" "subsector" "stub.technology" "year" "calOutputValue" "year.share.weight" "subs.share.weight" "share.weight"` and using level2 would replace `share.weight.year` with `year.share.weight` and `share.weight` with `tech.share.weight`. Should I be matching the level2 data names and applying some kind of flag to it, manually matching the old data system list, or somehow trying to change the level2 data names? The old data system uses `names_StubTechYr` and then adds a bunch of custom columns.

bpbond: Follow the old data system please; use `LEVEL2_DATA_NAMES[["StubTechYr"]]` and then add columns as necessary.

inst/extdata/energy/A22.globaltech_co2capture.csv
cwroney: these files do have the final return but the formatting is getting broken in however github reads it...it's obviously not a single line (some of the others did not have it). Just an FYI that I checked them all for it, however it's showing up on here.

bpbond: This is a line ending problem then; the files have Windows line endings but we want Unix/Mac. I can fix this, don't worry about it.

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +93,550 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
+    }
 
+    if (exists("L222.SubsectorInterp_en")) {
     tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
+      add_title("Subsector shareweights of energy transformation sectors") %>%
+      add_units("unitless") %>%
       add_comments("comments describing how data generated") %>%
       add_comments("can be multiple lines") %>%
```
bpbond: Missing comments

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +93,550 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
       L222.GlobalTechCapture_en
 
+    if (exists("L222.GlobalTechShutdown_en")) {
     tibble() %>%
-      add_title("descriptive title of data") %>%
+      add_title("Global tech lifetime and shutdown rate") %>%
       add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
+      add_comments("Filters for any technology that uses a phased retirement function") %>%
       add_comments("can be multiple lines") %>%
```
bpbond: Here and elsewhere, need to remove this

R/zchunk_L222.en_transformation.R
```diff
@@ -81,248 +93,550 @@ module_energy_L222.en_transformation_DISABLED <- function(command, ...) {
...
       add_legacy_name("L222.GlobalTechShutdown_en") %>%
-      add_precursors("precursor1", "precursor2", "etc") %>%
+      add_precursors("energy/A22.globaltech_retirement") %>%
       # typical flags, but there are others--see `constants.R`
```
bpbond: Please remove these lines

cwroney:
I updated the missing comments. I've now done everything on my end, I believe.

bpbond:
Holding on @mwisepnnl approval.

Issue #769, opened 2017-09-27
L232.industry


bpbond:
Per @d3y419 . ~Russell, we know you have a lot backed up, DSR-wise, so please keep us posted about feasibility etc.~

bpbond:
Whoops, mistaken assignment, fixing.

fengly20:
@bpbond  When you have time can you add the files for this chunk?

PR #784, opened 2017-10-16
LY - L232.industry

fengly20:
#769

Finished re-write and passed tests. 

The chunk preserved two for-loops from the old chunk and I haven't figured out a way to remove those loops and change into pipeline style. Would you mind taking a look? @bpbond 

Thanks!


R/zchunk_L232.industry.R
```diff
@@ -89,295 +85,554 @@ module_energy_L232.industry_DISABLED <- function(command, ...) {
...
+             value = value.y) %>%
+      mutate(share = value/total) %>%
+      filter(year == max(BASE_YEARS)) %>%
+      mutate(fuelprefElasticity = 0) ->
+      L232.indenergy_fuel_shares
+
+    # Set fuel preference elasticities as indicated by exogenous rules
+    for( j in 1:nrow( L232.indenergy_fuel_shares ) ){
+      for( i in 1:nrow( A32.fuelprefElasticity ) ){
+        L232.indenergy_fuel_shares$fuelprefElasticity[j][
```
bpbond: Kind of confusing. I'm sure there's a more efficient way to do this but don't have time to drill down into this right now. @fengly20 please remind me to come back to this before we merge this PR.

mwisepnnl: It took me a while to figure out what this code was doing, though after going through it seems pretty simple. The description line you added in the A32.fuelprefelasticity.csv file helps, but I would like it if you added a sentence or two in this code explaining what it does rather than saying"by exogenous rule"

R/zchunk_L232.industry.R
```diff
@@ -89,295 +80,554 @@ module_energy_L232.industry_DISABLED <- function(command, ...) {
...
+      elec_ratio <- output.ratio <- . <- year.x <- output.ratio.x <- output.ratio.y <-
+      input.cost <- minicam.non.energy.input <- GCAM_region_ID <- value <-
+      calibrated.value <- sector.name <- subsector.name <- region <-
+      calOutputValue <- subs.share.weight <- calOutputValue.x <- calOutputValue.y <-
+      output_tot <- value.x <- value.y <- total <- fuelprefElasticity <- NULL
+
     # ===================================================
+    # 1. Perform computations
+    # Create tables to delete technologies and subsectors in regions where heat is not modeled as a fuel
+    has_not_heat <- filter(A_regions, has_district_heat == 0) # intermediate tibble
```
mwisepnnl: This is a much more succinct way to distinguish regions with and without district heat than the old code.  But with the old code.  Could you add more comments as to what is being done and why?

fengly20:
@bpbond Would you have a bit of time to review the for loops in this chunk? Thanks!

bpbond:
@fengly20 Yes, will do, sorry.

bpbond:
The `for` loops here are ugly and could I think be done with a join. But they're also inexpensive. I've tweaked them for efficiency, and provided a clear comment as to what's going on, but am otherwise inclined to leave them as is for now.

bpbond:
@mwisepnnl - any other comments? If not please approve.

fengly20:
Thanks Ben! @bpbond 

I'll revise the chunk, add missing outputs, address comments then commit the changes for review.

bpbond:
@fengly20 All looks good except same problem here as #806 #778 - the xml-generating chunks wasn't enabled (I have done this, `module_energy_batch_industry_xml`); there are a couple of its inputs that aren't present; and nothing is being done with the GCAM3, SSP, and gSSP outputs from `L232.industry`. Presumably this should be resolved in exactly the same way #806 is.

bpbond:
See last commit message and files added. No guarantees that this runs! @fengly20

fengly20:
Sure. Thanks Ben!

I was also working on the chunk and will do a few commits to fix things.

fengly20:
@bpbond 

The industry_xml and industry_SSP_xml chunks works fine now, but one of the chunk output 'L232.IncomeElasticity_ind_ssp1' cannot pass the old-new test, which is very weird. 
 
I manually compared the new chunk output with the test data in R but could not find any difference. Would you have any suggestions for things I can look for? 

Thanks!

bpbond:
Added FLAG_SUM_TEST for `L232.IncomeElasticity_ind_ssp1`:

```
> summary(comp$income.elasticity-comp$income.elasticity.old)
-1.110e-16  0.000e+00  0.000e+00  1.337e-18  0.000e+00  1.110e-16 
```

bpbond:
Nice job @fengly20 - thanks

PR #789, opened 2017-10-17
LY -  L2321.cement

fengly20:
#778 

Finished re-write and passed tests.


R/constants.R
```diff
@@ -269,6 +269,9 @@ energy.DIGITS_MPKM <- 0
 energy.DIGITS_SPEED <- 1
 energy.DIGITS_LOADFACTOR <- 2
 
+energy.CEMENT_CCS_COST_2000USDTCO2 <- 50 # Starting point of supply curve in Mahasenan et al 2003; come from ENERGY_ASSUMPTIONS/A_ccs_data.R
+energy.CO2_STORAGE_COST_1990_USDTC <- 42 # #From GCAM 1.0 inputs; come from ENERGY_ASSUMPTIONS/A_ccs_data.R
```
bpbond: Nicely documented, thanks! 

`# From`

R/constants.R
```diff
@@ -335,7 +338,11 @@ CONV_HA_M2 <- 1e4 # ha to m2
 CONV_BM2_M2 <- 1e9
 CONV_MILFT2_M2 <- 92900 # Million square feet to square meters
 CONV_FT2_M2 <- 0.0929 # Square feet to square meters
+CONV_C_CO2 <- 44/12 # come from _common/assumptions/units_conversions.R
```
bpbond: Doesn't this already occur elsewhere is `constants.R`?

R/constants.R
```diff
@@ -335,7 +338,11 @@ CONV_HA_M2 <- 1e4 # ha to m2
...
 
+# Dollar conversion factors
+CONV_2000_1975_USD <- 0.38 # come from _common/assumptions/units_conversions.R
+CONV_1990_1975_USD <- 0.4649 # come from _common/assumptions/units_conversions.R
```
bpbond: These two shouldn't be used. Use the `gdp_deflator` function instead.

R/zchunk_L2321.cement.R
```diff
@@ -1,20 +1,23 @@
 #' module_energy_L2321.cement
 #'
-#' Briefly describe what this chunk does.
+#' The chunk provides final energy keyword, supplysector/subsector information,
+#' supplysector/subsector interpolation information, global technology share weight,
+#' global technology efficiency, global technology coefficients, global technology cost,
+#' price elasticity, stub technology information, stub technology interpolation information,
+#' stub technology calibrated inputs, and etc for cement sector.
```
bpbond: I'd prefer this be shortened, and spelled out in full in `@details` below. E.g. "Compute a variety of final energy keyword, sector, share weight, and technology information for cement-related GCAM inputs."

R/zchunk_L2321.cement.R
```diff
@@ -73,243 +70,360 @@ module_energy_L2321.cement_DISABLED <- function(command, ...) {
...
+
+    # Note: adjusting non-energy costs of technologies with CCS to include CO2 capture costs
+    # Note: The additional CCS-related non-energy costs are not included in the global technology assessment. Calculate here.
+    # First calculate the additional CCS costs per unit of carbon produced in 1975$
+    cement_CCS_cost_total_1975USDtC <- energy.CEMENT_CCS_COST_2000USDTCO2 * CONV_2000_1975_USD * CONV_C_CO2
+    CO2_storage_cost_1975USDtC <- energy.CO2_STORAGE_COST_1990_USDTC * CONV_1990_1975_USD
+    cement_CCS_cost_1975USDtC <- cement_CCS_cost_total_1975USDtC - CO2_storage_cost_1975USDtC
 
+    # Next calculate the quantity of CO2 produced per unit of cement produced (in kgC per kg cement)
+    cement_CO2_capture_frac <- mean( L2321.GlobalTechCapture_cement[["remove.fraction"]] )
```
bpbond: No spaces with parentheses please

R/zchunk_L2321.cement.R
```diff
@@ -73,243 +70,360 @@ module_energy_L2321.cement_DISABLED <- function(command, ...) {
...
+    CO2_IO_kgCkgcement <- mean(unlist(select(filter(L2321.GlobalTechCoef_cement, minicam.energy.input == "limestone"), coefficient))) *
```
bpbond: This should be clarified; also rather that `unlist` we've been using `.[["coefficient"]]` at end of a pipeline.

```R
L2321.GlobalTechCoef_cement %>%
  filter() %>%
  select() %>%
  .[["coefficient"]] %>%
  mean ->
  CO2_IO_kgCkgcement
```

R/zchunk_L2321.cement.R
```diff
@@ -73,243 +70,360 @@ module_energy_L2321.cement_DISABLED <- function(command, ...) {
...
+             subs.share.weight = if_else(calOutputValue > 0, 1, 0),
+             tech.share.weight = subs.share.weight) %>%
+      select(one_of(LEVEL2_DATA_NAMES[["StubTechProd"]])) ->
+      L2321.StubTechProd_cement
+
+    # L2321.StubTechCoef_cement: region-specific coefficients of cement production technologies
+    # Take this as a given in all years for which data is available
+    calibrated_techs %>%
+      select(sector, fuel, supplysector, subsector, technology, minicam.energy.input) %>%
+      unique ->
```
bpbond: `distinct` ?

R/constants.R
fengly20: Yes and it actually got defined twice under emissions section. Duplicates removed.

fengly20:
@bpbond 

I realized this might be an issue so I wanted to check. 

In `L2321.cement`, the old chunk produces some outputs which are not declared as outputs in the new chunk. 

For example, `L2321.IncomeElasticity_cement` is generated in the old chunk containing income elasticity for different scenarios then written out separately for different scenarios as `L2321.IncomeElasticity_cement_GCAM3.csv`, `L2321.IncomeElasticity_cement_ssp1.csv`, `L2321.IncomeElasticity_cement_ssp2.csv`, and etc. But either `L2321.IncomeElasticity_cement` or `L2321.IncomeElasticity_cement_[scen]` is declared in the new chunk.

Will those outputs get handled in downstream chunks or should I add them as new chunk outputs? 

Same situation for my other chunk `L232.industry`. 

Thanks!

bpbond:
@fengly20 Ugh, I'm not sure about this, especially the `L2321.IncomeElasticity_cement_[scen]` variants. I _guess_ we want to generate all of them? Or... if you can hold on this for a couple days, I'll sit down and look on Tuesday.

fengly20:
Sure. Thanks!

bpbond:
Hi @fengly20 - sorry for my slowness is dealing with this. Unless we're stripping out all GCAM3 outputs, which I don't think is the case, yes, you'll need to add all those scenario outputs to this chunk.

Basically, the skeleton-generating script isn't smart enough to 'see' outputs generated in a loop, as happens here.

bpbond:
Waiting on @skim301 review.

fengly20:
@bpbond 

When adding the missing outputs in the chunk, I've encountered two timeshift errors. The error messages are not quite helpful, since it's my first time having timeshift error. Would you mind telling me what is the general procedure to check the issue? 

```
2. Failure: chunks handle timeshift (@test_timeshift.R#34) ---------------------
exists("x") isn't true.
Timeshift error invoking module_energy_L2321.cement

3. Error: chunks handle timeshift (@test_timeshift.R#34) -----------------------
object 'x' not found
1: with_mock(run_chunk = function(chunk, all_data) {
       try({
           x <- do.call(chunk, list(driver.MAKE, all_data))
       })
       expect_true(exists("x"), info = paste("Timeshift error invoking", chunk))
       x
   }, driver(quiet = TRUE, write_outputs = FALSE)) at C:\Users\feng999\Documents\GitHub\gcamdata/tests/testthat/test_timeshift.R:34
2: with_mock_(.dots = .dots, .parent = .parent, .env = .env)
3: evaluate_with_mock_env(get_code_dots(dots), mock_env, .parent)
4: lazyeval::lazy_eval(code[[length(code)]])
5: eval(x$expr, x$env, emptyenv())
6: eval(expr, envir, enclos)
7: driver(quiet = TRUE, write_outputs = FALSE)
8: run_chunk(chunk, all_data[input_names]) at C:\Users\feng999\Documents\GitHub\gcamdata/R/driver.R:252
```

bpbond:
We don't have a lot of documentation about this, but see https://github.com/JGCRI/gcamdata/wiki/Common-Fails-and-How-to-Fix-Them

The easiest way to debug is to uncomment these lines in `constants.R`:

```R
# Uncomment these lines to run under 'timeshift' conditions
# HISTORICAL_YEARS <- 1971:2005       # normally 1971:2010
# FUTURE_YEARS <- seq(2010, 2100, 5)  # normally seq(2015, 2100, 5)
# BASE_YEARS <- c(1975, 1990, 2005)   # normally (1975, 1990, 2005, 2010)
# MODEL_YEARS <- c(BASE_YEARS, FUTURE_YEARS)
```
...and then rebuild and run the driver and your chunk. That should let you see exactly where the error is happening.

Feel free to ping me or e.g. Russell, Abigail, Kalyn, etc. for further tips on timeshift-debugging if useful.

bpbond:
...once done debugging, don't forget to re-comment those lines though!

fengly20:
Thanks Ben! @bpbond 

But I've got a related question... 

After I uncommented timeshift lines in `constants.R`, some of the chunk inputs (e.g. `L102.pcgdp_thous90USD_Scen_R_Y`)  started to have NAs/NaNs/Infs in the value column. Should those NAs be handled in my chunk or in the upstream chunk? 

Apologies in advance if I totally made it wrong.

bpbond:
Hmm. That's coming from `L102.GDP` by @rplzzz . I _think_ just handle the `NA`s as you need to, e.g. filter them out. Maybe also look at how other chunks handle this object–it looks like it's used by a number of them.

skim301:
I was wondering why this chunk is so long compared to the original. I see that a loop was used for the ssp scenario outputs. Is there a way to write out the ssp's in a loop? That would prevent some manual effort, make this chunk more compact and also ensure that if the scenario names change then we would still get proper outputs. @bpbond do you know of a chunk that can be used as an example to write out in a loop?

fengly20:
Thanks Ben! @bpbond 

In order to fix the timeshift error in my chunk, I ended up with using `left_join` (line362) instead of the more strict `left_join_error_no_match` with a comment explaining why. In this way the chunk won't be having any error thus can pass the timeshift test although **things might go wrong if timeshift is actually happening**, as in some other chunks. For example, under timeshift mode, the output `102.pcgdp_thous90USD_Scen_R_Y` of chunk `L102.GDP` will have NAs. Then the chunk `L244.building_det` uses it as input and produces output `L244.SatiationAdder` with only NAs in the value column. The chunk `L244.building_det` passes timeshift test because no actual error message is raised when running the chunk, like my chunk after using `left_join`.

The current chunk has one for loop as inherited from old chunk. The section calculates new values based on previous year values so using a for loop makes sense to me and I don't know whether would there be a pipeline solution.  

For the comments from @skim301 , yes, I can put those into a for loop if using loop is accepted in the '# Produce outputs' section.

bpbond:
Re the timeshift, @fengly20 , that's fine.

Re loops, let me look...

bpbond:
I've pushed a commit that changes all the output of `ssp1`, `ssp2`, etc., to use loops, which as @skim301 noted makes things much shorter and clearer. Note that it runs, but I haven't done any testing.

fengly20:
Thanks! I'll test it out. 

My other chunk `L232.industry` has the same situation in terms of having outputs by scenarios. I'll adapt the same approach to that chunk.

bpbond:
@skim301 Please approve if no other comments. Thanks.

bpbond:
Thanks, Sonny. @fengly20 any more commits coming? If not I'll move to final tests and merge.

fengly20:
I'm doing package check and may add in one or two variable name bindings. I'll let you know once I'm done. @bpbond

fengly20:
Hi @bpbond , the chunk runs good on my end.

bpbond:
So I don't think the xml files are being done fully/correctly here, but am going to merge this and open a separate issue.

Issue #705, opened 2017-09-08
L244.building_det


bpbond:
This is a _very_ big chunk of code.

Aaron, Haewon @d3y419 has this tentatively assigned to you. A big lift; OK?

d3y419:
now that aaron's off commission for a while, i'll need to borrow @russellhz for this monster chunk!

bpbond:
OK @gokuliyer @russellhz ?

gokuliyer:
Fine by me. This will set up Russel for the (monster)^3 in GCAM-USA :)

PR #791, opened 2017-10-23
Russell l244.building det

russellhz:
#705 
I think this is fully completed. Not 100% if I did the batch files correctly


R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+             "L244.Intgains_scalar",
+             "L244.ShellConductance_bld",
+             "L244.Supplysector_bld",
+             "L244.FinalEnergyKeyword_bld",
+             "L244.SubsectorShrwt_bld", # if exists
+             "L244.SubsectorShrwtFllt_bld", # if exists
+             "L244.SubsectorInterp_bld", # if exists
+             "L244.SubsectorInterpTo_bld", # if exists
+             "L244.SubsectorLogit_bld",
+             "L244.FuelPrefElast_bld", # units
```
bpbond: What do these `units` comments mean?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+
+    L244.Satiation_flsp <- write_to_all_regions(A44.gcam_consumer, c("region", "gcam.consumer", "nodeInput", "building.node.input"), # replace with LEVEL2_DATA_NAMES[["BldNodes]]
+                                                GCAM_region_names = GCAM_region_names) %>%
+      # Match in the region class, and use this to then match in the satiation floorspace
+      left_join_error_no_match(A_regions %>% select(region, region.class),
+                               by = "region") %>%
+      left_join_error_no_match(L244.Satiation_flsp_class, by = c("region.class", "gcam.consumer" = "sector")) %>%
+      select(LEVEL2_DATA_NAMES[["Satiation_flsp"]])
+
+    # Satiation adder - Required for shaping the future floorspace growth trajectories in each region
```
bpbond: Should this ("Satiation adder") be defined more clearly?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+    # So instead, we will take floorspace for maximum year
+    Floorspace_timeshift_pass <- L244.Floorspace %>%
+      filter(year == max(L244.Floorspace$year)) %>%
+      select(-year)
+
+    # Match in the per-capita GDP, total floorspace, and population (for calculating per-capita floorspace)
+    L244.SatiationAdder <- L244.Satiation_flsp %>%
+      left_join_keep_first_only(L102.pcgdp_thous90USD_Scen_R_Y %>%
+                                  filter(year == energy.SATIATION_YEAR), by = "region") %>%
+      # timeshift error here
```
bpbond: still?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+      left_join_error_no_match(Floorspace_timeshift_pass, by = c("region", "gcam.consumer", "nodeInput", "building.node.input")) %>%
+      left_join_error_no_match(L101.Pop_thous_R_Yh, by = c("region", "year", "GCAM_region_ID")) %>%
+      mutate(pcFlsp_mm2 = base.building.size / pop_thous,
+             # We now have all of the data required for calculating the satiation adder in each region
+             satiation.adder = round(satiation.level - exp(log(2) * pcGDP_thous90USD / energy.GDP_MID_SATIATION) *
```
bpbond: An explanatory comment would probably be good here--e.g. what's the `log` calculating?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+    # Generic services: read these values to all regions because they're all the same
+    L244.GenericServiceSatiation <- L244.ServiceSatiation_USA %>%
+      filter(building.service.input %in% generic_services) %>%
+      select(-region, -GCAM_region_ID, -year, -value) %>%
+      distinct() %>%
+      write_to_all_regions(LEVEL2_DATA_NAMES[["GenericServiceSatiation"]], GCAM_region_names = GCAM_region_names)
+
+    # This is bad. Need to match in the floorspace into the base service table, divide to calculate the service demand
+    # per unit floorspace in the final calibration year. This (increased slightly) is then the minimum satiation level that needs to be read in.
+    # TODO: fix the bad code in the model. need a flexible building service function
```
bpbond: Is this your TODO, or Page's?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+    L244.BS <- L244.GenericBaseService %>%
+      left_join_error_no_match(L244.Floorspace, by = c(c("region", "gcam.consumer", "nodeInput", "building.node.input"), "year")) %>% # LEVEL2_DATA_NAMES[["BldNodes"]]
```
bpbond: Why nested `c()`?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+      mutate(service.per.flsp = base.service / base.building.size) %>%
+      filter(year == max(BASE_YEARS)) %>%
+      select(c("region", "gcam.consumer", "nodeInput", "building.node.input"), building.service.input, service.per.flsp) # LEVEL2_DATA_NAMES[["BldNodes"]]
```
bpbond: What's the comment signify here?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+      rename(thermal.building.service.input = building.service.input) %>%
+      # Thermal service satiation is modified in each region according to the HDD/CDD ratio to the USA in a given year
+      mutate(variable = if_else(thermal.building.service.input %in% heating_services, "HDD", "CDD")) %>%
+      left_join_error_no_match(L244.HDDCDD_normal_R_Y, by = c("region", "variable")) %>%
+      group_by(gcam.consumer, variable) %>%
+      mutate(satiation.level = round(satiation.level * degree.days / degree.days[region == "USA"], digits = energy.DIGITS_CALOUTPUT)) %>%
+      ungroup()
+
+    # This part here is bad. The service satiation in the final cal year can not be lower than the observed demand, so need to use pmax to set a floor on the quantity
+    # TODO: fix model code.
```
bpbond: Again...should this TODO be an issue? Is it from original d.s.?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1044 @@
...
+
+    if(exists("L244.SubsectorShrwt_bld")) {
+      L244.SubsectorShrwt_bld %>%
+        add_title("Subsector shareweights for building sector") %>%
+        add_units("Unitless") %>%
+        add_comments("A44.subsector_shrwt written to all regions") %>%
+        add_legacy_name("L244.SubsectorShrwt_bld") %>%
+        add_precursors("energy/A44.subsector_shrwt", "common/GCAM_region_names", "L144.end_use_eff")  ->
+        L244.SubsectorShrwt_bld
+    }else{
```
bpbond: Here and throughout, `} else {` please

R/zchunk_batch_building_SSP.xml.R
```diff
@@ -0,0 +1,85 @@
...
+
+    all_data <- list(...)[[1]]
+
+    for (i in SSP_NUMS){
+      # Names of inputs and outputs
+      Satiation_flsp <- paste0("L244.Satiation_flsp_SSP", i)
+      SatiationAdder <- paste0("L244.SatiationAdder_SSP", i)
+      GenericServiceSatiation <- paste0("L244.GenericServiceSatiation_SSP", i)
+      # SSP2 is unique, doesn't have fuelprefelast or deletethermalservice
+      if (i != 2){
```
bpbond: ?

R/zchunk_L244.building_det.R
russellhz: Maybe, but I would need @pkyle to help with that

R/zchunk_L244.building_det.R
russellhz: This is and below are both Page's. Should we keep them?

R/zchunk_batch_building_SSP.xml.R
russellhz: SSP2 uses less files than the other SSPs (because they are all changing from the default)

R/zchunk_L244.building_det.R
bpbond: Let's move them into issues please, referencing file and line numbers.

pkyle: No need to repeat that text exactly in any code or issues; basically what's going on here is we have to jump through a bunch of hoops on the data side to make up for inflexible functional forms in the model. That said, developing new equations describing the buildings sector and re-coding the thing is probably not going to be a priority in the next few years. Dunno if it's worth an issue on github, Jira, or neither.

bpbond: You call I guess @pkyle - but if in doubt, please do capture in an issue. @russellhz - let's delete the TODO text from the code please.

R/zchunk_L244.building_det.R
pkyle: The satiation adder allows the starting (final calibration year) position of any region and sector to be set along the satiation demand function. It was initially intended to be a non-zero, positive "subsistence" requirement (so that as GDP approaches 0, floorspace per capita doesn't do the same). In its current usage, it is calculated so that the mid-point of the floorspace satiation demand function to an exogenous level of per-capita GDP. And it is negative in a number of regions.
The unit is the same as per-capita floorspace - million square meters per person (unconverted from the model's units of billion square meters and thousand persons). So, it has a physical unit but not a physical meaning!

R/constants.R
```diff
@@ -275,8 +275,18 @@ energy.DIGITS_COEFFICIENT <- 7
...
+energy.DIGITS_SATIATION_ADDER <- 9
+energy.DIGITS_HDDCDD <- 0
 energy.DIGITS_CAPACITY_FACTOR <- 2
 
+energy.SATIATION_YEAR <- 2010
+energy.GDP_MID_SATIATION <- 10.5
+energy.FLOOR_TO_SURFACE_RATIO <- 5.5
+
+energy.INTERNALGAINSSCALAR_USA_H <- -930
+energy.INTERNALGAINSSCALAR_USA_C <- 350
```
kvcalvin: could you put an `_` between words here?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+    L244.DemandFunction_serv <- write_to_all_regions(A44.demandFn_serv, LEVEL2_DATA_NAMES[["DemandFunction_serv"]],
+                                                     GCAM_region_names = GCAM_region_names)
+    L244.DemandFunction_flsp <- write_to_all_regions(A44.demandFn_flsp, LEVEL2_DATA_NAMES[["DemandFunction_flsp"]],
+                                                     GCAM_region_names = GCAM_region_names)
+
+    # Floorspace demand satiation
+    # L244.Satiation_flsp: Satiation levels assumed for floorspace
+    L244.Satiation_flsp_class <- A44.satiation_flsp %>%
+      gather(sector, value, resid, comm) %>%
+      mutate(satiation.level = value * CONV_THOUS_BIL) %>%
```
kvcalvin: What unit is this in before and after the conversion?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+    # L244.SatiationAdder: Satiation adders in floorspace demand function
+    # First, prepare socioeconomics tables by adding region names
+    L102.pcgdp_thous90USD_Scen_R_Y <- L102.pcgdp_thous90USD_Scen_R_Y %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID") %>%
+      rename(pcGDP_thous90USD = value)
+    L101.Pop_thous_R_Yh <- L101.Pop_thous_R_Yh %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID") %>%
+      rename(pop_thous = value)
+
+    # In order to pass timeshift, we need floorspace for energy satiation year, which we don't have
```
kvcalvin: What is the energy satiation year? And, why don't we have data? Is the maximum year a reasonable default?

R/zchunk_L244.building_det.R
kvcalvin: Yeah, any sort of information about what is going on here would be helpful.

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+    # Separate thermal and generic services into separate tibbles
+    L244.GenericBaseService <- L244.base_service %>%
+      filter(building.service.input %in% generic_services)
+
+    L244.ThermalBaseService <- L244.base_service %>%
+      filter(building.service.input %in% thermal_services) %>%
+      rename(thermal.building.service.input = building.service.input)
+
+    # L244.HDDCDD: Heating and cooling degree days by scenario
+    L244.all_sres_gcm <- unique(L143.HDDCDD_scen_R_Y[c("SRES", "GCM")]) # These HDD/CDD scenarios are pretty old, should be updated eventually
```
kvcalvin: Maybe open an issue about updating HDD/CDD scenarios

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+
+    L244.HDDCDD_scen_R_Y <- L143.HDDCDD_scen_R_Y %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID")
+
+    # Let's make a climate normal for each region, using a selected interval of years
```
kvcalvin: "climate normal" seems odd. Is this a no climate change file? If so, maybe say that.

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+    L144.flsp_bm2_R_res_Yh <- L144.flsp_bm2_R_res_Yh %>%
+      mutate(gcam.consumer = A44.gcam_consumer$gcam.consumer[grepl("res", A44.gcam_consumer$gcam.consumer)]) %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID")
+
+    L144.flsp_bm2_R_comm_Yh <- L144.flsp_bm2_R_comm_Yh %>%
+      mutate(gcam.consumer = A44.gcam_consumer$gcam.consumer[grepl("com", A44.gcam_consumer$gcam.consumer)]) %>%
+      left_join_error_no_match(GCAM_region_names, by = "GCAM_region_ID")
+
+    L244.flsp_bm2_R <- bind_rows(L144.flsp_bm2_R_res_Yh, L144.flsp_bm2_R_comm_Yh) %>%
+      # Again, used to be energy.SATIATION_YEAR, changed to pass timeshift test
```
kvcalvin: Again, is this a good assumption? How much does this differ across years?

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+      select(-multiplier)
+
+    # Generic services: read these values to all regions because they're all the same
+    L244.GenericServiceSatiation <- L244.ServiceSatiation_USA %>%
+      filter(building.service.input %in% generic_services) %>%
+      select(-region, -GCAM_region_ID, -year, -value) %>%
+      distinct() %>%
+      write_to_all_regions(LEVEL2_DATA_NAMES[["GenericServiceSatiation"]], GCAM_region_names = GCAM_region_names)
+
+    # This is bad. Need to match in the floorspace into the base service table, divide to calculate the service demand
```
kvcalvin: "This is bad."? Does this mean that the calculation is complicated? Or wrong? Maybe clean up the comment and open issues as appropriate.

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+      mutate(satiation.level = round(value[year == max(HISTORICAL_YEARS)] * multiplier / floorspace_bm2, energy.DIGITS_CALOUTPUT)) %>%
+      ungroup()
+
+    # Generic services: read these values to all regions because they're all the same
+    L244.GenericServiceSatiation_SSPs <- L244.ServiceSatiation_USA_SSPs %>%
+      select(LEVEL2_DATA_NAMES[["GenericServiceSatiation"]], SSP, -region) %>%
+      distinct() %>%
+      write_to_all_regions(c(LEVEL2_DATA_NAMES[["GenericServiceSatiation"]], "SSP"), GCAM_region_names = GCAM_region_names)
+
+    # This is bad. Need to match in the floorspace into the base service table, divide to calculate the service demand
```
kvcalvin: Same comment as above.

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+      write_to_all_regions(LEVEL2_DATA_NAMES[["GenericServiceSatiation"]], GCAM_region_names = GCAM_region_names) %>%
+      rename(thermal.building.service.input = building.service.input) %>%
+      # Thermal service satiation is modified in each region according to the HDD/CDD ratio to the USA in a given year
+      mutate(variable = if_else(thermal.building.service.input %in% heating_services, "HDD", "CDD")) %>%
+      left_join_error_no_match(L244.HDDCDD_normal_R_Y, by = c("region", "variable")) %>%
+      group_by(gcam.consumer, variable) %>%
+      mutate(satiation.level = round(satiation.level * degree.days / degree.days[region == "USA"], digits = energy.DIGITS_CALOUTPUT)) %>%
+      ungroup()
+
+    # This part here is bad. The service satiation in the final cal year can not be lower than the observed demand, so need to use pmax to set a floor on the quantity
```
kvcalvin: Same as above.

R/zchunk_L244.building_det.R
```diff
@@ -0,0 +1,1040 @@
...
+      filter(base.service == 0) %>%
+      select(-base.service) %>%
+      mutate(supplysector = building.service.input)
+
+
+
+    # ===================================================
+    # Produce outputs
+    L244.SubregionalShares %>%
+      add_title("Subregional population and income shares") %>%
```
kvcalvin: What are the subregions? And why are the shares all set to 1?

pkyle: This is for disaggregating the buildings sector into components that follow their own gdp and population pathways (implemented as multipliers on the whole-region pop and gdp). E.g., for urban/rural, or for South/North, without having to create new model regions. Its usage pre-dates the capability to have inter-regional trade of secondary goods, so the parameters are likely obsolete now. Still I suspect (and feel free to check me on this) that it doesn't have an assigned default value, or the default is zero, so it is required to be read in to XML.

R/zchunk_L244.building_det.R
pkyle: The text of that comment doesn't need to be copied! What I meant was, having to jump through all of these hoops is bad. These are adjustments to the input data that are required in order to cover for functional forms in the model that are not capable of handling heterogeneity, and would cause the model to crash out.

russellhz: All of the "this is bad" comments are from the old data system (along with some more colorful language that I deleted). I can delete them. Here is what Page said about something similar "basically what's going on here is we have to jump through a bunch of hoops on the data side to make up for inflexible functional forms in the model. That said, developing new equations describing the buildings sector and re-coding the thing is probably not going to be a priority in the next few years. Dunno if it's worth an issue on github, Jira, or neither."

R/zchunk_L244.building_det.R
pkyle: Climate normal = average climatic conditions in historical years. This is calculated from climate model output tho; at the time we didn't have any re-analysis data available, and when we did get that, it didn't connect at all with the climate model outputs. E.g. historical HDD of Colombia was about 1000 in this climate model, and was <10 in the re-analysis data. We never committed the re-analysis data to the core model.

R/zchunk_L244.building_det.R
pkyle: See my comment above--this is just reverse engineering the building satiation equation in the C++ code so that the satiation function (i.e., the fraction of the assumed satiation level as a function of per-capita GDP in any time period) passes through the calibration year's data point, and will hit 50% of the assumed satiation level at energy.GDP_MID_SATIATION (currently $10,500 per capita in 1990$). Obvi this could have been easier if we just read in the energy.GDP_MID_SATIATION instead!

R/zchunk_L244.building_det.R
russellhz: `energy.SATIATION_YEAR` is just a constant. The issue is not having the data during the time shift because `energy.SATIATION_YEAR` is greater than the historical year under time shift conditions. I will clarify the comment.

R/zchunk_L244.building_det.R
russellhz: #803

R/zchunk_L244.building_det.R
kvcalvin: Can we call it "historical" or "average" or "contemporary"? "normal" seems abnormal to me.

pkyle: Uh, obvi this isn't super important but "climate normal" is the technical term. E.g., https://www.ncdc.noaa.gov/data-access/land-based-station-data/land-based-datasets/climate-normals

kvcalvin: Had no idea. I guess it is fine then.

R/zchunk_L244.building_det.R
pkyle: Before: square meters per person. After: million square meters per person. Inconvenient, but it falls out of the model's units for population (thousands) and floorspace (billion m2)

bpbond:

bpbond:
@russellhz Since merging latest `master` this is now throwing errors:

```
Warnings -----------------------------------------------------------------------
1. matches old data system output (@test_oldnew.R#75) - number of columns of result is not a multiple of vector length (arg 1)

2. matches old data system output (@test_oldnew.R#75) - 4 parsing failures.
row # A tibble: 4 x 5 col     row         col               expected actual expected   <int>       <chr>                  <chr>  <chr> actual 1  2380 degree.days no trailing characters     e3 file 2  2424 degree.days no trailing characters     e3 row 3  2584 degree.days no trailing characters     e3 col 4  2628 degree.days no trailing characters     e3 expected # ... with 1 more variables: file <chr>

3. matches old data system output (@test_oldnew.R#75) - number of columns of result is not a multiple of vector length (arg 1)

4. matches old data system output (@test_oldnew.R#75) - 2 parsing failures.
row # A tibble: 2 x 5 col     row         col               expected actual expected   <int>       <chr>                  <chr>  <chr> actual 1  1764 degree.days no trailing characters     e3 file 2  1808 degree.days no trailing characters     e3 row # ... with 1 more variables: file <chr>

Failed -------------------------------------------------------------------------
1. Failure: matches old data system output (@test_oldnew.R#145) ----------------
Rows in x but not y: 1214, 1278, 412, 476. Rows in y but not x: 2628, 2584, 2424, 2380. 
L244.HDDCDD_A2_CCSM3x.csv doesn't match

2. Failure: matches old data system output (@test_oldnew.R#145) ----------------
Rows in x but not y: 469, 405. Rows in y but not x: 1808, 1764. 
L244.HDDCDD_constdd_no_GCM.csv doesn't match
```
If you get a chance, please take a look, otherwise I will try to do so tomorrow. Thanks.

russellhz:
was googling for the issue, seems you and Robert have already encountered it

[https://github.com/tidyverse/readr/issues/645](url)

Basically, I end up rounding the degree days to `energy.DIGITS_HDDCDD`, which happens to be 0. Then 1000 gets converted to 1e3 and 3000 gets converted to 3e3. 

I added `FLAG_PROTECT_FLOAT`, which I think was your solution

bpbond:
🎃

PR #793, opened 2017-10-25
Russell-L201.socioeconomics_USA

russellhz:
#710 

2 of the outputs have the same names as files from a socioeconomics chunk. I changed them by adding `_GCAMUSA` to the names, but I'm not sure if that's correct.


bpbond:
Re names of outputs, that's fine–thanks.

R/zchunk_L201.socioeconomics_USA.R
```diff
@@ -0,0 +1,132 @@
...
+    # NOTE: No model of labor force used; labor force participation set to a constant
+    L201.LaborForceFillout <- tibble(region = states_subregions$state,
+                                     year.fillout = min(MODEL_YEARS),
+                                     laborforce = socioeconomics.DEFAULT_LABORFORCE)
+
+    # LABOR PRODUCTIVITY GROWTH RATE CALCULATION
+    # Labor productivity growth is calculated from the change in per-capita GDP ratio in each time period
+    # Calculate the growth rate in per-capita GDP
+    L201.LaborProductivity_GCAMUSA <- L102.pcgdp_thous90USD_GCAM3_ctry_Y %>%
+      filter(iso == "usa",
```
bpbond: Is this guaranteed to always be lower case? Maybe `tolower(iso)` is safer? Not sure.

russellhz:
Both comments addressed

Issue #797, opened 2017-10-27
Optional User Outputs  for GCAM-USA


kdorheim:
While working on cement and fertilizer GCAM-USA chunks we have run into problems with conditional outputs in the old data system. For example `zchunk_L2321.cement.R` normally doesn't create `L2321.SubsectorShrwt_cement.csv` because of `A321.subsector_shrwt.csv` contains a year column with missing values, but if a user were to make changes to this assumption file then the SubsectorShrwt file would be created an would need to be processed downstream (See #723 for more details on this exact issue). 

For now we are going to strip these conditional outputs from the new data system and address this issue in v2.  

As of now the following outputs are **NOT**  produced in the new data system. 
- `L2322.SubsectorInterpTo_Fert` from `L2322.Fert` affecting `L2322.Fert_USA`
- `L2322.SubsectorShrwt_Fert` from `L2322.Fert` affecting `L2322.Fert_USA`
- `L2321.SubsectorInterpTo_cement` from `L2321.cement` affecting `L2321.cement_USA`
- `L2321.SubsectorShrwt_cement` from `L2321.cement` affecting `L2321.cement_USA`

** please add to this list as other conditional outputs are removed

bpbond:
@abigailsnyder @russellhz Please add to list above as necessary.

bpbond:
Thanks @kdorheim . Right, we need to figure out the best way to do this--right now the 'optional output' mechanism in the new data system is a bit hackish, and it will be good to take an overall look at the problem and possible mechanisms to address...AFTER we get to v1.

abigailsnyder:
- `L226.SubsectorShrwt_en` from `L226.en_distribution.R` affecting `L226.en_distribution_USA.R` 

- ` L226.SubsectorInterpTo_en` from `L226.en_distribution.R` affecting `L226.en_distribution_USA.R` 

- `L222.GlobalTechShutdownProfit_en` from `L222.en_transformation.R` affecting `L222.en_transformation_USA.R` 

- `L222.GlobalTechSCurveProfit_en` from `L222.en_transformation.R` affecting `L222.en_transformation_USA.R` 

- `L222.GlobalTechLifetimeProfit_en` from `L222.en_transformation.R` affecting `L222.en_transformation_USA.R`

Issue #648, opened 2017-08-16
L2233.electricity_water update attributes


swd-turner:
`L2233.electricity_water` to be updated with improved attributes once upstream inputs are available from #517

swd-turner:
Also batch_xml files to be updated and enabled at this point. Note, following files are passed straight from L223 into batch_electricity_water.xml:

"L223.ElecReserve"
"L223.EQUIV_TABLE"
"L223.StubTechCapFactor_elec"
"L223.SubsectorInterp_elec"
"L223.SubsectorInterpTo_elec"
"L223.SubsectorLogit_absolute-cost-logit"
"L223.SubsectorLogit_elec"
"L223.SubsectorLogit_relative-cost-logit"
"L223.SubsectorShrwt_nuc"
"L223.SubsectorShrwt_renew"
"L223.SubsectorShrwtFllt_elec"
"L223.Supplysector_absolute-cost-logit"
"L223.Supplysector_elec"
"L223.Supplysector_relative-cost-logit"
"L223.StubTech_elec"

swd-turner:
Attributes updated and committed to #798

bpbond:
Thanks @swd-turner - closing.

PR #798, opened 2017-10-27
L223.electricity

cwroney:
#517 

For the moment, I removed the two extra lines from A23.globaltech_eff.csv to enable the merge and match the old data system. We should still figure out why those lines (two lines on backup electricity efficiency for PV and CSP) were changed in some version of the old data system (they're in the 4.3 release version I downloaded from Github in June but not the data system used here). Something for when I have more budgeted time on this. 

Also, it's throwing warnings about the temp-data-inject of these outputs into the GCAM_USA and Water electricity chunks. Not sure if I should be going in and adjusting those, but I didn't want to go over time.


bpbond:
I have made some minor tweaks and some bigger changes: in particular, both this chunk and `LA118.hydro` were leaving `year` as character, not numeric, in many data objects, which broke the downstream `L2233.electricity_water` file.

All the `temp-data-inject`s are now removed, but `L2233.electricity_water` is throwing an error because of NULL objects. It may be the `grepl` calls at lines 304-305 aren't working correctly now? Not sure yet. @swd-turner if you have time to look at this please do. Otherwise @cwroney you and I can come back to this next week.

cwroney:
@bpbond Haven't spent any time on this (don't worry), but possibly related to the same conditional output issue over here, https://github.com/JGCRI/gcamdata/issues/797? This chunk is definitely making some empty tibbles.

swd-turner:
@bpbond Ok - I'll look into these downstream errors in `L2233.electricity_water` tomorrow.

swd-turner:
@bpbond you're correct that the `grepl` calls on lines 304-305 are the problem. When the inputs are temp-inject the `get_title` call returns the actual name of the object, which is the string we want to search. But with the new objects fed in from upstream, `get_title` searches the title written by whoever coded the output. The fix is just to change from `get_title` to `get_legacy_name` in these lines. I'll commit this change shortly along with the other updates to the attributes for `L2233.electricity_water` outputs -- see issue #648.

bpbond:
Ah awesome, thank you!

ssmithClimate:
On the first comment regarding the lines for two lines on backup electricity efficiency for PV and CSP), if that's in 4.3, is there a way of documenting that change in the .csv somewhere other than just removing those lines so that change doesn't just mysteriously happen? (Also @skim301 has spent a lot of time looking at renewables lately, he might have some insights.)

bpbond:
Any other comments @ssmithClimate ? If this OK to merge please approve. Thanks.

ssmithClimate:
@bpbond I haven't looked through the code yet, but will try to do this later today.

bpbond:
Thanks Steve! Chris - when you get a chance, please address these comments, and we'll get this merged.

bpbond:
@cwroney This is failing a timeshift test, looks like.

cwroney:
@bpbond The timeshift is passing on my machine. Not sure what's going on with it!

PR #804, opened 2017-10-30
Chris-L222.en_Transformation-Fix

cwroney:
This fixes L222.en_transformation's outputs. The only note is that L222.GlobalTechShrwt_en kept failing the old new test. I've pinpointed it to six rows (the rows in 2035 with CCS level 1 and 2 with values 0.6665) that are absolutely identical in every way. I've even written the dfs out to .csv and used tortoisegit diff as an additional check. The only possible thing I can think of is that it's reading in a different arbitrary number of trailing 0s for the test, as it's the only interpolated value that doesn't get truncated and they are unrounded. To force a pass I flagged it with SUM_TEST, and it worked fine.


R/zchunk_L222.en_transformation.R
```diff
@@ -166,12 +168,12 @@ module_energy_L222.en_transformation <- function(command, ...) {
       complete(nesting(supplysector, subsector, technology, minicam.energy.input), year = c(year, BASE_YEARS, FUTURE_YEARS)) %>%
       arrange(supplysector, year) %>%
       group_by(supplysector, subsector, technology, minicam.energy.input) %>%
-      mutate(coefficient = approx_fun(as.numeric(year), coefficient, rule =1)) %>%
+      mutate(coefficient = approx_fun(as.numeric(year), coefficient, rule = 1)) %>%
```
bpbond: Aw, look at you putting in appropriate spacing!  👏

Issue #802, opened 2017-10-28
driver could check that `year` columns are integers


bpbond:
A lot of character `year` columns are floating around and that screw up joins.

PR #820, opened 2017-11-08
Enforce that 'year' columns have to be numeric

bpbond:
This is a frequent source of join problems. Includes test code. Addresses #802 .


bpbond:
Note: @rplzzz asked about whether `year` should be enforced as integer, not just (as here) numeric. I looked into this but to do this right I think we'd want to go back and e.g. change the system-wide constant to `HISTORICAL_YEARS <- 1971L:2010L`. I'm opening issue #826 to follow up on this.

Issue #716, opened 2017-09-08
L222.en_transformation_USA


gokuliyer:

abigailsnyder:
Precursors: @bpbond  15 outputs will have to come in as temp data inject until L222.en_transformation.R is done in the energy system.
One output from L202.Ccoef.R - gcamusa will be temp data inject as well.

bpbond:
I don't see a ` L202.Ccoef` in gcam-usa; do you mean energy? That was just merged yesterday.

abigailsnyder:
Whoops, you're right.

PR #825, opened 2017-11-13
Snyder l222 en transformation usa

abigailsnyder:
#716 

@bpbond The output process for L222.DeleteStubTech_USAen is being finicky. When I try to include add_title and add_legacy_name, I get errors when trying to build the driver that the data from the input L222.StubTech_en (read in, filtered, output to get L222.DeleteStubTech_USAen) can't be overwritten. L222.StubTech_en is the only precursor to L222.DeleteStubTech_USAen and is defnitely an input, but including it for add_precursors causes the driver to error that some precursors for L222.DeleteStubTech_USAen are not inputs. 

So now now, I've swapped L222.DeleteStubTech_USAen for an empty tibble  so that the driver can build and the other outputs can be checked (fine on my machine at least). 
Otherwise this chunk is ready for review.


R/zchunk_L222.en_transformation_USA.R
```diff
@@ -67,210 +75,638 @@ module_gcam.usa_L222.en_transformation_USA_DISABLED <- function(command, ...) {
...
+             market.name = technology) %>%
+      separate(market.name, c("market.name", "trash"), extra = "merge", sep = gcamusa.STATE_SUBSECTOR_DELIMITER) %>%
+      select(-trash) ->
+      L222.TechCoef_USAen
+
+
+    # L222.Production_USArefining: calibrated refinery production in USA (consuming output of states)
+    # Aggregated to the supplysector/subsector/technology level
+    ### I'm pretty sure all this aggregation does is remove the state column, since the state information is
+    ### implicit in the technology column. That is, the aggregation doesn't actually change any values - is
```
bpbond: If that step doesn't do anything, let's remove it.

R/zchunk_L222.en_transformation_USA.R
```diff
@@ -67,210 +75,638 @@ module_gcam.usa_L222.en_transformation_USA_DISABLED <- function(command, ...) {
...
+               minicam.energy.input = subsector) %>%
+        filter((subsector == "oil refining" & state %in% oil_refining_states) |
+                 subsector != "oil refining") %>%
+        # Aggregate
+        group_by(region, supplysector, subsector, technology, minicam.energy.input, year) %>%
+        summarise(calOutputValue = sum(calOutputValue)) %>%
+        ungroup %>%
+        mutate(share.weight.year = year) %>%
+        set_subsector_shrwt %>%
+        mutate(tech.share.weight = if_else(calOutputValue == 0, 0, 1)) %>%
```
bpbond: `= abs(sign(calOutputValue))` probably faster although I guess not as clear.

R/zchunk_L222.en_transformation_USA.R
```diff
@@ -67,210 +75,638 @@ module_gcam.usa_L222.en_transformation_USA_DISABLED <- function(command, ...) {
...
-    # There's also a `same_precursors_as(x)` you can use
-    # If no precursors (very rare) don't call `add_precursor` at all
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+    # L222.DeleteStubTech_USAen %>%
+      tibble() %>%
+      add_title("Removes existing stub technologies in the USA region") %>%
```
bpbond: Odd title--can it be clearer? E.g., is this a table of technologies that GCAM will delete in the USA?

bpbond:
@abigailsnyder I've made some minor syntax tweaks and fixed the `L222.DeleteStubTech_USAen` issue–basically, we need to create an entirely new data frame, not just `filter`. Doing so strips off the object attributes and allows us to then call `set_title` etc.

HOWEVER now the chunk is erroring in an unrelated spot. Did I break it? Ugh, sorry if so. Ping me to fix if necessary!

abigailsnyder:
@bpbond let me test it on my machine first. I feel like this keeps happening - I run it, I test it, I check it, everything passes on my machine but then the PR fails. So I'll check if I can get errors locally.

abigailsnyder:
@bpbond ah yes it is at least one of your syntax tweaks. Fixing those now, I'll re run and test everything

bpbond:
My bad. Sorry!

abigailsnyder:
@bpbond I fixed the syntax tweaks and some column labels for one of the outputs. 
I'm a little stuck now. The output `L222.GlobalTechShrwt_en_USA.csv` fails oldnew tests on 6 rows. But I cannot for the life of me identify them. 
The following code throws no errors and has differences between old and new on the order of 1e-16; I can switch old and new in the left_join pipeline with no errors, so I'm interpreting that as all rows being the same. 

```
# L222.GlobalTechShrwt_en_USA.csv
old <- tibble::as_tibble(read.csv("C:/Users/snyd535/Documents/GitHub/gcamdata/tests/testthat/comparison_data/gcam-usa/L222.GlobalTechShrwt_en_USA.csv"))


new <- tibble::as_tibble(read.csv("C:/Users/snyd535/Documents/GitHub/gcamdata/outputs/L222.GlobalTechShrwt_en_USA.csv", comment.char = "#"))

new %>% 
  left_join_error_no_match(old, by = c("sector.name", "subsector.name", "technology", "year")) ->
  both


max(both$share.weight.x - both$share.weight.y)
min(both$share.weight.x - both$share.weight.y)
```

bpbond:
Thanks! I'll take a look later today.

bpbond:
Used `FLAG_SUM_TEST` on `L222.GlobalTechShrwt_en_USA`, as the differences are tiny:
```R
> summary(comb)
 sector.name        subsector.name      technology             year     
 Length:308         Length:308         Length:308         Min.   :1975  
 Class :character   Class :character   Class :character   1st Qu.:2020  
 Mode  :character   Mode  :character   Mode  :character   Median :2048  
                                                          Mean   :2046  
                                                          3rd Qu.:2075  
                                                          Max.   :2100  
 share.weight.old  share.weight         diff           
 Min.   :0.0000   Min.   :0.0000   Min.   :-3.331e-16  
 1st Qu.:0.6665   1st Qu.:0.6665   1st Qu.: 0.000e+00  
 Median :1.0000   Median :1.0000   Median : 0.000e+00  
 Mean   :0.7745   Mean   :0.7745   Mean   : 3.920e-18  
 3rd Qu.:1.0000   3rd Qu.:1.0000   3rd Qu.: 0.000e+00  
 Max.   :1.0000   Max.   :1.0000   Max.   : 3.331e-16  
```

R/zchunk_L222.en_transformation_USA.R
abigailsnyder: I'm tempted to leave it as is in case we do something more dynamic in the future?

abigailsnyder: it's also possible that that is a mistake and not the intended behavior

bpbond:
This looks good. Holding on @gokuliyer review.

R/zchunk_L222.en_transformation_USA.R
```diff
@@ -67,210 +75,575 @@ module_gcam.usa_L222.en_transformation_USA_DISABLED <- function(command, ...) {
     L222.GlobalTechCost_en <- get_data(all_data, "L222.GlobalTechCost_en")
     L222.GlobalTechShrwt_en <- get_data(all_data, "L222.GlobalTechShrwt_en")
     L222.GlobalTechCapture_en <- get_data(all_data, "L222.GlobalTechCapture_en")
-    L222.GlobalTechShutdownProfit_en <- get_data(all_data, "L222.GlobalTechShutdownProfit_en")
-    L222.GlobalTechShutdown_en <- get_data(all_data, "L222.GlobalTechShutdown_en")
-    L222.GlobalTechSCurveProfit_en <- get_data(all_data, "L222.GlobalTechSCurveProfit_en")
+    # L222.GlobalTechShutdownProfit_en <- get_data(all_data, "L222.GlobalTechShutdownProfit_en")
```
gokuliyer: Why are we commenting these out?

abigailsnyder: I don't really understand the details but it is related to this
https://github.com/JGCRI/gcamdata/issues/797

abigailsnyder: Essentially, they're either inputs that aren't created with the current regional markets settings, or they're created but just completely empty

gokuliyer: Got it. Thanks.

Issue #770, opened 2017-09-27
LA1321.Cement


bpbond:
Per @d3y419 . Chris, we know you have a bunch backed up, DSR-wise, so please keep us posted about feasibility etc.

cwroney:
@bpbond It should be fine, I'm hopefully turning in L222 today (just getting final descriptions, etc., sorted) and then this should be next for me after L223.

cwroney:
@bpbond I think the zchunk for this is missing from master?

bpbond:
Hmm. Not only is the file there, but it seems to have been done by Aaron six months ago!

https://github.com/JGCRI/gcamdata/blob/master/R/zchunk_LA1321.Cement-gcam-usa.R
https://github.com/JGCRI/gcamdata/pull/420

Record-keeping fail.

bpbond:
Right? Closing.

cwroney:
@bpbond isn't that GCAM_USA not the core?

bpbond:
Sigh. Yeah. Sorry.

cwroney:
so close! thanks.

PR #828, opened 2017-11-15
Chris-LA1321.cement

cwroney:
No concerns on my end from this one. Fixed the temp-data-injects as well, so you'll changes in a few code chunks.

#770


R/zchunk_LA1321.cement.R
```diff
@@ -1,24 +1,26 @@
...
 #' @param ... other optional parameters, depending on command
 #' @return Depends on \code{command}: either a vector of required inputs,
 #' a vector of output names, or (if \code{command} is "MAKE") all
 #' the generated outputs: \code{L1321.out_Mt_R_cement_Yh}, \code{L1321.IO_GJkg_R_cement_F_Yh}, \code{L1321.in_EJ_R_cement_F_Y}, \code{L1321.in_EJ_R_indenergy_F_Yh}. The corresponding file in the
 #' original data system was \code{LA1321.cement.R} (energy level1).
-#' @details Describe in detail what this chunk does.
+#' @details This chunk generates input, output, and IO coefficients for the cement sector. It begins by downscaling Worrell regional data from 1994
+#' to set up process emissions factors that are multiplied by country emissions from CDIAC to determine production. Limestone consumption is calculated from the same downscaling.
+#' IEA fuelshares and heat and electricity are used to determine energy use by fuel. Energy inputs are then subtracted from industrial energy use and any resulting negative values
+#' are dealt with by moving their accounting to the cement sector.
```
bpbond: 👏 Nice commenting

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # 2. Perform computations
+
+    # Set constants used for this chunk
+    # ---------------------------------
+
+    # Extract carbon coefficient for limestone from assumption file
+    LIMESTONE_CCOEF <- A_PrimaryFuelCCoef$PrimaryFuelCO2Coef[A_PrimaryFuelCCoef$PrimaryFuelCO2Coef.name == "limestone"]
+    # Determine historical years not available in data set (additional years) to be copy values from final available year (final_CO2_year)
+    ADDITIONAL_YEARS <- HISTORICAL_YEARS[!HISTORICAL_YEARS %in% energy.CDIAC_CO2_HISTORICAL_YEARS]
+    FINAL_CO2_YEAR <- energy.CDIAC_CO2_HISTORICAL_YEARS[length(energy.CDIAC_CO2_HISTORICAL_YEARS)]
```
bpbond: Could also do `dplyr::last(energy.CDIAC_CO2_HISTORICAL_YEARS)`

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # Sets maximum for IO coefficient calculated below
+    MAX_IOELEC <- 4
+
+    # =======================================================================================
+    # Derivation of cement production and limestone consumption by region and historical year
+    # =======================================================================================
+
+    # Downscale Worrell's process CO2 emissions and cement production in 1994 to country level using CDIAC emissions inventory
```
bpbond: Who is Worrell? Citation?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # ------------------------------------------------------------------------------------------------------------------------
+    cement_regions %>%
+      left_join(L100.CDIAC_CO2_ctry_hist, by = "iso") %>%
+      filter(year == 1994) %>%
```
bpbond: ?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+
+    # Select country-level emissions to prepare for scaling
+    Worrell_1994_cement %>%
+      select(Country, cement_prod_Mt, process_emissions_MtC) ->
+      Worrell_1994_cement_slim
+
+    # Compute country-level shares of regional emissions and multiply the region-level process emissions
+    # and cement production by these shares to get the country-level estimates
+    L1321.Cement_Worrell_ctry %>%
+      left_join(L1321.Cement_Worrell_reg, by = "Worrell_region") %>%
```
bpbond: Can't use l.j.e.n.m.?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+      L1321.IO_Cement_R_limestone_Yh
+
+    # Derive energy inputs to cement production by region and historical year -
+    # Calculate average electric and TPE intensity for each GCAM region (use process emissions as a weighting factor)
+    # ---------------------------------------------------------------------------------------------------------------
+
+    # Interpolate available data on electricity intensity to all historical years
+    IEA_cement_elec_kwht %>%
+      gather(year, value, matches(YEAR_PATTERN)) %>%
+      mutate(year = as.integer(year)) %>%
```
bpbond: Use new `gather_years` function

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+      arrange(Country, year) %>%
+      group_by(Country) %>%
+      mutate(value = approx_fun(year, value, rule = 1) * CONV_KWH_GJ / CONV_T_KG) %>%
+      ungroup() ->
+      L1321.IEA_cement_elec_intensity
+
+    # Interpolate available data on total primary energy intensity to all historical years by region
+    IEA_cement_TPE_GJt %>%
+      gather(year, value, matches(YEAR_PATTERN)) %>%
+      mutate(year = as.integer(year)) %>%
```
bpbond: Ditto

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # Calculate average regional input energy for electricity across all fuels
+    L123.in_EJ_R_elec_F_Yh %>%
+      group_by(GCAM_region_ID, year) %>%
+      summarise(in.value = sum(value)) %>%
+      ungroup() ->
+      L1321.in_EJ_R_elec_Yh
+
+    # Calculate average regional output energy for electricity across all fuels matching input, join to input energy, and calculate the IO coefficient
+    L123.out_EJ_R_elec_F_Yh %>%
+      semi_join(L123.in_EJ_R_elec_F_Yh, by = "fuel") %>%
```
bpbond: A comment might be helpful here--what's this `semi_join` doing, and why?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # ---------------------------------------------------------------------------------------------------
+
+    # Compile regional historical data on energy use for cement by fuel
+    L1321.Cement_ALL_R_Yh %>%
+      select(GCAM_region_ID, year, elec_EJ, Coal_EJ, Oil_EJ, Gas_EJ, Biomass_EJ) %>%
+      gather(fuel, value, c(elec_EJ, Coal_EJ, Oil_EJ, Gas_EJ, Biomass_EJ)) %>%
+      # Match fuel names to GCAM, removing _EJ and replacing oil and elec
+      mutate(sector = "cement", fuel = tolower(gsub("_EJ", "", fuel))) %>%
+      mutate(fuel = gsub("elec", "electricity", fuel)) %>%
+      mutate(fuel = gsub("oil", "refined liquids", fuel)) ->
```
bpbond: Combine into a single `mutate` call please

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+      bind_rows(L1321.IO_Cement_R_limestone_Yh) ->
+      L1321.IO_GJkg_R_cement_F_Yh
+
+    # Calculate input energy for cement production by region, fuel, and year for L1321.in_EJ_R_cement_F_Y
+    # ---------------------------------------------------------------------------------------------------
+
+    # Compile regional historical data on energy use for cement by fuel
+    L1321.Cement_ALL_R_Yh %>%
+      select(GCAM_region_ID, year, elec_EJ, Coal_EJ, Oil_EJ, Gas_EJ, Biomass_EJ) %>%
+      gather(fuel, value, c(elec_EJ, Coal_EJ, Oil_EJ, Gas_EJ, Biomass_EJ)) %>%
```
bpbond: No need for `c(`

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+      # and aggregate those negative values (in case one region/year has multiple fuels) as a positive adjustment to biomass
+      L1321.in_EJ_R_indenergy_F_Yh %>%
+        filter(value < 0) %>%
+        mutate(sector = "cement") ->
+        L1321.cement_adj_neg
+
+      L1321.cement_adj_neg %>%
+        mutate(fuel = "biomass") %>%
+        group_by(GCAM_region_ID, sector, fuel, year) %>%
+        summarise(value = sum(value) * -1) %>%
```
bpbond: This seems like a bunch of work that might be better accomplished with a single pipeline that computes `abs(sum(value))`? Not sure, but consider if there's a way to streamline, eliminating the `if` above

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
       L1321.out_Mt_R_cement_Yh
 
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
+    L1321.IO_GJkg_R_cement_F_Yh %>%
+      add_title("Input-output coefficients for cement production") %>%
+      add_units("GJ/kg cement") %>%
       add_comments("comments describing how data generated") %>%
       add_comments("can be multiple lines") %>%
```
bpbond: Remove

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    L1321.out_Mt_R_cement_Yh %>%
+      add_title("Historical cement outputs by region, fuel, and year") %>%
+      add_units("Mt cement") %>%
+      add_comments("outputs are calculated by by downscaling Worrell regions using CDIAC country emissions and then aggregating to GCAM regions") %>%
+      add_comments("final outputs are a product of regional emissions times the production emissions ratio") %>%
```
bpbond: Picky Ben comes out: please capitalize first word of comments.

bpbond:
Thanks for handling the `temp-data-inject`s @cwroney . FYI I believe this has the honor of being the VERY LAST energy chunk!

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    # Set constants used for this chunk
+    # ---------------------------------
+
+    # Extract carbon coefficient for limestone from assumption file
+    LIMESTONE_CCOEF <- A_PrimaryFuelCCoef$PrimaryFuelCO2Coef[A_PrimaryFuelCCoef$PrimaryFuelCO2Coef.name == "limestone"]
+    # Determine historical years not available in data set (additional years) to be copy values from final available year (final_CO2_year)
+    ADDITIONAL_YEARS <- HISTORICAL_YEARS[!HISTORICAL_YEARS %in% energy.CDIAC_CO2_HISTORICAL_YEARS]
+    FINAL_CO2_YEAR <- energy.CDIAC_CO2_HISTORICAL_YEARS[length(energy.CDIAC_CO2_HISTORICAL_YEARS)]
+    # Sets maximum for IO coefficient calculated below
+    MAX_IOELEC <- 4
```
kvcalvin: Should this be moved to the constants file?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+    L1321.Cement_Worrell_ctry %>%
+      group_by(Worrell_region) %>%
+      summarise(reg_process_emissions = sum(process_emissions_ktC)) %>%
+      ungroup() ->
+      L1321.Cement_Worrell_reg
+
+    # Select country-level emissions to prepare for scaling
+    Worrell_1994_cement %>%
+      select(Country, cement_prod_Mt, process_emissions_MtC) ->
+      Worrell_1994_cement_slim
```
kvcalvin: Couldn't you just do the select when you are joining it later?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+      complete(nesting(GCAM_region_ID), year = c(year, HISTORICAL_YEARS)) %>%
+      arrange(GCAM_region_ID, year) %>%
+      group_by(GCAM_region_ID) %>%
+      mutate(value = approx_fun(year, value)) %>%
+      filter(year %in% HISTORICAL_YEARS) %>%
+      select(GCAM_region_ID, year, value) %>%
+      ungroup() ->
+      L1321.IO_R_elec_Yh
+
+    # Set cap on IO coefficients for regions and years exceeding maximum value - NOTE: Not sure why we have this cap? Worth revisiting.
```
kvcalvin: Can you open an issue? Also, is it binding in a lot of regions or just 1 or 2?

R/zchunk_LA1321.cement.R
```diff
@@ -52,72 +63,369 @@ module_energy_LA1321.cement_DISABLED <- function(command, ...) {
...
+
+    # Replace NA values in sectors with no match in cement with original values from L132.in_EJ_R_indenergy_F_Yh
+    L1321.in_EJ_R_indenergy_F_Yh_NAs %>%
+      filter(is.na(value)) %>%
+      select(-value) %>%
+      left_join(select(L132.in_EJ_R_indenergy_F_Yh, -sector), by = c("GCAM_region_ID", "fuel", "year")) %>%
+      bind_rows(filter(L1321.in_EJ_R_indenergy_F_Yh_NAs, !is.na(value))) ->
+      L1321.in_EJ_R_indenergy_F_Yh_negbio
+
+    # This dataset may now have negative values. If it's biomass, doesn't matter for energy balances; set the rest-of-industry to exogenous minimum value
```
kvcalvin: Why not?

R/zchunk_LA1321.cement.R
cwroney: Worrell data is only for 1994, so those ratios are assumed to be constant for other years. You have to filter CDIAC to the 1994 values to match it. It's not great, but it's the way it's set up.

R/zchunk_LA1321.cement.R
cwroney: makes the left_join a bit messier but yea, went for it.

R/zchunk_LA1321.cement.R
bpbond: Can you add an explanatory comment about this then?

cwroney: yup, already done! (this time without clicking the wrong button and triggering a review)

R/zchunk_LA1321.cement.R
cwroney: @bpbond @kvcalvin I didnt move additional or final CO2 year because they're dependent on a time constant that would change under timeshift conditions. I was curious so I looked at some of the other sector constants and they do call time constants before they would be shifted, e.g. `energy.CDIAC_CO2_HISTORICAL_YEARS <- HISTORICAL_YEARS[HISTORICAL_YEARS < 2010]` and `emissions.MODEL_BASE_YEARS <- BASE_YEARS[BASE_YEARS < 2008]`. That would put unshifted values in those constants, right? Should I move the commented out timeshift code up to directly below the Time Constants section, so it would update these?

bpbond: >Should I move the commented out timeshift code up to directly below the Time Constants section, so it would update these?

Oh jeez good catch! But NO...let's not make trouble at this stage in the game. Leave the commented out timeshift code alone for now please.

bpbond: I started to open an issue for this, but then decided against it. We've always known the timeshift test was highly imperfect; it's meant to be a way to identify fragile code, but it's not exhaustive.

R/zchunk_LA1321.cement.R
cwroney: I don't think you can use a single pipeline here because it's adding and subtracting on only some rows and at different levels of aggregation of fuel. You might be able to use `abs(sum(value))`, `left_join` and then `mutate` but you still need to add the summed fossil fuels onto biomass beforehand using a couple `if_else` in lieu of separate frames. I spent like 20 min with it and didn't find anything more elegant (easily could be wrong though!). 

I think you're right the `if` isn't needed though, I tested the false case and it just binds empty data frames back to the original and spits it back out.

R/zchunk_LA1321.cement.R
cwroney: Sounds good, I'll leave it well enough alone for now then.

R/zchunk_LA1321.cement.R
cwroney: yeah, it affects 116 out of 1280 entries. There's only a few that occur after the year 2000 and most are before 1990. I'll open an issue and leave it alone for now.

cwroney:
I believe I've addressed everything, and I opened an issue on the Max_IOElec.

bpbond:
Thanks for handling those temp-data-injects, @cwroney .

PR #832, opened 2017-11-16
Snyder finish protected lands

### LINKED ISSUE #730: put protected land code in their own chunk after land input 5
collect protected land steps from land input 2, 3, 4, 5 and put all in one chunk

abigailsnyder:
I had to add `FLAG_PROTECT_FLOAT`, otherwise the oldnew test complained about a single row disagreeing in each output (different sector/region every time, all differences around 1e-13).

I also labeled the files as `L2331.` ; in the old DS, they had names `L233` so I don't know if that will cause errors

Fixes #730


abigailsnyder:
@bpbond it only ended up taking about 1.5 hrs (working on other stuff while driver and tests run), so I've got 6 hours of dsr still to take whatevs on land.

R/zchunk_L2231.land_input_3_irr.R
```diff
@@ -2,7 +2,11 @@
 #'
 #' Produce L2231.LN3_Logit, L2231.LN3_HistUnmgdAllocation, L2231.LN3_UnmgdAllocation,
 #' L2231.LN3_HistMgdAllocation_noncrop, L2231.LN3_MgdAllocation_noncrop, L2231.LN3_UnmgdCarbon,
-#'  L2231.LN3_MgdCarbon_noncrop, L2231.NodeEquiv, L2231.LN3_NoEmissCarbon, L2231.LN3_NodeCarbon.
+#' L2231.LN3_MgdCarbon_noncrop, L2231.NodeEquiv, L2231.LN3_NoEmissCarbon, L2231.LN3_NodeCarbo, and
```
kvcalvin: `L2231.LN3_NodeCarbon` ?

Issue #715, opened 2017-09-08
L210.resources_USA


gokuliyer:

bpbond:
Hi @gokuliyer - FYI the aglu team has hit a few snags (see #730 and #729 ) and @abigailsnyder will be needed to help with dealing with them over the next couple weeks, so she won't be free soon to deal with the three `gcam-usa` issues you assigned her last week.

abigailsnyder:
@bpbond Looks like Kate can finish the aglu, so if you want to dump the files for all my usa chunks, I'll start cranking on them later this week.

bpbond:
Yes ma'am.

abigailsnyder:
@bpbond this chunk requires 
`us_state_wind <- readdata( "GCAMUSA_GIS_DATA", "us_state_wind" )`
`../gcam-usa-data/GIS//us_state_wind.csv`
whiiiich does not exist in either the old or the new ds. any idea what happened to the GIS folder?

abigailsnyder:
I have it from old installations of GCAM but I don't want to bring it in without clearing first?

bpbond:
Agreed. Nope, no idea. @pkyle any insight? Is the folder just too big and so not normally included with the (current) data system checkout?

abigailsnyder:
actually, 9 of the inputs for this chunk will have to come in as temp data inject. Seems like there's a lot of room for frustrating errors when you switch out 9 temp data injects. So I'm going to look through the other chunks I have to see if any of those are a little more ready to go.

Precursor chunks: 
LA115.RooftopPV.R - gcamusa
L210.resources.R - energy

bpbond:
Yeah, good call.

>LA115.RooftopPV.R - gcamusa

This one is an open PR right now.

pkyle:
There should be 5 files that used to be in gcam-usa-data/GIS, and I believe all will be necessary for replication of all files generated by the gcam-usa module of the data system. Should be 4 HDD/CDD ones (used by LA143.HDDCDD), and this us_state_wind.csv (used by L210.resources_USA, and LA114.Wind). Note that two of the HDD/CDD files were not called explicitly by LA143 (e.g., CDD_GFDL_A2); instead, they were read in from a list.files() call. The reason is that if someone were to generate more HDD/CDD scenarios, they could just insert them into that folder, name them accordingly, and the same code would produce an XML input file for each scenario.

bpbond:
[...]

PR #835, opened 2017-11-18
Russell-L210.resources_USA

russellhz:
#715


bpbond:
Holding on @gokuliyer review.

PR #844, opened 2017-11-21
LA143_hddcdd

kdorheim:
@bpbond this was a bit of a rush job, I will be out of town but can make the edits when I get back 11/29


R/zchunk_LA143.HDDCDD-gcam-usa.R
```diff
@@ -1,121 +1,354 @@
...
+#' Estimate heating and cooling degree days for gcam-usa
 #'
 #' @param command API command to execute
 #' @param ... other optional parameters, depending on command
 #' @return Depends on \code{command}: either a vector of required inputs,
 #' a vector of output names, or (if \code{command} is "MAKE") all
 #' the generated outputs: \code{L143.share_state_Pop_CDD_sR9}, \code{L143.share_state_Pop_CDD_sR13}, \code{L143.share_state_Pop_HDD_sR9}, \code{L143.share_state_Pop_HDD_sR13}, \code{L143.HDDCDD_scen_state}. The corresponding file in the
 #' original data system was \code{LA143.HDDCDD.R} (gcam-usa level1).
-#' @details Describe in detail what this chunk does.
+#' @details Estimate heating and cooling degree days for gcam-usa
```
bpbond: This is concise...maybe a bit too concise. A few more details would be useful.

bpbond:
I stitched the downstream chunks and an error is being thrown:
```
[1] "module_gcam.usa_LA144.Residential"
 Error: `by` can't contain join column `subregion13` which is missing from RHS 
```
(Will deal with this later.)

bpbond:
Holding on @gokuliyer review.

bpbond:
Thanks @kdorheim

R/zchunk_LA143.HDDCDD-gcam-usa.R
```diff
@@ -33,91 +35,319 @@ module_gcam.usa_LA143.HDDCDD_DISABLED <- function(command, ...) {
...
+    # the population heating degree days.
+    L143.HDD_state %>%
+      rename(degree_day = value) %>%
+      left_join_error_no_match(Census_pop_hist_subregion %>% rename(population = value),
+                               by = c("year", "state", "subregion9", "subregion13")) %>%
+      mutate(value = degree_day * population) %>%
+      select(state, subregion9, subregion13, year, value) ->
+      L143.Pop_HDD_state
+
+    # Aggregate the population cooling and heating degree days by census subregions.
```
gokuliyer: Unclear to me why the aggregation for each subregion needs to be done one by one.

gokuliyer: It would be  helful to add a comment on why this aggregation needs to be done.

tests/testthat/comparison_data/gcam-usa/L143.HDDCDD_scen_state.csv
```diff
@@ -1,6 +1,6 @@
 # ../gcam-usa-data/level1//L143.HDDCDD_scen_state.csv
 # Written by LA143.HDDCDD.R
-# Wed Sep 20 02:00:41 2017
+# Mon Nov 20 14:29:22 2017
 # " Heating and cooling degree days by state and scenario "
```
gokuliyer: We must track down the source of the HDD and CDD data. I am not sure off the top of my head. @pkyle , any thoughts?

gokuliyer: I see that the original files are in the GIS folder within gcam-usa-data. Is that where we plan to keep these files?

kdorheim: @gokuliyer @bpbond @pkyle I put the CDD and HDD input files into the `gcamdata\inst\extdata\gcam-usa` but @gokuliyer is right, in the old data system it was in the `gcam-usa-data\GIS` directory. Where do we want to put it in the new data system.

bpbond: I'd suggest `gcamdata\inst\extdata\gcam-usa\GIS\`

pkyle: Sorry guys, I just noticed I hadn't written back!
The HDD and CDD data in the model were all produced by Yuyu, and we don't have any archive of the raw data, processing code, etc. You can see from the file names that the future scenarios are from the GFDL climate model, A2 scenario; the historical ones are also from a climate model as well (as opposed to historical or re-analysis data), but the fact that they only go up to 2008 means that the dataset was produced in 2011. And I don't know which climate model was used--my guess is CCSM3. Here's an excerpt from some documentation I wrote up in 2012:
"Historical gridded population data for the US are from HYDE (Klein Goldewikj et al. 2011), which has population at the 5’ scale. Global gridded population data for future scenarios are from the IIASA GGI database. These scenarios are available at 0.5 degree latitude by longitude at a 10-year resolution from 2000 to 2100. Historical and future climate data are from three climate models, depending on the scenario being run. These models include CCSM3 (Collins et al. 2006), GISS (Russell et al. 2000), and Hadley (Gordon et al. 2000). These population and climate data are used to generate historical heating and cooling degree days (HDD and CDD), as well as future scenarios."...
"The method for calculating heating and cooling degree days by state using gridded data of temperature and population over time is documented in Zhou et al. (in review); a brief summary follows. First, the monthly temperature data and estimates of the daily standard deviation in each grid cell, derived from analysis of historical climate data, are used to calculate the monthly HDD and CDD for each grid cell. This step is necessary because climate model output for future scenarios is in monthly timesteps, whereas heating and cooling degree days are calculated from daily temperature. Then, these monthly HDD and CDD are aggregated to annual values.Then, the population-weighted HDD and CDD over a given geographical area (in this case, states) are calculated as follows:"
hddcddcalc

bpbond:
@gokuliyer OK to merge?

tests/testthat/comparison_data/gcam-usa/L143.HDDCDD_scen_state.csv
gokuliyer: Thanks @pkyle!

Issue #839, opened 2017-11-18
LA119.Solar


bpbond:
Accidentally skipped chunk that needs to be done.

d3y419:
@fengly20 : one last one?

fengly20:
Sure. @d3y419

PR #846, opened 2017-11-22
LY-LA119.solar

fengly20:
#839 

Passed tests.


R/zchunk_L223.electricity.R
```diff
@@ -61,7 +61,7 @@ module_energy_L223.electricity <- function(command, ...) {
              FILE = "energy/A23.globaltech_co2capture",
              "L114.RsrcCurves_EJ_R_wind",
              "L118.out_EJ_R_elec_hydro_Yfut",
-             FILE = "temp-data-inject/L119.Irradiance_rel_R",
+             "L119.Irradiance_rel_R",
```
bpbond: Thanks for doing this!

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,114 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
+      irradiance.area.R <- irradiance_avg <- irradiance_avg_rel <- iso <-
+      region_GCAM3 <- value <- year <- NULL
+
     # ===================================================
+    # 2. Perform computations
+    # The irradiance data contained other * region categories which we will have to downscale
+    # Seperate the other * region from the ones with valid country codes that do not need additonal
+    # downscaling
+    Smith_irradiance_ctry_kwh %>%
+      filter(iso %in% grep("OTHER", iso, value = T)) %>%
```
bpbond: `TRUE`

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,114 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
+      summarise(value = sum(value)) %>%
+      ungroup %>%
+      inner_join(L119.Other_ctry, "iso") ->
+      L119.LC_bm2_other_ctry # intermediate tibble
+
+    L119.LC_bm2_other_ctry %>%
+      group_by(region_GCAM3) %>%
+      summarise(value = sum(value)) %>%
+      ungroup %>%
+      rename(Area.bm2.R = value) %>%
```
bpbond: Why not `Area.bm2.R = sum(value)` above and remove this line?

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,114 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
+      mutate(Area.share = value / Area.bm2.R) %>%
+      # Now include the irradiance data at the regional level and multiply the share to downscale to the
+      # country
+      inner_join(L119.Irradiance_kwh_otherR, by = c( "region_GCAM3", "iso", "country_name", "GCAM_region_ID")) %>%
+      mutate( irradiance = irradiance.R * Area.share,
+              irradiance.area = irradiance.area.R * Area.share,
+              dni = dni.R * Area.share,
+              dni.area = dni.area.R * Area.share) %>%
+      # Add the downscaled coutries in with the rest of the country irradiance data
+      select(one_of(names(L119.Irradiance_kwh_ctry))) %>%
```
bpbond: Just `select(names(...))` - we've upgraded to dplyr 0.7 exclusively now.

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,114 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
+    L119.Irradiance_kwh_ctry %>%
+      filter(iso == "usa") %>%
+      mutate(irradiance_avg = irradiance / irradiance.area,
+             dni_avg = dni / dni.area) ->
+      L119.Irradiance_kwh_usa # intermediate tibble
+
+    L119.Irradiance_kwh_R %>%
+      mutate(irradiance_avg_rel = irradiance_avg / L119.Irradiance_kwh_usa[["irradiance_avg"]],
+             dni_avg_rel = dni_avg / L119.Irradiance_kwh_usa[["dni_avg"]]) %>%
+      mutate( irradiance_avg_rel = if_else(is.na(irradiance_avg_rel), 0.001, irradiance_avg_rel),
```
bpbond: Combine into a single `mutate` call

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,114 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
-    tibble() %>%
-      add_title("descriptive title of data") %>%
-      add_units("units") %>%
-      add_comments("comments describing how data generated") %>%
-      add_comments("can be multiple lines") %>%
+
+    L119.Irradiance_rel_R %>%
+      add_title("Relative solar irradiance by GCAM region") %>%
+      add_units("Unitless") %>%
+      add_comments("The irradiance and dni for aggregated 'OTHER_*' regions are first downscaled to country level then combined with irradiance and dni for rest countries to aggregate to GCAM regions. ") %>%
```
bpbond: Nice comments!

bpbond:
Holding on @skim301 review.

R/zchunk_LA119.solar.R
```diff
@@ -27,44 +27,113 @@ module_energy_LA119.solar_DISABLED <- function(command, ...) {
...
+    # The irradiance data contained other * region categories which we will have to downscale
+    # Seperate the other * region from the ones with valid country codes that do not need additonal
+    # downscaling
+    Smith_irradiance_ctry_kwh %>%
+      filter(iso %in% grep("OTHER", iso, value = TRUE)) %>%
+      # Rename columns to indicate these are the regional values which will be used later
+      # to downscale to the country
+      mutate(region_GCAM3 = sub("OTHER_", "", iso)) %>%
+      select(-iso) %>%
+      rename(irradiance.R = irradiance,
```
skim301: I'm concerned that the ".R" used throughout may be confused with the R coding language. Could we use "reg" instead?

R/zchunk_LA119.solar.R
```diff
@@ -1,23 +1,23 @@
 #' module_energy_LA119.solar
 #'
-#' Briefly describe what this chunk does.
+#' Computes relative average irradiance and direct normal irradiance by GCAM region for solar sector.
```
skim301: could you add "(dni)" so that the acronym is clear.

R/zchunk_LA119.solar.R
fengly20: Good point! It confused me when I started re-writing the chunk. Using `reg` sounds good to me unless `.R` is also used in other chunks that we want to be consistent with. Do we have a decision? @bpbond

skim301: I think for file names, we need to leave alone. But for local variable names, it should be fine.

bpbond: Agreed--good suggestion @skim301 , thanks. No, I've never seen another ".R" variable in the data system.

bpbond:
Thanks @fengly20 ! Nice job.

PR #849, opened 2017-11-30
Resolve rounding issues that were causing discrepencies in the socio-economic XML

pralitp:
This seems to take care of #620


bpbond:
Was this a mistake in translating from the old data system? Or some numerical quirk introduced by the shift to `dplyr` etc.?

pralitp:
It was a couple of _mistakes_.  In one case we forgot to round to some predefined number of digits as was the case in old data system.

In the other case we were using population to calculate GDP per capita after the population had been rounded instead of before as was the case in the old data system.

PR #853, opened 2017-12-01
Modify the XML comparison to ignore small numerical discrepancies.

rplzzz:
We were comparing numbers by rounding, and that often has the effect
of ignoring small discrepancies, but not always.  We now handle these
situations more carefully.

This should have been included in the last PR, but that got merged before I had a chance to finish this one.


rplzzz:
This update takes care of some of the problems where nearly equal numerical values were getting flagged because they rounded differently.  Now the rounded version is used for determining sort order, but the raw values are compared for equality within a specified tolerance.

rplzzz:
No worries.  The previous one worked.  This one just works a little better.

PR #854, opened 2017-12-20
Switch to using system2 instead of pipe for running the CSVToXML Java

### LINKED ISSUE #102: Get warning messages from CSVToXML converter into R
Currently most of the error checking when converting CSV (or really tabular data) to XML is done in the Java CSVToXML tool.  Unfortunately the way we are calling this tool (using `pipe`) makes it impossible to capture the STDOUT/STDERR where the warning/error messages are printed thus unable to check them in the test cases.

We could get these messages if we switch to `system2` to call the Java tool however there is a steep performance penalty in using that instead of `pipe`.

pralitp:
Switch to using system2 instead of pipe for running the CSVToXML Java process despite the performance hit since pipe wasn't behaving correctly on Windows anyhow.  As a added bonus this also closes #102.

### DIFF
--- a/R/xml.R
+++ b/R/xml.R
@@ -105,7 +105,7 @@ make_run_xml_conversion <- function() {
         dot$mi_header,
         dot$xml_file
       )
-      warning_msgs <- system2("java", args, stdout = T, stderr = T)
+      warning_msgs <- system2("java", args, stdout = TRUE, stderr = TRUE)
       unlink(tmpfn)
 
       # Note warnings and errors will have been combined together which ideally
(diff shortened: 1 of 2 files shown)


R/xml.R
```diff
@@ -75,26 +75,44 @@ make_run_xml_conversion <- function() {
...
         "-Xmx1g", # TODO: memory limits?
         "ModelInterface.ModelGUI2.csvconv.CSVToXMLMain",
-        "-", # Read from STDIN
+        tmpfn, # Read from the temporary file
         dot$mi_header,
         dot$xml_file
       )
-      conv_pipe <- pipe(paste(cmd, collapse=" "), open = "w")
-      on.exit(close(conv_pipe))
+      warning_msgs <- system2("java", args, stdout = T, stderr = T)
```
bpbond: `TRUE` please for consistency w/ rest of code

Issue #858, opened 2018-02-14
Check tidyr 0.8 for use with gcamdata


bpbond:
Check `tidyr` 0.8 for use with gcamdata.

https://twitter.com/hadleywickham/status/963772224225337345

h/t @kdorheim

PR #859, opened 2018-02-20
tidyr_update

kdorheim:
complete from tidyr 0.8.0 passes along observations that are grouped by a variable with a value of NA. This fix removes NA before complete is called so that the resulting data frames pass the old new tests. 

Addresses #858


kdorheim:
@bpbond sorry I thought that my commit early fixed this. Testing a different fix right now.

bpbond:
🎉 Nice work @kdorheim - thank you

PR #873, opened 2018-03-05
partial_constants_clean_up

### LINKED ISSUE #647: Go through constants.R and clean things up
Make sure everything in right section, has module prefix (`energy.xxx`), has good comments. Use git blame to assign issues as necessary.

kdorheim:
Fixes #647


kdorheim:
@bpbond as part of this do you want to rename/move the following? Also the water constants do not have the correct prefixes, when I tried to rename them the driver failed and I was unable to figure out what was causing the problem. 

old name | new name | move to
-- | -- | --
HISTORICAL_YEARS | modeltime.HISTORICAL_YEARS | modeltime?
FUTURE_YEARS | modeltime.FUTURE_YEARS | modeltime?
BASE_YEARS | modeltime.BASE_YEARS | modeltime?
MODEL_YEARS | modeltime.MODEL_YEARS | modeltime?
GCAM_REGION_ID | gcam.GCAM_REGION_ID | no

bpbond:
Thanks @kdorheim ! Re moving those variables to `modeltime`...they're so frequently used throughout the code, i.e. so universal, maybe not. That would add a lot of verbiage.

PR #877, opened 2018-03-09
Modeltime_xml_check

kdorheim:
This change will allow the model time xmls to pass the xml tests.
```
 run-xml-tests.py ../old-xml/modeltime-xml ../new-xml/modeltime
3 files tested.
0 missing output files.
0 file discrepancies.
```
 See #619 for more details. 

gcamdata will still pass package checks but `L200.hector.csv` will *fail* the old new test.

### DIFF
--- a/R/constants.R
+++ b/R/constants.R
@@ -63,7 +63,7 @@ modeltime.MAGICC_DEFAULT_EMISS_FILE   <- "../input/magicc/Historical Emissions/D
 modeltime.MAGICC_C_START_YEAR         <- 1705
 
 # Hector model assumptions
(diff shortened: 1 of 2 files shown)


bpbond:
We should change the comparison file too, right?

kdorheim:
Make the corresponding change in the old data system replace the csv in `gcamdata\tests\testthat\comparison_data`?

bpbond:
I think so. We want tests to pass before merging. This can be considered part of the GCAM5 changes that @pralitp I doing I think.

kdorheim:
This change is actually part of another PR.

PR #884, opened 2018-03-20
Elec water update

skim301:
Fixed electricity water demand xml file issues. Added missing table, removed unnecessary tables, fixed header mismatch, and logit type write out.


bpbond:
@kdorheim Waiting on your approval/comments.

PR #886, opened 2018-03-22
Enable_energy_batch_xml_chunks

kdorheim:
Enable the batch xml chunks for the energy module in preparation for energy version 1 checkoff.


kdorheim:
`  GCAM_DATA_MAP precursors doesn't match.  Rerun generate_package_data to update.`

I think GCAM_DATA_MAP needs to be rebuilt or something.

bpbond:
@kdorheim I don't mind doing this, but feel free to do it yourself (load the package, then `source` the data/generate_package_data` file. This needs to be done anytime the structure (chunks, dependencies, data flow) of the data system changes.

kdorheim:
@bpbond thanks will do in the  future :+1:

PR #894, opened 2018-03-27
XML-generating files as listed in #890.

### LINKED ISSUE #890: Emissions xmls missing code in new data system
Some of the emissions module xmls are not created in the new data system because the new data system is missing the `create_xml` code in the new data system. @bpbond can you add template code for the following? 

#### L2522.ag_MACC_IRR_MGMT.R (old data system code)
* MACC_TC_SSP5_IRR_MGMT.xml
* MACC_TC_SSP2_IRR_MGMT.xml
* MACC_TC_SSP1_IRR_MGMT.xml

<br>

#### L251.en_ssp_nonco2.R (old data system code)
* delete_gdp_control.xml
* ssp15_emissions_factors.xml
* ssp34_emissions_factors.xml
* ssp2_emissions_factors.xml

bpbond:
@kdorheim This is throwing a couple errors for reasons I don't understand at the moment.

First, `L111.RsrcCurves_EJ_R_Ffos.csv` is failing the oldnew test because it's missing 96 `NA`s that exist in the old data system. This doesn't have anything to do with the files I added in this branch, so I'm not clear what's going on, or why `master` is passing.

Second, `module_emissions_batch_ssp15_emissions_factors_xml` is failing the timeshift test, which is bizarre.

I may not have time to dive into this until the afternoon, so feel free to take a look. Thanks.

Closes #890


R/zchunk_batch_ssp15_emissions_factors_xml.R
```diff
@@ -24,6 +24,10 @@ module_emissions_batch_ssp15_emissions_factors_xml <- function(command, ...) {
 
     # ===================================================
 
+    # Rename L251.ssp15_ef column to match the L2 data names
+    # in order to meet the requirements for the xml header names.
+    L251.ssp15_ef <- rename(L251.ssp15_ef, `emiss.coef` = `emiss.coeff`)
```
bpbond: Why does this fix / how does this related to the timeshift text?

kdorheim: Agh it was somewhat of a guess but in short is was a necessary change.

The driver failed to run on my local because the header used in the `add_xml_data` requires an `emiss.coef` instead of an  `emiss.coeff`. I thought that the error might have been misdiagnosed on travis as a time shift error. 

I checked in with @pralitp and he said that I could rename the `emiss.coef` column in the L2 Data Names list but I was afraid of the number of files that has correctly used `emiss.coef`. Otherwise I `NULL` as the third argument in the function assuming that the order of the columns is correct. I don't want to make assumptions when it comes to the xmls.

bpbond: Well, mad props to you. Thanks.

bpbond:
@kdorheim Approve if OK to merge please.

Issue #891, opened 2018-03-26
Energy xmls missing code in new data system 


kdorheim:
Some of the energy module xmls are not created in the new data system because the new data system is missing the create_xml code in the new data system. @bpbond can you add template code for the following?

#### Created by L223.electricity.R (old data system code)
* wind_low.xml
* solar_low.xml
* nuclear_low.xml
* elec_bio_low.xml
* geo_tech_adv.xml
* geo_low.xml
* nuclear_adv.xml
* solar_adv.xml
* wind_adv.xml

#### Created by integration branch?? 
* transportation_UCD_CORE.xml (this xml is not created by the new data system or gcam-data-system-OLD but it is made in GCAM5 data-system)

kdorheim:
The final missing xml was added as part of #951

PR #895, opened 2018-03-27
Files as listed in #891, except `transportation_UCD_CORE.xml`

bpbond:
Note that `transportation_UCD_CORE.xml` may be a weird one. The `module_energy_batch_transportation_UCD_xml` is pretty unique, and it's not clear to me if it is supposed to be generating the CORE file, or if that deserves its own chunk.


bpbond:
@kdorheim Check and approve when ready; see note above. Thanks.

Issue #887, opened 2018-03-23
Missing zchunk_L2233.electricity_water.R outputs 


kdorheim:
It looks like ` zchunk_L2233.electricity_water.R ` is missing some csv files. I want to check to see if these have been intentionally left out of the new data system or if we need to add them. 

The missing csvs fall into two categories, the Logit and non Logit csv. I am pretty sure that the missing Logit csv files are not a cause for concern. 

@pralitp I know we have already talked about this but could you confirm that the missing logit csv files are intentionally missing from the new data system? 

|missing logit csv files                               |
|:--------------------------------------------------|
|L2233.Supplysector_absolute-cost-logit             |
|L2233.Supplysector_relative-cost-logit             |
|L2233.SubsectorLogit_absolute-cost-logit           |
|L2233.SubsectorLogit_relative-cost-logit           |
|L2233.SubsectorLogit_elec                          |
|L2233.Supplysector_relative-cost-logit_elec_cool   |
|L2233.Supplysector_absolute-cost-logit_elec_cool   |
|L2233.SubsectorLogit_relative-cost-logit_elec_cool |
|L2233.SubsectorLogit_absolute-cost-logit_elec_cool |

- [ ] the missing logit csv files are intentional
- [ ] the missing logit csv files are unexpected 
- [x] need to add `L2233.SubsectorLogit_elec.csv`

<br>
<br>

@pralitp  and/or @skim301  Could you indicate if the other missing csvs are intentional or unexpected? 

| other missing csv files                     |
|:-----------------------------------|
|L2233.GlobalTechInterp_elec_cool    |
|L2233.GlobalTechCapFac_elec_cool    |
|L2233.ElecReserve                   |
|L2233.StubTechCapFactor_elec        |
|L2233.SubsectorShrwtFllt_elec       |
|L2233.SubsectorInterp_elec          |
|L2233.SubsectorInterpTo_elec        |
|L2233.SubsectorShrwt_nuc            |
|L2233.SubsectorShrwt_renew          |
|L2233.GlobalIntTechCapFac_elec_cool |

- [ ] these missing files are intentional
- [x] these missing files are unexpected, all have to be added to the new data system

kdorheim:
@pralitp thanks!

skim301:
It doesn't look like absolute cost logit is used for the electricity water, but Page @pkyle would know better if it is still used anywhere. Page can't you add any info on this?

kdorheim:
[...]

PR #902, opened 2018-03-29
Add_missing_L2233.water_electricity_outputs

kdorheim:
This PR partially addresses #887 by adding two new GCAM 5 upstream outputs as inputs to this chunk thereby enabling the chunk to create two new L2233 outputs that will be used to create `electricity_water.xml`.


kdorheim:
Rebuilding the package data to pass  travis.

PR #904, opened 2018-03-30
Use LOGIT_COLUMN_NAME throughout code

bpbond:
It’s better to use this already-defined constant than the “logit.type”
string


pralitp:
@bpbond, @kdorheim I just saw this PR.  Generally it looks good but can we change the constant from `LOGIT_COLUMN_NAME` to `LOGIT_TYPE_COLUMN_NAME`?  There are two column names typically associated with logits, one is `logit` and the other is `logit.type`.  I think using `LOGIT_COLUMN_NAME` to refer to the later is confusing.

bpbond:
@pralitp Yeah, agreed.

PR #910, opened 2018-04-03
Gpk l2233.water elec tech interp

pkyle:
The old data system had an error whereby the share-weights between single-cycle and combined-cycle gas and oil power plants were not interpolating to 1 over time, as was intended, but instead were going to 1 in the first future time period, 2015. As combined cycle plants have lower levelized costs of generation, this causes significant (up to ~15%) drops in electricity prices in regions where most power plants are single cycle and the dominant fuels are gas and oil.
The reason for the issue was that the technology-level share-weight interpolation file, L2233.GlobalTechInterp_elec_cool.csv, had zero rows, because it was created by joining two tables, wherein the by variables included "year", but the "year" column in one of the tables was NA. This pull request changes the NA to a valid model year (within an if-OLD_DATA_SYSTEM_BEHAVIOR block) so that the join works, and the table L2233.GlobalTechInterp_elec_cool.csv is written out with the appropriate number of rows.


R/zchunk_L2233.electricity_water.R
```diff
@@ -418,7 +418,12 @@ module_water_L2233.electricity_water <- function(command, ...) {
       nondataCols <- names(elecTable)[defCols]
       dataCols <- names(elecTable)[!defCols]
       if(!("year" %in% nondataCols)) {
-        elecTable$year <- NA
+        if(OLD_DATA_SYSTEM_BEHAVIOR){ # old data system set year to NA, which caused all rows of 1 table to be dropped
+          elecTable$year <- NA
+        } else { # setting year to any model year allows the join from L2233.TechMapYr.
+          # Note that the year column is not in the nondataCols, and as scuh will be dropped in select().
```
bpbond: "such"

R/zchunk_L2233.electricity_water.R
```diff
@@ -418,7 +418,12 @@ module_water_L2233.electricity_water <- function(command, ...) {
       nondataCols <- names(elecTable)[defCols]
       dataCols <- names(elecTable)[!defCols]
       if(!("year" %in% nondataCols)) {
-        elecTable$year <- NA
+        if(OLD_DATA_SYSTEM_BEHAVIOR){ # old data system set year to NA, which caused all rows of 1 table to be dropped
```
bpbond: `) {`

Issue #912, opened 2018-04-04
Aglu temp inject data still in use


kdorheim:
As part of #814 we are removing the files from `temp-inject-data` dir. Two aglu files `L201.AgYield_bio_grass.csv` and `201.AgYield_bio_tree.csv` still exist in the `temp-data-inject` folder and are not currently saved as csvs by the data system.  Are they supposed to be an outputs of `module_aglu_L2012.ag_For_Past_bio_input_irr_mgmt` or are they being removed from the new data system?

bpbond:
@abigailsnyder might have insight here too.

abigailsnyder:
my guess is that `L201` is now outputting a combined `L201.AgYield_bio.csv` instead of writing two `csv` files for no real reason.

abigailsnyder:
emphasis on guess

abigailsnyder:
or if the `L201` chunk was completely removed, then those are just intermediate steps in the pipeline that we don't need anymore. Probs.

abigailsnyder:
@rynacui  That is a very excellent point

kdorheim:
The data system uses them as inputs to `module_aglu_L205.ag_prodchange_cost_input_DISABLED` (which may or may not be a problem see  #914) and also in `module_aglu_L2052.ag_prodchange_cost_irr_mgmt`.

rynacui:
Interesting...Let me take a look of L2052.

rynacui:
Just talked to @kdorheim. She is going to try using L2012 instead of L201 in the `module_aglu_L2052.ag_prodchange_cost_irr_mgmt`. It is supposed to work, but if it does not, we probably will just generate L201 in `module_aglu_L2012.ag_For_Past_bio_input_irr_mgmt` as outputs.

rynacui:
I agree that's the quickest way. But there is also a small chance that the L201 outputs fail the oldnew tests...since they are never been tested before. Let's see.

kdorheim:
@rynacui oh I just tried it with the `L2012.AgYeild_bio_ref` was that output you were thinking of using?

rynacui:
@kdorheim yes, that's the one. 

I was saying if we instead generate the L201 outputs, just in case they may fail the oldnew test, we'll need to figure things out.

kdorheim:
Closed by #917

PR #917, opened 2018-04-06
Remove temp-data-inject for bio yields

kvcalvin:
Removes the two remaining temp-data-inject files in the aglu system. 

https://github.com/JGCRI/gcamdata/issues/912


kdorheim:
Pretty sure that travis is failing because the package data needs to be rebuilt, I can do that.

kvcalvin:
Thanks! I always forget about that part.

PR #930, opened 2018-04-12
L107 an feedio precision fix

abigailsnyder:
I had to keep the original rounding for SheepGoat Pastoral FeedCrops in GCAM_regions 10 and 29 to be correct.

Since this is the original problem and gives too  many NA values that then get replaced with 100s, I added the gcd5 version of L107.an_FeedIO into the temp-data-inject folder. 

It is then used to take all of those extra 100's and replace them with the gcd5 values. 

I also added the gcd5 version of L107.an_FeedIO into the comparison data. 

In other words, it should pass all the check package etc, but I didn;t actually check because this has already been a lot of hours. I DID check the old new plot (included) so it should be fine? If it's not just ping it back to me, @kdorheim doesn't need to track down any more weirdness
on this.


abigailsnyder:
[L107.an_FeedIO_should pass tests now.pdf](https://github.com/JGCRI/gcamdata/files/1904219/L107.an_FeedIO_should.pass.tests.now.pdf)

kdorheim:
Currently failing because down stream comparison data needs to be changed. I can take care of this. 

Do we care that this relies on data in the `temp-data-inject` dir? Removal of temp-data-inject dependencies is part of the version 1 check off do we want to make an expectation here? @bpbond

tests/testthat/compdata/aglu/L202.StubTech_an.csv
```diff
@@ -1,7 +1,6 @@
-INPUT_TABLE
-Variable ID
-StubTech
-
+#INPUT_TABLE
```
bpbond: You shouldn't need to do this (test code will skip), but whatever.

kdorheim:
I am working on resolving the L202 that is causing PR to fail.

PR #932, opened 2018-04-13
Aglu gcam5 comp

kdorheim:
This PR replaces the comparison data with gcam 5 outputs (expect for the problem L107 output that will be fixed in #930)


bpbond:
@kdorheim Appveyor is failing because it's trying to use R 3.5.0 beta (2018-04-13 r74592) (bad Appveyor!) and `data.table` isn't available yet for this version. You can go ahead and merge this, ignoring the Appveyor failure.

kdorheim:
@bpbond great thanks!

PR #934, opened 2018-04-16
enabled files and rename column names

cahartin:
3 file had an issue with column names.  I've renamed the column names to
generate the xml files.


cahartin:
@kdorheim can you check to see if these now run and i'll keep pushing forward on the rest of the issues on the emissions side.

kdorheim:
@cahartin thanks!

bpbond:
This will need a GCAM_DATA_MAP update; @kdorheim should I do?

kdorheim:
@bpbond yes please! thanks

bpbond:
```
NOTE: chunk module_aglu_L202.an_input reads `temp-data-inject/L107.an_FeedIO_R_C_Sys_Fd_Y`
	but this is available from module_aglu_LA107.an_IMAGE_R_C_Sys_Fd_Y
NOTE: chunk module_aglu_L202.an_input reads `temp-data-inject/L107.an_FeedIO_R_C_Sys_Fd_Y` but this is handled by a data chunk
```

Pretty sure this is intended, for now?

kdorheim:
@bpbond this is the `temp-inject-data` that @abigailsnyder added as part of the old data system behavior block.

PR #939, opened 2018-04-18
Sort GCAM_DATA_MAP before comparing

bpbond:
The order of the GCAM_DATA_MAP generated by the driver can vary: in
particular, when @kdorheim runs `generate_package_data.R` on pic, the
oldnew test will subsequently fail on Travis. Arrange the data to a
common order before comparing.


bpbond:
@kdorheim Can you check this out on pic, rebuild GCAM_DATA_MAP, commit and push? That will test whether these changes address this problem.

Issue #950, opened 2018-04-23
LoadFactor discrepancies in transportation_UCD_CORE.xml


kdorheim:
### Problem: 

` transportation_UCD_CORE.xml` fails the python old new xml test because there are  three LoadFactor values that differ. In the python output below the left value refers to the old xml and the right value refers to values from the dsr xml. 

```
At: [('scenario', ''), ('world', ''), ('region', 'Southeast Asia'), ('supplysector', 'trn_freight_road'), ('tranSubsector', 'Truck (0-6t)'), ('stub-technology', 'Liquids'), ('period', '')]
	Left:
		loadFactor
		
		[]
		[]
		0.52
		
	Right:
		loadFactor
		
		[]
		[]
		0.53
		

At: [('scenario', ''), ('world', ''), ('region', 'Southeast Asia'), ('supplysector', 'trn_freight_road'), ('tranSubsector', 'Truck (0-6t)'), ('stub-technology', 'NG'), ('period', '')]
	Left:
		loadFactor
		
		[]
		[]
		0.52
		
	Right:
		loadFactor
		
		[]
		[]
		0.53
		

At: [('scenario', ''), ('world', ''), ('region', 'Southeast Asia'), ('supplysector', 'trn_freight_road'), ('tranSubsector', 'Truck (0-6t)'), ('stub-technology', 'NG'), ('period', '')]
	Left:
		loadFactor
		
		[]
		[]
		0.52
		
	Right:
		loadFactor
		
		[]
		[]
		0.53
		

At: [('scenario', ''), ('world', ''), ('region', 'Southeast Asia'), ('supplysector', 'trn_freight_road'), ('tranSubsector', 'Truck (0-6t)'), ('stub-technology', 'NG'), ('period', '')]
	Left:
		loadFactor
		
		[]
		[]
		0.52
		
	Right:
		loadFactor
		
		[]
		[]
		0.53
		
```

These values come from `L254.StubTranTechLoadFactor.csv` . When we compare the LoadFactor from dsr and the gcam5 comparison data we see that there are a total of 50 entries in  `Central America and Caribbean`, `South America_Northern`, ` EU-15`, `South America_Southern`, `Southeast Asia`, and `Mexico` have an absolute difference between gcam5 and dsr LoadFacotor greater than `0.01`. 

### Why: 
- The gcam 5 and dsr xmls contain different LoadFactor values because the `L254.StubTranTechLoadFactor` are different from one another. 
[...]

PR #952, opened 2018-04-24
fix transportation loadFactor values

kdorheim:
Add to the old data system behavior block to get the loadFactor values in the transportation_UCD_CORE xml to pass the xml test.


kdorheim:
will close #950

kdorheim:
yup! there was already one in the chunk so I added it there instead of making a new one

PR #961, opened 2018-05-03
Self test outputs and remove unneeded FLAGS

### LINKED ISSUE #457: Remove FLAG_NO_TESTs once we jettison old data system as baseline
E.g. in `L100.GDP_hist` - see #456 .

### LINKED ISSUE #918: Replace comp data with dsr output
The numerical instability of the new data system is causing some dsr outputs to fail the old new tests. When the relative and absolute difference between dsr and gcam 5 data system outputs are small we have replaced the comparison data with dsr output, in order to make progress on the version 1 check off. 

When the numerical instability is addressed (in 1.1) the comparison data for the following will need to be replaced.

kdorheim:
This PR replaces the comparison data with dsr output, and removes the `FLAG_NO_TEST`, `FALG_SUM_TEST` and other unneeded FLAGS. 

Closes #918 
Closes #457


kdorheim:
FYI the dsr xmls match the gcam 5 xmls. 
```
198 files tested.
0 missing output files.
0 file discrepancies.
```

bpbond:
Wow that's right–most of the FLAG_LONG_FORMs can go away now.

PR #966, opened 2018-05-23
Save in GCAM_DATA_MAP an MD5 hash of each chunk data.

pralitp:
We will then use this as a quick check for old/new tests instead of the
full comparison of the data as a compromise to avoid having to version
control the full compdata dataset which is hefty in size and could
pretty quickly cause bloat in the Git repository.

Users can still perform the full old/new test by setting the option
gcamdata.validate_with_compdata however they must manually provide the
data to compare against in compdata.


pralitp:
Note we may want to add a `.gitignore` in the `tests/testhat/compdata` to ignore `*.csv`.  But after we clean out all of the old compdata?

bpbond:
>Note we may want to add a .gitignore in the tests/testhat/compdata to ignore *.csv. But after we clean out all of the old compdata?

Yeah, I'd do that in the same commit as deleting the compdata I think.

tests/testthat/test_oldnew.R
```diff
@@ -13,48 +13,37 @@ test_that("matches old data system output", {
...
-    expect_identical(gdm_internal$name, gcam_data_map$name, info =
-                       "GCAM_DATA_MAP name doesn't match. Rerun generate_package_data to update.")
-    expect_identical(gdm_internal$output, gcam_data_map$output, info = "GCAM_DATA_MAP output doesn't match")
-    expect_identical(gdm_internal$precursors, gcam_data_map$precursors, info =
-                       "GCAM_DATA_MAP precursors doesn't match. Rerun generate_package_data to update.")
-    expect_identical(gdm_internal$md5hash, gcam_data_map$md5hash, info =
-                       "GCAM_DATA_MAP md5hash doesn't match. Rerun generate_package_data to update only if you do not expect outputs to change.")
-  }
+  # Run the driver only to get the gcam data map for comparison
+  gcam_data_map <- driver(quiet = TRUE, return_data_map_only = TRUE)
```
bpbond: Have you tried this locally? It will be slow. That was the advantage of doing it only on Travis.

PR #968, opened 2018-05-31
Split out compdata into its own package

pralitp:
Split out compdata into its own R package and load it in the old/new test as needed.

### DIFF
--- a/tests/testthat/test_oldnew.R
+++ b/tests/testthat/test_oldnew.R
@@ -50,6 +50,11 @@ test_that("matches old data system output", {
                        "GCAM_DATA_MAP precursors doesn't match. Rerun generate_package_data to update.")
   }
 
+  if(!requireNamespace("gcamdata.compdata", quietly = TRUE)) {
+    skip("gcamdata.compdata package not available unable to load COMPDATA")
+  } else {
+  library(gcamdata.compdata)
+  data(COMPDATA)
   # For each file in OUTPUTS_DIR, look for corresponding file in our
   # comparison data. Load them, reshape new data if necessary, compare.
   for(newf in list.files(outputs_dir, full.names = TRUE)) {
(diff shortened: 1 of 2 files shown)


pralitp:
Some clean up to remains.  I had a couple of questions though.  I made the new compdata off of the outputs of gcamdata v1.0.  Thus the flags for X years or no Long years are no longer needed?  Also do we still want to make the old/new tests optional (We would still get failures when people make changes that *should* change the outputs?

tests/testthat/test_oldnew.R
```diff
@@ -50,6 +50,11 @@ test_that("matches old data system output", {
                        "GCAM_DATA_MAP precursors doesn't match. Rerun generate_package_data to update.")
   }
 
+  if(!requireNamespace("gcamdata.compdata", quietly = TRUE)) {
```
bpbond: Why not simply `require` here?

tests/testthat/test_oldnew.R
```diff
@@ -50,6 +50,11 @@ test_that("matches old data system output", {
...
+    skip("gcamdata.compdata package not available unable to load COMPDATA")
```
bpbond: "gcamdata.compdata package not available" probably clearer and shorter

bpbond:
>Thus the flags for X years or no Long years are no longer needed? 

Right!

>Also do we still want to make the old/new tests optional (We would still get failures when people make changes that should change the outputs?

Hmm. It seems like the old/new test should be required if you're doing development work?

pralitp:
Ok, I've cleaned up the commented out code and made your suggested changes.

> Hmm. It seems like the old/new test should be required if you're doing development work?

Yea, maybe I'm jumping the gun on a problem that doesn't exist yet.  If we decide getting "failures" from expected changes is distracting and happens with enough frequency in the future we can put in some way of controlling its then.

PR #975, opened 2018-06-06
OLDNEW-LB132.ag_an_For_Prices_USA_C_2005

russellhz:
Removed old data behavior that used slightly different GDP deflator numbers

Changed files:
1. L132.ag_an_For_Prices
2. L133.ag_Cost_75USDkg_C
3. L164.ag_Cost_75USDkg_C
4. L2012.AgSupplySector
5. L202.GlobalTechCost_an
6. L2052.AgCost_ag_irr_mgmt
7. L2062.AgCost_ag_irr_mgmt_adj
8. L2072.AgCoef_IrrWaterWdraw_ag_mgmt
9. L221.StubTechFractProd_en


bpbond:
@russellhz - have you looked at how much numbers in those files change?

bpbond:
@pralitp So this branch, as well as `master`, are currently throwing all kinds of driver warnings and test errors on my machine, as well as on Travis; it doesn't look like your last commit (739af0849) before merging #971 passed either. Could you please re-visit?

pralitp:
@bpbond, 739af08 was passing all tests with no warnings as far as I can tell.  It did however decrease code coverage.  I can go back and add tests if you like?

With regards to this pull request, if we are changing the data system behavior don't we expect the old / new test to fail?  Is the problem we weren't expecting these 9 files to change with data system change?

bpbond:
Hmm, see https://github.com/JGCRI/gcamdata/commits/master - big red X's on yesterday's commits.

I don't have `gcamdata.compdata` installed, so it shouldn't be doing any comparisons...and the errors are all related to the xml tests. I know you changed the default; but that means tests will fail too? How is anything going to pass on Travis then? Anyway maybe I'm misunderstanding new system. Thanks.

pralitp:
Right. The big red X was due to decreased close coverage but the travis and appveyor tests passed. 

The XML tests have been enabled on Travis for a long time now. 

Do you have Java installed locally (also incompatibility with Java 9/10 is a known issue and I'm waiting toget the Gcam proposal approved to back port the fix to gcamdata). Also you should be able to disable Java locally via options or environment variable if you don't want to install it.

bpbond:
Ah, OK, thanks. And ah, I have java 10 installed. But now I'm not sure why Travis is failing. Ah well, tomorrow.

russellhz:
@bpbond I wrote a script to automatically generate some stats about what's changing. Let me know if there is anything else other than what's below that we might want to see.

It looks like for all of these files, values change by less than 2%

diff_lb132

bpbond:
@russellhz - nice. We may want to make that script a permanent part of this repo!

@kvcalvin - changes above look OK?

bpbond:
Thanks Kate. @russellhz at this point please go and update the gcamdata [comparison data](https://github.com/JGCRI/gcamdata.compdata) - clone it, replace the outputs above with what's generated from this PR, and open a PR there. Pralit or I will approve and merge, and then when we re-run Travis here, things should pass. Make sense? Thanks.

russellhz:
@bpbond @pralitp I'm having some trouble cloning `gcamdata.compdata` 

I installed git-lfs, but things still aren't working. Any ideas?

screen shot 2018-06-07 at 3 38 45 pm

bpbond:
Tag you're it @pralitp

pralitp:
Well, that was not something I was expecting.  Could be a git for windows problem:
https://github.com/git-for-windows/git/issues/1158

@russellhz do you know how old your git install is?  Maybe updating will fix it?

russellhz:
@pralitp following the directions on the 4th comment [here](https://github.com/git-lfs/git-lfs/issues/911), plus removing some certification stuff, allowed me to clone it

pralitp:
Great!  Looks like it might be time for me to add a wiki page.

bpbond:
Hmm, still failing even after `gcamdata.compdata` was updated. I wonder if Travis is caching the older version (as our config file instructs)? I'll look at this but not until tomorrow.

PR #980, opened 2018-06-12
OLDNEW_BEHAVIOR-L104.bcoc_en_USA_S_T_Y

russellhz:
This one causes a bunch of changes.

The fix to match the old-data-system that is removed in `zchunk_L104.bcoc_en_USA_S_T_Y.R` is pretty large. 
Getting rid of this code changes the refined liquids fuel use in:
trn_air: from 8.8 to 0.9
trn_rail: from 17.3 to 3.7
trn_road: from 0 to 33.9
trn_ship: from 0 to 1.4

The downstream effects of this include some fairly large percent changes. All of them are in liquids use for transportation.

Three files lose rows. This is because `max-reduction`  for emissions factors is set in `zchunk_L151.ctrl_R_en_S_T.R`. A bunch of values change to 0 because the calculated emissions factors are now below a threshold. These 0s are then removed.

l104

l104 2crop


bpbond:
Thanks for tagging @kvcalvin - Kate, if there's someone more appropriate, please suggest.

PR #983, opened 2018-06-15
OLDNEW_BEHAVIOR-L254.transportation_UCD

russellhz:
3 files have changed values, all with fairly small percentage differences.

`L254.tranSubsectorShrwt` was not previously created due to an incorrect if statement. It is now created, so has no old file to compare to.

l254

diff_l254


bpbond:
@d3y419 - this PR shifts `L254.transportation_UCD` away from data system behavior, changing some values slightly. Is there anyone in particular who should review this before merging? Thanks.

bpbond:
@mwisepnnl Marshall you reviewed this originally--can you look at these changes please? Thanks.

pkyle:
Have any of you tried running the XML files in GCAM? We need to make sure the model still calibrates/solves the base years, with the revisions to the quantity flows.

bpbond:
@pkyle This v1 data system was checked in this way, but we haven't (to my knowledge) checked these changes, no. That seems like a great idea to do, maybe not with every PR, but maybe every omnibus one? Thoughts @pralitp ?

pralitp:
We will have to check all gcamdata 1.1 XMLs (and summarize their net effect) as a core model proposal  once we have them all so I have been waiting until we get there.

That being said.  As @pkyle points out this directly changes calibration values so it would be useful to double check this one still runs and calibrates the historical years.  So I have re-generated and replaced just the `transportation_UCD_CORE.xml` (here is the XML diff if that is helpful: [trn_xml_changes.txt](https://github.com/JGCRI/gcamdata/files/2134315/trn_xml_changes.txt)) and ran that in GCAM and it still solves and calibrates just fine:

```
Period 1: 1990
Model solved normally. Iterations period 1: 131. Total iterations: 132

Period 2: 2005
Model solved normally. Iterations period 2: 197. Total iterations: 328

Period 3: 2010
Model solved normally. Iterations period 3: 91. Total iterations: 419

Period 4: 2015
Model solved normally. Iterations period 4: 343. Total iterations: 761
```

Issue #830, opened 2017-11-16
Data relationships - kind of a mess, no dedicated layer


bpbond:
Right now functions like `dstrace` and `graph_chunks` and others get the raw inputs-and-outputs data from `chunk_inputs` and `chunk_outputs` and then have to manually construct/figure out relationships via joins, etc. It would be good to abstract and centralize this behind a standard layer of functions.

Just saving some notes on this topic...

High-level functions:

```R
inputs_of(chunk_name)
  # returns character vector of data object names
  # convenience function; calls upstream_data(..., data_store = NULL, max_generations = 1)
outputs_of(chunk_name)
  # returns character vector of data object names
  # convenience function; calls downstream_data(..., data_store = NULL, max_generations = 1)

upstream_data(chunk_name, data_store = NULL, max_generations = Inf)
upstream_data(data_name, data_store = NULL, max_generations = Inf)
  # returns character vector if data_store is NULL
  # returns a named list of objects if not NULL
downstream_data(chunk_name, data_store = NULL, max_generations = Inf)
  # returns character vector if data_store is NULL
  # returns a named list of objects if not NULL
upstream_chunks(chunk_name, max_generations = Inf)
  # returns character vector if data_store is NULL
  # returns a named list of objects if not NULL
downstream_chunks(chunk_name, max_generations = Inf)
  # returns character vector if data_store is NULL
  # returns a named list of objects if not NULL
```

Low-level functions (already exist):

```R
chunk_inputs()
chunk_outputs()
chunk_list()
 ```

pralitp:
Note, #985 piles on to this issue by adding yet another place where we have custom calculation of dependencies.  Namely in the driver around line 230 where we calculate the minimum chunks required to run a `stop_before` or `stop_after`. And then that information is used in a new function called `dstrace_chunks`.

bpbond:
This basically duplicates #757 - closing but will add a note there.

PR #985, opened 2018-06-20
Min path v2

pralitp:
@bpbond, I'm opening this PR to get your input on the approach before I get too far.

I decided not to try to refactor `dstrace` to do the depth first search.  There would have been a few things I would have needed to add/change such as getting chunk inputs/outputs before making the call, being able to trace more than 1 thing at a time, and of course parsing it's output to get the unique chunks names which is what we were really interested in here.

Instead I created a separate function called `dstrace_chunks` which is super simple.  It may not be optimal in terms of performance but the amount of time it takes to run isn't noticeable so I'm not worried about it anymore.

I refactored when we load CSV files (for `unfound_inputs`) instead of all upfront I took a "lazy" approach to get them when needed by adding them in with the `chunks_to_run` list (it's not really lazy because I added them all at the front of the list anyways).  However if we process them like this then we can only load the minimal set of files needed when running the driver with `stop_before` / `stop_after`.

One things left to figure out is `dstrace_chunks` is tailored to what I need right now.  It operates some merged tibble of `chunkoutputs`, `chunkinputs`, and `unfound_inputs`.  Should I make it flexible to also take the `GCAM_DATA_MAP` as well?  Should it only operate on `GCAM_DATA_MAP` (this would imply users have updated it before attempting to use `stop_before`/`stop_after`)?


R/driver.R
```diff
@@ -224,19 +224,28 @@ driver <- function(all_data = empty_data(),
...
   metadata_info <- list()
-  if(return_data_map_only) {
-    if(!quiet) cat("Extracting metadata from inputs...\n")
-    metadata_info <- list(INPUT = tibbelize_outputs(all_data, chunk_name = "INPUT"))
-  }
 
   # Initialize some stuff before we start to run the chunks
-  chunks_to_run <- chunklist$name
+  if(!missing(stop_before) || !missing(stop_after)) {
+    run_chunks <- ifelse(missing(stop_before), stop_after, stop_before)
```
bpbond: `ifelse` in R is bizarre enough I try to avoid it.

R/driver.R
```diff
@@ -224,19 +224,28 @@ driver <- function(all_data = empty_data(),
...
+    # calc min list
+    verts <- inner_join(bind_rows(chunkoutputs,
```
bpbond: See #830 - do we want to abstract this into a helper function, the start of a new data layer?

R/driver.R
```diff
@@ -253,13 +262,24 @@ driver <- function(all_data = empty_data(),
...
-      if(!quiet) print(chunk)
-
       if(chunk %in% stop_before) {
         chunks_to_run <- character(0)
         break
       }
 
+      if(chunk %in% unfound_inputs$input) {
+        unfound_chunk = unfound_inputs[unfound_inputs$input == chunk, ]
+        chunk_data <- load_csv_files(unfound_chunk$input, unfound_chunk$optional, quiet = TRUE)
```
bpbond: I think it makes sense to shift to this lazy approach. Before, when development was so active and files were changing constantly, it was valuable to load and QC all inputs at the beginning. Now, it's not as important.

R/dstrace.R
```diff
@@ -96,6 +96,33 @@ dstrace <- function(object_name, direction = "upstream", graph = FALSE,
...
+#'
+#' The algorithm is performing a depth first search where chunks are "nodes"
+#' and data objects are "verticies" connecting them.
+#' @param chunk_names A list of chunks to find the precursors for.
+#' @param gcam_data_map A tibble of metadata information; normally a built-in package dataset.
+#' @param prev_trace_list The list of chunks already found.
+#' @return The unique list of chunk names which are precursors for \code{chunk_names}.
+#' @export
+#' @importFrom magrittr %$%
+dstrace_chunks <- function(chunk_names, gcam_data_map, prev_trace_list = c()) {
```
bpbond: See comment above about #830

R/utils.R
```diff
@@ -349,27 +349,30 @@ chunk_inputs <- function(chunks = find_chunks()$name) {
   assertthat::assert_that(is.character(chunks))
 
   # Get list of data required by each chunk
-  chunkinputs <- list()
+  names <- character()
```
bpbond: This masks `names`, a very common base R function--would be better to use something else.

pralitp:
Ok, I was just going to summarize what is in the PR and why I made the changes:
* Calculate the minimum number of chunks that are required when running driver with `stop_before` or `stop_after`
* Lazily load `FILE` data from CSV instead of all up front.  This is consistent with the former goal so that we do not load any `FILE` data that wasn't required to  `stop_before` or `stop_after` some chunk.
* Make some optimizations to speed up `chunk_inputs` and `chunk_outputs`.  These functions indirectly lead to a start up time of around 10 seconds which isn't that much time when running the full driver, however if you are only running a few chunks then it is noticeable.

R/driver.R
```diff
@@ -342,11 +367,13 @@ warn_data_injects <- function() {
   # Are any chunks are using temp-data-inject data that are also available to them through the data system?
   ci <- chunk_inputs(find_chunks(include_disabled = FALSE)$name)
   chunk_outputs(find_chunks(include_disabled = FALSE)$name) %>%
-    rename(upstream_chunk = name) ->
+    rename(upstream_chunk = name) %>%
+    mutate(output = output) ->
```
bpbond: ?

pralitp: Good question.. it has been too long for me to remember exactly but I was getting some odd error message something along the lines of join looses tibble meta data or something.

I think I tried this change thinking it could be something like [Common Fails - mutate forces copy](https://github.com/JGCRI/gcamdata/wiki/Common-Fails-and-How-to-Fix-Them#error-in-add_title---not-allowed-to-overwrite-current-title-).

But I think ultimately the fix was the change in the line below.  I removed this change and verified everything still works for me with no odd error messages so I will push that up.

Issue #979, opened 2018-06-11
OLDNEW-L244.building_USA needs updating to A44.demand_satiation_mult as well


pralitp:
It turns out the OLD_DATASYSTEM_BEHAVIOR in L244.building_USA was also masking an error in the assumptions file `gcam-usa/A44.demand_satiation_mult` -- where the multiplier for heating services was set to 1.0.  By definition this value must be > 1.0 and will cause an `abort()` in GCAM when run.

Zarrar and I ran across this when trying to fix this on a development branch we have been working on.  We updated these values to match values @gokuliyer and @pkyle made in their work on GCAM-USA, although I am not sure if ultimately those are the values we wanted to use or not.

pralitp:
FYI, @bpbond and (@russellhz in case you will be working on this version 1.1 fix).

zarrarkhan:
Attaching the updated gcam-usa/A44.demand_satiation_mult file. 
[A44.demand_satiation_mult.csv.zip](https://github.com/JGCRI/gcamdata/files/2091165/A44.demand_satiation_mult.csv.zip)

bpbond:
Thanks you guys. So what's the timing on making this change here in `gcamdata`?

gokuliyer:
@pkyle , I'd leave this one to you.

pkyle:
Not sure why this requires me or anyone else specifically...someone just needs to copy that enclosed file over the existing one, right? Do we need a core model proposal or something?

pralitp:
No, I don't think there is anything to do right now.  Maybe when it comes time to write the proposal we will have to note why we choose the multipliers to be what they were.  We took it from the EPSA branch so I assume we can copy some text from the broader GCAM-USA proposal(s).

pralitp:
This is closed in #986

PR #986, opened 2018-06-20
OLDNEW fixes for L244.building_USA

pralitp:
Using the wrong year for floorspace(first instead of last) when calculating per floorspace building service satiation level.

Note taking care of this exposes another bug (#979) which gets fixed here too.

This effects two data products as we expect:
```
1. Failure: matches old data system output (@test_oldnew.R#115) ----------------
Rows in x but not y: 261, 353, 128, 506, 183, 616, 631, 281, 195, 419, 270[...]. Rows in y but not x: 662, 660, 659, 656, 654, 653, 651, 649, 648, 644, 643[...].
L244.GenericServiceSatiation_gcamusa.csv doesn't match

2. Failure: matches old data system output (@test_oldnew.R#115) ----------------
Rows in x but not y: 204, 203, 202, 200, 199, 198, 197, 196, 195, 194, 192[...]. Rows in y but not x: 204, 202, 201, 199, 198, 196, 194, 193, 192, 190, 189[...].
L244.ThermalServiceSatiation_gcamusa.csv doesn't match
```

Further statistics attached.
[diff_L244.building_USA.txt](https://github.com/JGCRI/gcamdata/files/2120664/diff_L244.building_USA.txt)

### DIFF
--- a/R/zchunk_L244.building_USA.R
+++ b/R/zchunk_L244.building_USA.R
@@ -518,20 +518,9 @@ module_gcam.usa_L244.building_USA <- function(command, ...) {
     # L244.GenericServiceSatiation_gcamusa: Satiation levels assumed for non-thermal building services
     # Just multiply the base-service by an exogenous multiplier
     L244.GenericServiceSatiation_gcamusa <- L244.GenericBaseService_gcamusa %>%
-      filter(year == max(BASE_YEARS))
-
-    # When adding floorspace, we should take floorspace from max(BASE_YEARS) as well
-    # Instead we take the first value, which ends up being floorspace from min(BASE_YEARS)
-    if(OLD_DATA_SYSTEM_BEHAVIOR) {
-      L244.GenericServiceSatiation_gcamusa <- L244.GenericServiceSatiation_gcamusa %>%
-        # Add floorspace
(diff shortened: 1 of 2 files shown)


pkyle:
Darn, I hadn't paid attention to one thing here--the satiation multiplier on the "commercial non-building" energy went from 1.3 to 2 in this revision, and I think in general that's realistic, because this service stands for a wide range of electricity demands that might increase significantly in the future. However Gokul just alerted me a few days ago to a problem here, whereby we're getting something like 30% growth in this demand from 2010 to 2015. It has to do with the satiation service function, and that we're not using a "satiation-adder" to calibrate our starting position on that function. Without telling the model where to start, it ends up wherever it ends up, which in this case is apparently not where we want it, and in the first time period we get unrealistic growth. So, if it's not too much trouble, I'd suggest reducing that to 1.5 or keeping it at 1.3, just until we look into this further. I could see reading in a "satiation-adder" here to keep the growth steady over time instead of the heavily near-term-loaded. What do you think?

pralitp:
If that is what we are going to set in the latest GCAM-USA work then I'm fine with it.

The idea of this PR isn't really to tweak any model results.  It was just that the values as they _were_ set was invalid, thus we needed to set them to _something_ and so my thought was just set it to whatever they would be in the next version of GCAM-USA.

Issue #179, opened 2017-03-31
Possible mistake in original chunk L110.water.demand.primary


swd-turner:
Output for this chunk ("L110.water_demand_primary_R_S_W_m3_GJ") is uniform across all regions... does not reflect different fresh/saline fractions for Middle East (see input file resource_water_share.csv). This seems to be caused by a merging error in the original code.

I've coded a fix in the updated chunk and maintained the old behavior under the OLD_DATA_SYSTEM_BEHAVIOR trap.

skim301:
Good catch. The saline fraction should be different for Middle East. Looks like the new code has caught that.

skim301:
I think we can close the issue since you have a fix.

PR #994, opened 2018-06-21
OLDNEW - L110.water.demand.primary

russellhz:
This code changes two files: `L110.water_demand_primary_R_S_W_m3_GJ` and `L210.TechCoef`. The changes are small in absolute value but large in percentage change. They are all changes to the Middle East, which used to be given the coefficients for USA, but now use their own calculated values.

diff_l110


bpbond:
Someone should look at those changed values and confirm. @mihejazi @d3y419 @mwisepnnl @skim301 Can you one you check and say yea/nay please?

bpbond:
See also issue #179 where this is documented.

PR #1002, opened 2018-06-26
OLDNEW for LA154.Transport

pralitp:
Having to do with not taking into account mode before a merge.

This only affects a couple data products:
```
Rows in x but not y: 4079, 4077, 4073, 4059, 4057, 4053, 4051, 4049, 4041, 4039, 4033[...]. Rows in y but not x: 4076, 4073, 4067, 4065, 4062, 4060, 4053, 4049, 4042, 4035, 4022[...]. 
L154.out_mpkm_state_trn_nonmotor_Yh.csv doesn't match

Rows in x but not y: 403, 401, 399, 397, 395, 393, 391, 389, 387, 385, 383[...]. Rows in y but not x: 405, 401, 399, 395, 390, 388, 387, 381, 379, 369, 354[...]. 
L254.StubTranTechProd_nonmotor_USA.csv doesn't match
```

Statistical differences look like they may be big:
[diff_LA154.Transport.txt](https://github.com/JGCRI/gcamdata/files/2138262/diff_LA154.Transport.txt)

### DIFF
--- a/R/zchunk_LA154.Transport.R
+++ b/R/zchunk_LA154.Transport.R
@@ -124,79 +124,17 @@ module_gcam.usa_LA154.Transport <- function(command, ...) {
         select(state, year, value_share) ->
         Pop_state_share
 
-      if(OLD_DATA_SYSTEM_BEHAVIOR) {
-        # Old code used the function, apportion_to_states, but did not match transportation mode (i.e., cycle and walk).
-        # For every year, there is a single national value for each mode. I believe the old code alternated between the cycle and walk
-        # values for each state down the list. However, the list was arranged with the first half being cycle and the last half being walk.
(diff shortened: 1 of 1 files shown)


pralitp:
@pkyle, this affects calibration data but only non-motor.  Do you think we still need to check calibration on this one?

pkyle:
Since these modes don't demand any fuel, they won't be capable of causing calibration/solution errors on their own. However if the base-service values for "trn_pass" that are read to the model are not exactly equal to the sum of all the modes' output, including these, then we may fail calibration. I don't expect that that's an issue here though, but if you're at all concerned about it then you can test it.

pralitp:
It appears to be re-calculating the base service so I think it should be good, thanks!

PR #1012, opened 2018-06-28
OLDNEW for LA114.wind

pralitp:
Which switches to a different strategy for evaluating the country level wind supply curves (we wanted to do so for a couple of reasons: avoid excessive loops, and vastly speed up the calculation).  The OLD_DATA_SYSTEM just read the previous price points from temp-data-inject because the old method was so slow.

The old approach was to continue to evaluate the curves until the value wasn't changing anymore with in some tolerance (with some minumum number of calculations to ensure we don't quit in a shallow start to the curve).  The new approach would only evaluate prices evenly spaces from min.price to the 95th percentile of max production (the percent is configurable)

Note this removed the need for temp-data-inject/L114.OLDregional_price_supply_points.csv the last temp-data!

This affects a couple data products:
```
Rows in x but not y: 32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22[...]. Rows in y but not x: 32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22[...]. 
L114.RsrcCurves_EJ_R_wind.csv doesn't match

Rows in x but not y: 32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22[...]. Rows in y but not x: 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21[...]. 
L210.SmthRenewRsrcCurves_wind.csv doesn't match
```

And the statistical differences:
[diff_LA114.wind.txt](https://github.com/JGCRI/gcamdata/files/2145324/diff_LA114.wind.txt)


bpbond:
>the last temp-data!

🎉 

Looks like differences up to 5%?

R/zchunk_LA114.wind-energy.R
```diff
@@ -72,35 +70,16 @@ module_energy_LA114.wind <- function(command, ...) {
...
-        mutate(mid.price = mid.price * gdp_deflator(1975, 2007) / CONV_KWH_GJ,
-               base.price = base.price * gdp_deflator(1975, 2007) / CONV_KWH_GJ) ->
-        L114.RsrcCurves_EJ_ctry_wind
-
-    } # End of old system behavior function
+    # Country-level supply curves from the Zhou file are used to build up the regional supply curves
+    Zhou_wind_supply_ctry_EJ %>%
+      rename(mid.price = `mid-price`, curve.exponent = `curve-exponent`, base.price = `base-price`) %>% # Modifying title names
+      # Romania is listed in the Zhou file as rou and in the iso_GCAM_regID file as rom. rou will be changed to rom.
+      mutate(iso = replace(iso, iso == "rou", "rom")) %>%
```
bpbond: Do we want to open an issue to fix this?

pralitp: Ehh, so easy and doesn't affect anything else:
```
WE34383:server2 pralitp$ diff -q outputs/L114.RsrcCurves_EJ_R_wind.csv compdata_updates/L114.RsrcCurves_EJ_R_wind.csv 
WE34383:server2 pralitp$
```

Might as well take care of it now.

pralitp:
Yes, some differences show up that they may be large but these parameters are at the tail end of the curve and are misleading.  We could choose to go further out (we settled on 95% with the trade off of more processing time) however when we look at the results in the realm that actually matters these are really close:
[wind_curves.pdf](https://github.com/JGCRI/gcamdata/files/2146051/wind_curves.pdf)

Except maybe `EU-12` however I think this is due to Romania (from iso_GCAM_regID): `rom,Romania,Eastern Europe,12`

PR #1018, opened 2018-06-28
OLDNEW for L111.nonghg which had division of small numbers

kvcalvin:
Which switches to default values when dividing rounded numbers. The original data system had division of a lot of small numbers, resulting in non-zero emissions factors. The new approach results in zeroes in these cases.

This affects one data product:
```
Rows in x but not y: 218617, 319012, 373487, 373484, 368849, 368848, 364288, 48270, 321726, 321724, 321719[...]. Rows in y but not x: 373487, 373485, 373484, 371169, 371167, 371166, 364291, 364289, 364288, 321726, 321724[...]. 
L111.nonghg_tgej_R_en_S_F_Yh.csv doesn't match
```

And the statistical differences:
[diff_L111.nonghg_en_R_S_T_Y.txt](https://github.com/JGCRI/gcamdata/files/2146984/diff_L111.nonghg_en_R_S_T_Y.txt)


kvcalvin:
Also, note that this PR doesn't affect any level2 files, so I don't think it changes GCAM inputs.

PR #1019, opened 2018-06-29
Fix a regression of proposal 244

### LINKED ISSUE #875: reduce_mgd_carbon is not being applied
In GCAM 4.4 we fixed a bug where carbon densities were not being adjusted between managed forest and unmanned forest, for instance.  That fix didn't get properly merged in GCAM 5.  We need to just apply the `reduce_mgd_carbon` on the appropriate land level 3 tibbles as before.

pralitp:
Reduce carbon contents of managed forest and pasture where the carbon content was not being properly reduced from the unmanged values.  The regression came somewhere in the merge of GCAM 4.4 changes into GCAM 5.  This closes #875.

Only direct data products are affected:
```
Rows in x but not y: 377, 376, 374, 372, 369, 368, 366, 365, 363, 362, 361[...]. Rows in y but not x: 378, 377, 375, 373, 371, 370, 369, 368, 366, 364, 363[...]. 
L222.LN2_MgdCarbon.csv doesn't match

Rows in x but not y: 336, 335, 331, 330, 328, 327, 326, 325, 324, 323, 321[...]. Rows in y but not x: 336, 335, 332, 328, 326, 324, 318, 317, 311, 309, 304[...]. 
L2231.LN3_MgdCarbon_noncrop.csv doesn't match
```

And they are reduced exactly as we expected:
[diff_reduce_mgd_carbon.txt](https://github.com/JGCRI/gcamdata/files/2147634/diff_reduce_mgd_carbon.txt)

### DIFF
--- a/R/constants.R
+++ b/R/constants.R
@@ -240,6 +240,12 @@ aglu.AVG_WOOD_DENSITY_KGM3 <- 500 # In kg per m3
 # Carbon content of wood is about 50 percent across species
 aglu.AVG_WOOD_DENSITY_KGCM3 <- 250 # In kg carbon per m3
 
+# Carbon content adjustments from unmanaged to managed
+aglu.CVEG_MULT_UNMGDFOR_MGDFOR <- 0.5
+aglu.CSOIL_MULT_UNMGDFOR_MGDFOR <- 0.87      #source: Guo and Gifford 2002
+aglu.CVEG_MULT_UNMGDPAST_MGDPAST <- 0.5
+aglu.CSOIL_MULT_UNMGDPAST_MGDPAST <- 0.9     # stay conservative here b/c no data source
+
 # Average Agriculture Density kg/m^3 for mass conversion
 # Source: http://www.engineeringtoolbox.com/wood-density-d_40.html
 aglu.AVG_AG_DENSITY <- 1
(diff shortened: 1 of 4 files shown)


R/constants.R
```diff
@@ -240,6 +240,12 @@ aglu.AVG_WOOD_DENSITY_KGM3 <- 500 # In kg per m3
 # Carbon content of wood is about 50 percent across species
 aglu.AVG_WOOD_DENSITY_KGCM3 <- 250 # In kg carbon per m3
 
+# Carbon content adjustments from unmanaged to managed
+aglu.CVEG_MULT_UNMGDFOR_MGDFOR <- 0.5
```
bpbond: What are the units here? Fraction C in biomass?

R/constants.R
```diff
@@ -240,6 +240,12 @@ aglu.AVG_WOOD_DENSITY_KGM3 <- 500 # In kg per m3
...
+aglu.CSOIL_MULT_UNMGDFOR_MGDFOR <- 0.87      #source: Guo and Gifford 2002
```
bpbond: Provide DOI if possible

R/constants.R
pkyle: No, this is unitless--it's the conversion factor from unmanaged forest to managed forest, where the latter is understood to be forest not in logging rotation, and the latter is forest in logging rotation. The average vegetation biomass of the logged forest is assumed to be 50% of that of the unlogged forest (integrated over the rotation period). Using 50% under the assumption that the veg biomass of the logged forest over the rotation period can be approximated by a triangle.

R/constants.R
pkyle: https://doi.org/10.1046/j.1354-1013.2002.00486.x

bpbond: Thanks, Page. @pralitp I'd suggest adding these clarifications to `constants.R`.

PR #1017, opened 2018-06-28
OLDNEW for LA143.HDDCDD

pralitp:
having to do with:
* Not dropping Cote 'dIvoire
* New method for weighted mean DD using population as weights

This changes a surprising number of data products:
```
1. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
Rows in x but not y: 2409, 2410, 2902. Rows in y but not x: 2902, 2410, 2409. 
L101.in_EJ_R_en_Si_F_Yh.csv doesn't match

2. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
Rows in x but not y: 311855, 54526, 54517, 33022, 14086, 26375, 59649, 54332, 54529, 14287, 59838[...]. Rows in y but not x: 311855, 59838, 59649, 54529, 54526, 54517, 54332, 53208, 40919, 33022, 14287[...]. 
L111.nonghg_tg_R_en_S_F_Yh.csv doesn't match

3. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
Rows in x but not y: 18539, 9704. Rows in y but not x: 9704, 18539. 
L112.ghg_tg_R_en_S_F_Yh.csv doesn't match

4. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
Rows in x but not y: 20241, 19811, 924, 1364, 3072, 19807, 20236, 933, 19810, 19805, 1353[...]. Rows in y but not x: 20236, 19811, 19810, 19807, 19805, 3072, 1364, 1353, 935, 933, 924[...]. 
L141.hfc_ef_R_cooling_Yh.csv doesn't match

5. Failure: matches old data system output (@test_oldnew.R#99) -----------------
dim(olddata) not identical to dim(newdata).
1/2 mismatches
[1] 103008 - 103240 == -232
Dimensions are not the same for L143.HDDCDD_scen_ctry_Y.csv

6. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
Different number of rows
L143.HDDCDD_scen_ctry_Y.csv doesn't match

7. Failure: matches old data system output (@test_oldnew.R#112) ----------------
round_df(olddata) not equivalent to round_df(newdata).
[...]

PR #1020, opened 2018-06-29
Omnibus 1015 1017 1018

pralitp:
This includes #1015 #1017 #1018


pralitp:
@rplzzz, do you mind checking off.  This is meant to combine already approved PRs listed above where we now expect all of the tests to pass as the comparison data has been updated.

PR #1016, opened 2018-06-28
OLDNEW for L161.nonghg, which keeps region 25

kvcalvin:
This keeps region 25 in the SSP2 emissions factors. The old approach dropped this region, since it didn't have coal electricity generation.

This affects a few data products: 
```
dim(olddata) not identical to dim(newdata).
1/2 mismatches
[1] 24460 - 25180 == -720
Dimensions are not the same for L161.SSP2_EF.csv

test_oldnew.R:112: failure: matches old data system output
round_df(olddata) not equivalent to round_df(newdata).
Different number of rows
L161.SSP2_EF.csv doesn't match

test_oldnew.R:99: failure: matches old data system output
dim(olddata) not identical to dim(newdata).
1/2 mismatches
[1] 5397 - 5563 == -166
Dimensions are not the same for L251.ctrl.delete.csv

test_oldnew.R:112: failure: matches old data system output
round_df(olddata) not equivalent to round_df(newdata).
Different number of rows
L251.ctrl.delete.csv doesn't match

test_oldnew.R:99: failure: matches old data system output
dim(olddata) not identical to dim(newdata).
1/2 mismatches
[1] 12464 - 12711 == -247
Dimensions are not the same for L251.ssp2_ef_vin.csv

test_oldnew.R:112: failure: matches old data system output
round_df(olddata) not equivalent to round_df(newdata).
Different number of rows
L251.ssp2_ef_vin.csv doesn't match

test_oldnew.R:99: failure: matches old data system output
dim(olddata) not identical to dim(newdata).
1/2 mismatches
[1] 116185 - 119605 == -3420
Dimensions are not the same for L251.ssp2_ef.csv

test_oldnew.R:112: failure: matches old data system output
round_df(olddata) not equivalent to round_df(newdata).
Different number of rows
L251.ssp2_ef.csv doesn't match
```
And the statistical differences:
[diff_L161.nonghg_en_ssp_R_S_T_Y.txt](https://github.com/JGCRI/gcamdata/files/2147329/diff_L161.nonghg_en_ssp_R_S_T_Y.txt)

PR #1021, opened 2018-06-29
Omnibus 1012 1016 1019

pralitp:
This includes #1012 #1016 #1019


bpbond:
Thanks @pralitp - you've been tearing through these. 👏  I'll let @kvcalvin approve.

PR #1022, opened 2018-06-30
OLDNEW gdp_deflator

pralitp:
Sop using select hard coded deflator values from the old data system and instead use a consistent approach regardless of conversion years.

*This is the last OLD_DATA_SYSTEM_BEHAVIOR!*. Of course we saved the biggest doozy for last.  This one affects 174 data products!

The statistical differences are small where the biggest I noticed where around 3%.  There were a couple of things showed up in there that I flagged: IncomeElasticity and MAC curves.  I manually looked at the elasticities and I'm not sure why the percent difference show up as large, actual differences are small.  And the MAC curves didn't merge well for comparison because the price points changed slightly but the reduction values did not change at all, as we would expect.

[diff_gdp_deflator.txt](https://github.com/JGCRI/gcamdata/files/2151318/diff_gdp_deflator.txt)


bpbond:
At this point we want a full run-through of data into GCAM?

pralitp:
Yes we will do that next.  I have already copied the XMLs of this PR into GCAM and gave them a go to see if there were any problems and it all ran fine!  Once we get all this into GCAM we will run a more exhaustive set of runs and verify everything.

rynacui:
Sorry all for bringing this up again! But apparently, the old-new gdp deflator affects GCAM-China as well... Shall I start by adding the oldnew behavior? Or do you have alternate suggestions?

The difference on provincial GDP is 0.3%, but the values are large, so the range is beteen 1.2 ~ 9821 mil USD

PR #1023, opened 2018-07-03
Fix a mapping/summing error in GHG en

kvcalvin:


kvcalvin:
Closing this. We decided to fix this issue on stash instead of github

bpbond:
But it will get back ported here too I assume?

PR #1070, opened 2018-12-02
Warn about trailing header commas, and clean up existing input files

### LINKED ISSUE #1069: Input parser could check for trailing commas on inst/extdata headers
...and warn or fail if it sees any.

bpbond:
Closes #1069


pralitp:
This PR looks fine.  Although the underlying problem (I assume Excel?) still remains so I assume we should start checking for this sort of thing in future PR reviews.
Maybe if we really cared enough we could make a test that checks all the files and fail but given we have bigger fish to fry we can shelf that for a later time.

bpbond:
This is throwing all kinds of errors on Travis
```

 ERROR
Running the tests in ‘tests/gcamdata.R’ failed.
Last 13 lines of output:
  4: lazyeval::lazy_eval(code[[length(code)]])
  5: eval(x$expr, x$env, emptyenv())
  6: eval(x$expr, x$env, emptyenv())
  7: driver(quiet = TRUE, write_outputs = FALSE)
  8: run_chunk(chunk, all_data[input_names])
  
  ══ testthat results  ═══════════════════════════════════════════════════════════
  OK: 2503 SKIPPED: 1 FAILED: 4
  1. Failure: save_chunkdata saves (@test_fileutil.R#61) 
  2. Error: matches old data system output (@test_oldnew.R#26) 
  3. Failure: chunks handle timeshift (@test_timeshift.R#34) 
  4. Error: chunks handle timeshift (@test_timeshift.R#34) 
```
...but I assume this is being caused by the `readr` upgrade (see #1072 ). If everything passes on my local machine, running the older `readr 1.1.0`, I'm going to merge this, and move on ASAP to #1072 so we can be sure we have a `gcamdata` that works.

bpbond:
...wow my new laptop runs the driver a _lot_ faster. 🎉

pralitp:
@bpbond, I posted this on slack but looks like you didn't see it so I'll just post here too:
I notice #1063 is still open.
If I remember correctly I was basically ready to merge the thing.  I updated `gcamdata.compdata` but never finished it off because I could never get the tests to complete.
What should we do about this.  Specifically if:
1) The new `readr` version is likely going to torpedo any new attempts to runs tests anyways?
2) Any other PR tests will fail because I have already updated the `gcamdata.compdata` for that one

bpbond:
Bah sorry @pralitp - getting on Slack now. Yeah, we're hosed right now as far the PR tests, right? Can we update #1063 from `master`, test locally with old `readr`, and merge, so we can get #1072 taken care of? What do you think?

PR #1073, opened 2018-12-31
Changes for readr 1.3 behavior

### LINKED ISSUE #1072: Explicitly specify column data types for `load_csv_files`
Although it generally does a good job, we probably shouldn't rely on `readr`'s automatic column type detection for data that we store inside the instead. Instead, we should explicitly pass the `col_types` argument, which can be either a single abbreviated character string (`"ccind"`, which means "character", "character", "integer", "number/double", "date") or a list of column specifications.

A convenient way to do this would be to store the column type information in the file header. This is the approach I took in `metar` (#916; see [example file](https://github.com/ashiklom/metar/blob/master/inst/examples/example1.csvy#L5-L20)), though this requires the file to be read in two stages: first, parse the header to get the column types; and second, read the file body with the indicated header types. An advantage of using a `metar`-style YAML header is that it can readily be adapted to store factor levels, which can be another potential defense against typos in columns. more generally, YAML is a very versatile format that can be used to store a wide range of data structures, and is widely used.

As an alternative, we could store column types as another tag within the current header. We may even be able to store factor levels for some columns as a list, e.g.

```
# ...
# ColumnTypes: integer, date, factor, number
# FactorLevels: , , ["apples", "oranges", "bananas"], 
id, date, fruit, value,
1, 2018-10-10, "apple", 5,
2, 2018-10-11, "orange", 6
...
```

bpbond:
Closes #1072.

This PR adds a "Column types:" header to all input files; this is parsed and passed to the `col_types` argument of `readr::read_csv`. This addresses changes (again) in `readr` behavior, and should very slightly improve performance, as we're no longer asking `readr` to guess column types. (This raises an interesting question: do we need/want `readr` as a dependency at all, at this point?) The corresponding changes to the `gcamdata.compdata` package are in https://github.com/JGCRI/gcamdata.compdata/pull/22.

This PR runs and tests with no errors on my machine with the IEA energy data installed.

_Without_ the proprietary IEA data, however, a number of objects fail the oldnew test, although the numerical differences are tiny:
* L112.ghg_tg_R_en_S_F_Yh
* L132.water_km3_ctry_ind_Yh
* L154.cost_usdvkm_R_trn_m_sz_tech_F_Y
* L154.intensity_MJvkm_R_trn_m_sz_tech_F_Y
* L154.loadfactor_R_trn_m_sz_tech_F_Y
* L254.BaseService_trn
* L254.StubTechCalInput_passthru
* L254.StubTranTechLoadFactor

I'm not sure what the best approach is here, because we can't guarantee that any object _x_ written out as a CSV and then read back in (as in the prebuilt data) will be identical. Do we save the prebuilt objects using `devtools::use_data`, so they're guaranteed identical? Provide different versions in `gcamdata.compdata:::COMPDATA`?

### DIFF
--- a/tests/testthat/test_oldnew.R
+++ b/tests/testthat/test_oldnew.R
@@ -110,6 +110,10 @@ test_that("matches old data system output", {
         expect_equivalent(round_df(olddata), round_df(newdata), info = paste(basename(newf), "doesn't match"))
       }
     }
+    # Explicitly clean up COMPDATA as it uses a lot of memory and gets loaded into the
+    # Global environment
+    rm(COMPDATA, envir = .GlobalEnv)
+    gc()
   }
 })
 


R/utils.R
```diff
@@ -35,8 +35,14 @@ load_csv_files <- function(filenames, optionals, quiet = FALSE, ...) {
       if(!quiet) message("Note: optional input ", f, "not found")
       next
     }
-    suppressMessages(readr::read_csv(fqfn, comment = COMMENT_CHAR, ...)) %>%
-      parse_csv_header(fqfn) %>%
+
+    # Read the file header and extract the column type info from it
+    assert_that(file.exists(fqfn))
+    header <- readLines(fqfn, n = 20)  # only first 20 lines
```
ashiklom: 20 lines is probably a safe assumption, but if you wanted to make this more robust, you could do something like this:

``` r
# For testing
string <-
"# Description: Something
# Source: something else
# Column types: this
a, b, c, d"
con <- textConnection(string)

# Real code would be:
# con <- file(path, "r")

is_comment <- TRUE

header <- character()

while (is_comment) {
  line <- readLines(con, n = 1)
  is_comment <- grepl("^#", line)
  if (is_comment) {
    header <- c(header, line)
  }
}

header
#> [1] "# Description: Something" "# Source: something else"
#> [3] "# Column types: this"
```

This is the strategy I used to read the YAML header for [`csvy`](https://github.com/leeper/csvy/blob/master/R/get_yaml_header.R) and [`metar`](https://github.com/ashiklom/metar/blob/readr_backend/R/read_yaml_header.R).

R/admin.R
```diff
@@ -203,3 +203,99 @@ find_multiple_mutates <- function(path = "R/") {
...
+#' @return Nothing.
+add_column_types_header_line <- function() {
+
+  files <- list.files(path = "inst/extdata/", pattern = "csv$", full.names = TRUE, recursive = TRUE)
+
+  colchars <- c("character" = "c", "integer" = "i", "numeric" = "n", "double" = "d", "logical" = "I")
+
+  for(f in files) {
+    cat(f, "...")
+    header <- readLines(f, n = 20)
```
ashiklom: See my comment [below](https://github.com/JGCRI/gcamdata/pull/1073/files#r244758863).

inst/extdata/aglu/FAO/FAO_USA_an_Prod_t_PRODSTAT.csv
```diff
@@ -1,6 +1,8 @@
 # File: FAO_USA_an_Prod_t_PRODSTAT.csv
 # Title: FAO USA historical production animal products
 # Units: tonnes
+# Column types: ciciciiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii
```
ashiklom: Are we sure we want these (and related values) to be parsed as integers? I feel like production values, in tons, should allow fractions (i.e. should be parsed as "n").

Since I'm guessing we have fewer places where we expect integers than "numbers" (doubles), perhaps it would be safer to re-generate these files with the newer version of `readr` so all of these parse as numbers and then manually go back and fill in the few places where we actually expect integers (e.g. years).

DESCRIPTION
```diff
@@ -41,7 +41,7 @@ Imports:
     magrittr (>= 1.5),
     tibble (>= 1.1),
     tidyr (>= 0.7.1),
-    readr (>= 1.0.0),
+    readr (>= 1.3.1),
```
pralitp: I was going going to comment about if bumping the required version was really necessary but after talking to you today I realize it is (to ensure consistency for OLD/NEW tests).  We should make sure we bump the version in `gcamdata.compdata` as well.

R/admin.R
```diff
@@ -203,3 +203,99 @@ find_multiple_mutates <- function(path = "R/") {
...
+    header <- header[grepl("^#", header)]
+    original_header_length <- length(header)
+
+    # Remove any previous Column types entry
```
pralitp: Should we have this function keep previous column type entries?  I'm thinking of the case that a user needs to go back and correct a CSV manually because `read.csv` didn't detect it correctly.  We wouldn't want to have to go back and redo the manually corrections every time we run this function.

On a maybe related note if I re-run this function on the latest commit I do get some files that show up modified, was that expected?:
```
diff --git a/inst/extdata/gcam-usa/UCS_Database.csv b/inst/extdata/gcam-usa/UCS_Database.csv
-# Column types: iccccccccnnicinicccnncccccc 
+# Column types: iccccccccnniciniccccccccccc 
diff --git a/inst/extdata/socioeconomics/SSP_database_v9.csv b/inst/extdata/socioeconomics/SSP_database_v9.csv
-# Column types: cccccccccccccccccnnnnnnnnnnnnnnnnnnncccccccccc 
+# Column types: cccccccccccccccnnnnnnnnnnnnnnnnnnnnncccccccccc 
diff --git a/inst/extdata/water/IBNET_municipal_water_cost_USDm3.csv b/inst/extdata/water/IBNET_municipal_water_cost_USDm3.csv
-# Column types: icnnn 
+# Column types: icncn 
```

inst/extdata/aglu/FAO/FAO_USA_an_Prod_t_PRODSTAT.csv
bpbond: I think @pkyle said that, for whatever reason, these are all "integers"--that's how they're reported in the file from the FAO; this isn't a `readr`-versioning issue. We could still force them to be read as numeric, of course, but that breaks the existing code.  😢

R/utils.R
bpbond: Yeah, I was being lazy here. Thanks @ashiklom will fix.

DESCRIPTION
bpbond: I've done this.

R/admin.R
bpbond: To the second question, yes, I had to hand-correct some files, so that's what you're seeing when you rerun. To the first...not sure. 😕

R/utils.R
bpbond: Only thing I added was `on.exit(close(con))` because otherwise was getting lots of warnings.

bpbond:
These commits cover what we talked about today @pralitp . We now use internal package data to store prebuilt files (the outputs of the four immediate dependents of the proprietary IEA data chunk `module_energy_LA100.IEA_downscale_ctry`). This means that the outputs of these four...
* module_energy_LA101.en_bal_IEA
* module_energy_LA111.rsrc_fos_Prod
*  module_energy_LA118.hydro
*  module_energy_LA121.oil

...should be bit-for-bit identical between the IEA-present and non-IEA-present cases.

bpbond:
All the non-xml tests (haven't installed java yet on this new laptop) pass for both cases on my machine.

inst/extdata/aglu/FAO/FAO_USA_an_Prod_t_PRODSTAT.csv
ashiklom: Fair enough. This just caught my eye and I just wanted to make sure it was intentional.

bpbond:
OK, as noted above this passes all tests on my machine, although–important caveat!–I'm not running `test_xml_conv_utilities.R` or `test_xml_diff.R`.

bpbond:
@pralitp What do you think about the next step here? Merge the `gcamdata.compdata` PR and then see if this starts passing on Travis?

inst/extdata/aglu/FAO/FAO_USA_an_Prod_t_PRODSTAT.csv
pkyle: Wait, it's the opposite--these should be doubles, not integers. Sorry if I screwed something up here--any numeric value (whether in tonnes or ktoe) should be a double. It's a common practice in these datasets to use a small unit and then not write out any decimals.

bpbond: Sorry. I mean, they're reported as "integers", i.e. with no fractional part, and read in as such, but then quickly become numerics as they're computed on in [module_aglu_LB132.ag_an_For_Prices_USA_C_2005](https://github.com/JGCRI/gcamdata/blob/master/R/zchunk_LB132.ag_an_For_Prices_USA_C_2005.R). This is what currently happens, anyway, and matches the behavior in the old data system.

pralitp:
Thanks @bpbond, I've gone ahead and merged the `gcamdata.compdata` and I will go ahead and push one more commit that we talked about to work around memory issues which I think will finally get the Travis tests passing again.

PR #1077, opened 2019-01-04
make the creation of biomass_grass technologies/LandLeafs consistent

### LINKED ISSUE #1076: Force consistency in creation of biomass_grass technology/LandLeaf
This is an issue that came up for me in using the actual member states of the European Union to define our model regions. Our current EU regions (`EU-15` and `EU-12`) include the "micro-states" (e.g., Andorra, Liechtenstein, Monaco, San Marino, and others) that are not actually EU members. If we instead map those countries to one of our Non-EU Europe regions (e.g., `Europe_Non_EU` or `European Free Trade Association`), we generate some small land use regions whose cropland is assigned entirely to "OtherArableLand", as most of these countries aren't included in the FAO PRODSTAT data. (They are technically in there, but all values are zero.) That approach is fine, but the problem is that we are inconsistent in deciding whether to create a biomass_grass AgProductionTechnology and LandLeaf in land use regions that have cropland, but whose cropland could not be assigned to any specific crop commodities. The following XML files generate a biomass_grass technology in such a land use region:
```
ag_cost_IRR_MGMT.xml
ag_Fert_IRR_MGMT.xml
ag_prodchange_ref_IRR_MGMT.xml
ag_water_input_IRR_MGMT.xml
all_aglu_emissions_IRR_MGMT.xml (with a nocreate flag)
```
And, the following XML files exclude a biomass_grass technology/LandLeaf for such a land use region:
```
ag_For_Past_bio_base_IRR_MGMT.xml
land_input_4_IRR_MGMT.xml
land_input_5_IRR_MGMT.xml
```
This causes the model to crash. I think I'd recommend not creating them; for one thing, there's no data from which to set the rainfed:irrigated ghost shares. Also, the regions are so small we don't really lose anything by excluding them. Alternatively, if we did include them, the data therein would be iffy (biomass_grass would be the only crop available).

pkyle:
Closes #1076 

This PR avoids writing out yields of bioenergy grass crops in land use regions that have agriculture in Monfreda/LDS but not FAOSTAT, in `module_aglu_LB113.bio_Yield_R_GLU`. This addresses an issue whereby a slight change in GCAM's country-to-model-region assignments could cause the model to crash, as some of the XML files have a biomass_grass technology in the problematic land use regions, but other XML files do not. See [Issue 1076](https://github.com/JGCRI/gcamdata/issues/1076) for a more complete explanation. This change allows (e.g.) San Marino to be assigned to a non-EU region. It does not change any files written to the `/xml` directory with GCAM's default country-to-region mappings.


bpbond:
I don't see any reason to delay merging this--don't need to wait for #1073 . @pralitp ?

PR #1105, opened 2019-06-06
Don't let users load package if readr version too low

### LINKED ISSUE #1086: Need to error (or at least warn) with older readr!
Per @pkyle experience on Slack:

>Just FYI, as this is the first I've seen/heard of this--in my home directory on PIC, working off of a recent (post-readr-1.3.1) branch off of master, I built XMLs and tried to run the model, but failed calibration/solution in 1990 (in all markets I think) and crashed out shortly thereafter. Worked a while with @pralitp on it...basically, all of my XML files built without warnings or errors, but most had calibration quantities that were incorrect. Sometimes only ~1e-5 off, but in a lot of the ag sector the values were an order of magnitude off. No rhyme or reason that I could determine for which quantities were correct, but the cause of the problem ended up being the readr package on PIC, which for my home directory was 1.1.1. And the solution was to update readr.

>But yeah, it's hard to say what must have happened, that gcamdata cranked out a bunch of wrong numbers with no error messages or warnings, and somehow those bad numbers themselves didn't crash anything. :thinking_face: My hat is off to the readr developers!

bpbond:
Closes #1086

Note I put the check code in `.onload()` rather than `driver()` as suggested by @ashiklom in #1086 - mostly because the latter is already cluttered up with lots of code.


R/onload.R
```diff
@@ -1,6 +1,14 @@
 # onload.R
```
ashiklom: For future reference, the [convention](http://r-pkgs.had.co.nz/r.html#r-differences) (see to the end of the "When you do need side effects" section) in the R world seems to be to stick `.onLoad` and co. into a file called `R/zzz.R`. I think this is 95% convention, but the last 5% may be because, by default, I think the R package build process [sources files in alphabetical order](https://cran.r-project.org/web/packages/roxygen2/vignettes/collate.html), and this ensures that these are sourced last (and therefore, take precedence).

bpbond: Ah I knew about the `zzz` convention but always assumed it was a make-it-easy-to-find issue; didn't think about the scoping. Anyway, thanks; in this case we have so many `z*.R` chunks it seemed easier to stick above them.

bpbond:
@ashiklom or @pralitp , do you think OK to merge this given scoping issue mentioned above? I can also move things to a `zzz.R` to follow convention.

PR #1112, opened 2019-06-19
Protoextrapolate

abigailsnyder:
This branch contains two different implementations of `extrapolate_constant` - one that works on vectors, and one that works on grouped data frames that have a column named `value`. 
Both are implemented in the test chunk `zchunk_L100.GDP_hist.R`. 

Absolutely expect there will be iterations among us, but this PR is relatively short at least. Reviewers can take some time looking through it and I'll start on the implementations of the `extrapolate_linear`, `extrapolate_correlated`.


R/constants.R
```diff
@@ -25,6 +25,11 @@ MODEL_BASE_YEARS        <- c(1975, 1990, 2005, 2010)            # calibrated per
 MODEL_FUTURE_YEARS      <- seq(2015, 2100, 5)                   # future (i.e., not calibrated) time periods in the model
 MODEL_YEARS             <- c(MODEL_BASE_YEARS, MODEL_FUTURE_YEARS)
 
+# intermediate BYU year constant. Need for early dev where not all data sets get extended.
+# Goal would be to have it handled automatically with a change to HISTORICAL_YEARS.
+# Should be able to just find and replace BYU_YEAR with max(HISTORICAL_YEARS) once the
+# BYU skeleton is in place for all chunks.
+BYU_YEAR                <- 2015
```
bpbond: 👍

R/extrapolators.R
```diff
@@ -0,0 +1,86 @@
...
+#' @importFrom assertthat assert_that is.scalar
+#' @author ACS June 2019
+extrapolate_constant <- function(x, n=1){
+
+  assert_that(is.numeric(x))
+  assert_that(is.scalar(n))
+
+  meanval <- mean(last_n(x,n))
+
+  x[is.na(x)] <- meanval
```
bpbond: This will obviously fill in _all_ NA values, not just at the end of the time series. Is this OK? 🤔

R/extrapolators.R
```diff
@@ -0,0 +1,86 @@
...
+#' in with the specified mean value.
+#' @importFrom assertthat assert_that is.scalar
+#' @author ACS June 2019
+extrapolate_constant_optionB <- function(x, n=1){
+
+  assert_that(is.tibble(x))
+  assert_that(is.scalar(n))
+
+
+  # can't use the faster replace_na function because
```
bpbond: Ah!

R/extrapolators.R
```diff
@@ -0,0 +1,86 @@
...
+  # if x is grouped, e.g. on iso, each group will have
+  # its own, specific mean value to replace NAs with.
+  x %>%
+    mutate(meanval = mean(last_n(value, n)),
+           value = if_else(is.na(value),
+                           meanval,
+                           value)) %>%
+    select(-meanval) %>%
+    distinct
```
bpbond: ?

R/extrapolators.R
abigailsnyder: Good catch - that's what will happen in both implementations (vector and tibble) and should definitely be corrected.

R/extrapolators.R
abigailsnyder: For some reason, rows get duplicated in this pipeline. I didn't spend a ton of time tracking it down; I figured I would do that if we decided we preferred this tibble implementation to the vector one.

R/extrapolators.R
pralitp: This is a pretty important question.  Certainly we don't want to "interpolate" with the extrapolator.  I mean we could tell users to do any interpolations before extrapolations but that seems less than ideal.

What about a case for instance the dataset coming in was expected to be filled up to say 2010 and we want to extrapolate out to 2015 but region X was actually missing a value in 2010.  These methods would happily extrapolate the 2010 value as well in region X.  Maybe that is fine as well and we just make clear to users that if they expect values complete data up to year Y they should ensure that before passing it into the extrapolators.

abigailsnyder:
@pralitp returning to this now  - I too am leaning toward option A (the vector option). In some ways I think it's more flexible, and also I have never gotten NSE stuff to work to pass the column name as an argument for the grouped tibble version. 

In terms of shortening the pipeline leading to the call in each chunk, the only real shortening I can see (for either option) is rolling the 
```
repeat_add_columns(missingyears)
```
step into the extrapolator function. I am working on that now. It's a little tricky to think through in the vector form. Do you have an alternative suggestion?  
Also, I don't think we want the extrapolator functions to actually calculate which years are missing. I think part of our coding ethos for this package is keeping things sort of mono-use for user clarity.

abigailsnyder:
@pralitp update: I just realized rolling the 
```
repeat_add_columns(missingyears)
```
step into the vector form won't work, it will result in the wrong number of rows being returned to the data frame. 

If we want to roll that step into the extrapolate function(s), then we will have to go with the grouped tibble version B and either figure out the NSE stuff to have flexible column names, or just force a column name of `value`. I also don't like the latter option, but I am more ok than usual with it because these calls should be happening early in the pipelines when the tibbles are more likely to have a column named `value` anyway.

Alternative proposal (what I'm working on now): 
I'll add an extra helper function that will identify the missing years and do the `repeat_add_columns` step  to roll that up. That then separates figuring all that out to let us use the vector version A of extrapolation, while also cleaning up the pipelines in each chunk a little.
Let me know what you think!

pralitp:
@abigailsnyder, as I was taking another look.  Maybe the deal is `repeat_add_columns` isn't the right tool for the job because the data already has the `year` column.  Could it be done with just a `complete`?  Then the `group_by` would still be required but that isn't so bad.

abigailsnyder:
You are correct that it can be done by `complete`,  it will just involve information specific to the data frame being operated on in the call to complete. I'm fine with that  - I've basically been trying to do that in a more general helper function and that is turning into a nightmare because `complete` won't take column names as strings.
If we are ok having a call to complete for each data frame in each chunk, that's definitely easier.

abigailsnyder:
So it will look like this in this chunk
```
missingyears <- (max(long_iso_year_gdp$year) +  1):BYU_YEAR

      # Constant extrapolation operating only on
      # numeric vector. Can operate on any numeric vector, regardless of grouping
      # or column name.
      # What gets output in this chunk
      long_iso_year_gdp %>%
        complete(year = c(year, missingyears), iso) %>%
        group_by(iso) %>%
        mutate(value = extrapolate_constant(value, n = 1)) 
```

Which is definitely cleaner, but that call to `complete` will vary with each data set we deal with

pralitp:
Yea.. exactly the same arguments as to the `group_by`.  Not totally ideal but I am willing to live with it.

pralitp:
> won't take column names as strings.

Hmm, maybe this is where the `!!!` comes into play.

abigailsnyder:
Hm ok there's some other changes being made for this PR too, so I'll make all those with the ugly, in-chunk complete and then play around with trying to use 
```
!!!
``` 
in a more general function a little more after that.

pkyle:
One thing to look out for, I had a group_by() %>% complete() step in a data processing pipeline a few years ago and it took forever. So, if you do end up doing that, I'd recommend to test it out on a big dataset.

abigailsnyder:
@pkyle good point, but this usage should be `complete` (add missing years for every relevant IDs - just iso here) and _then_ the `group_by(ID)`. I think this should be pretty consistent across the BYU, but I'll keep your note in mind in case we run into any chunks that have your ordering

abigailsnyder:
@pralitp @bpbond ready for re-review with comments addressed. Passed check package locally right before I updated from master, so presumably that should still be fine

R/extrapolators.R
abigailsnyder: The new implementation of the extrapolator would handle this situation like this: 
The data set stops at 2010, but region X has an NA value in 2010. We want to do constant extrapolation to 2015. 

Assuming we grouped by region, everyone but X will behave the way we expect. 
As is, the extrapolator will take the 2009 value and fill it in for 2011-2015. The NA value in 2010 will be left alone. Endogenously determining how many tail end NA values are in a vector was proving frustrating, so I decided to just make an argument for that.

R/zchunk_L100.GDP_hist.R
```diff
@@ -34,13 +34,46 @@ module_socioeconomics_L100.GDP_hist <- function(command, ...) {
...
+
+      # Constant extrapolation operating only on
+      # numeric vector. Can operate on any numeric vector, regardless of grouping
+      # or column name.
+      # What gets output in this chunk
+      long_iso_year_gdp %>%
+        complete(year = c(year, missingyears), iso) %>%
+        group_by(iso) %>%
+        mutate(value = extrapolate_constant(value, n = 1,
+                                            numMissing = length(missingyears))) %>%
```
pralitp: In this example calculating `numMIssing` is straightforward.  But I think the typical use case will be that users just want to: `complete(year = HISTORICAL_YEARS, non, year, columns)` and may not explicitly know what years were originally present in their data set.

They could define some constant about what they are, and we do have many of these constants, but that is going to proliferate constants and the constant would need to get updated anytime the data set does.

Users could just figure it out by doing something like:
```
long_iso_year_gdp %>%
	        group_by(iso) %>%
                mutate(numMissing = BYU_YEAR - max(year)) %>% # BYU_YEAR for now, max(HISTORICAL_YEARS) at some point
	        complete(year = c(year, missingyears), iso) %>%

	        mutate(value = extrapolate_constant(value, n = 1, numMissing = numMissing) %>% # we get the whole vector but they are all the same
```
Which seems a bit cumbersome / wasteful.

What about in `extrapolate_constant` just filling from the "last non-NA value + 1" to the end?  That way it can be self-contained?

bpbond: >What about in extrapolate_constant just filling from the "last non-NA value + 1" to the end? That way it can be self-contained?

That was my thought too, on reading through the `extrapolate_constant()` code. Calculating `numMissing` seems like a huge drag.

The potential problem (recapping some of previous discussion) is that if there's a time series with final NA, this will overwrite that NA?

R/extrapolators.R
bpbond: Is this (filling in region X 2010 NA value) definitely a problem? As noted in previous comments, if we can allow this, it makes things much simpler (no need to calculate `numMissing`...).

R/zchunk_L100.GDP_hist.R
abigailsnyder: Also, right now the input argument `n` to `extrapolate_cosntate()`specifies how many of the last non-NA values to average to get a value for filling in tail end NA's. The idea is obviously that everyone will just use `n=1`, projecting the last reported value forward. 
However, if a value `n>1` is input, filling in the last "last non-NA value + 1" will overwrite a reported value. 
I think the case where 2010 value = NA so we take 2009 value and fill it in for 2011-onward but leave 2010 as NA is definitely weird and we would want to overwrite that final NA? One (possibly difficult) option would be to actually see if this comes up anywhere in the data system and if there's any special way of handling it other than 2010value=2009value when it does.
If we're ok with overwriting the final NA value by doing the last missingYears+1, then I do think we need to drop support for arbitrary `n` values as input and force that value to just be 1

R/extrapolators.R
pralitp: As I think about it, in our first go-round it won't be a problem because 2010 is our final calibration year, so by definition it is not missing.

Maybe we are just clear that this is what happens.  I think in future development the idea that region X would have to fill 2010 wouldn't be as strange as there wouldn't be this preconceived notion that we are only "extending to 2015 for BYU".

pralitp:
I've confused myself as to where we are on this discussion.  Did we settle on just filling tail `NA` values with average of last `n` and just being cool with that even if the number of values filled might be different by group (the should we fill missing `2010` problem)?

abigailsnyder:
I too am confused about where we have ended up. 
If we handle the <fill in missing 2010 values> via  <filling tail NA values with average of last n and just being cool with that even if the number of values filled might be different by group>, that's definitely more annoying to code but doable. 
I think it will be easiest for us to walk through options in person - I set a meeting for next week and I'll have some concrete examples of options prepared.
@bpbond is out of town all next week, so unless there is an objection, @pralitp and I will just make a decision (though Ben is still included on the meeting invite).

abigailsnyder:
@pralitp @bpbond updates made to the constant extrapolation method + the old-new test as discussed in person Aug 9, 2019. Just FYI

tests/testthat/test_oldnew.R
```diff
@@ -89,6 +89,18 @@ test_that("matches old data system output", {
...
 
+      # During the base year update development process, some outputs from the data
+      # system will be extended and therefore have different dimensions from the
+      # old comparison data. This causes the old-new test to fail.
+      # The extended extrapolation years will be dropped from the newdata
+      # so that comparison can be made to the old data. This also serves to check
+      # that the extrapolation procedure does not touch original data.
+      if(max(newdata$year) == BYU_YEAR){ # one way to check that it's a BYU without flags.
+        newdata %>%
+          filter(year <= max(HISTORICAL_YEARS)) ->
```
abigailsnyder: I think the `if` will do the trick but this `filter` is not the right way to handle the correction to get the dimensions to agree. So travis is still failing. 

The problem is that the last year of data in an input file may not actually be the `max(HISTORICAL_YEARS)`. For the GDP file, the last recorded year is 2013, not 2010. From the travis failures, it looks like there's probably 4 more files that happen to have the last recorded year = 2015 (`BYU_YEAR`) and not 2010 and so are falling into this `if` statement as well. There could be even more that have a recorded year like 2012 or 2013 but aren't yet BYU'd to 2015. 

I cannot think of an endogenous way to detect what the last recorded year was before extrapolation without adding additional information. I think I have an idea to incorporate it into a new FLAG with some `if` statements _in each zchunk that is BYU'd_ to handle it. Which should be fine but is also an extra place for the teams to make a mistake, even if we present careful notes and examples. Obviously I'd put some thought into minimizing that. 

Alternatively, if we don't think this part of the test will be kept beyond the actual base year updating process, we could also do 
```
olddata %>%
          filter(year <= max(HISTORICAL_YEARS)) ->
          olddata 
```
We will lose a few years of data from the oldnew test but there's less space for user error. 


So, do we want to go flag or slightly less robust oldnew test?

pralitp: There were only 10 or so that failed.  Could we just flag them as `FLAG_NO_TEST` for now?

pkyle: One thing to keep in mind, the FAO Aquastat data are typically written out in (at best) 5-year intervals, with the data falling in years like 2008, 2012, 2017. Often there will be 20-year lapses in a time series. So, at least for that specific data source we don't want to be filtering years prior to filling in the missing data.

bpbond: @pkyle Thanks! But here we're talking about the test code, not filtering before filling data.

@abigailsnyder @pralitp I'm also leery of adding extra logic/steps the user is responsible for. Another possible way to address this: weaken the oldnew test so it filters `newdata` to only the years present in `olddata` before doing the comparison. This would trim off any extrapolated data; the cost is that now it won't detect bad newdata that have extra years for non-BYU reasons.

R/extrapolators.R
```diff
@@ -0,0 +1,83 @@
...
+#' @return A vector with the last n non-NA values from input
+#' vector x.
+#' @importFrom assertthat assert_that is.scalar
+#' @author ACS June 2019
+last_n_nonNA <- function(x, n){
+
+  assert_that(is.scalar(n))
+
+  if(n > length(x[!is.na(x)])){
+    stop('asking for more nonNA years than you have.')
```
bpbond: 😆

russellhz:
@abigailsnyder where does the PR stand? I don't follow everything but it looks like this was primarily for the BYU? @ssmithClimate I think you are just extrapolating from the last available year of data? Would these functions be helpful to merge in?

PR #1118, opened 2019-06-27
Print filename (and stop) if there's a file-read warning

### LINKED ISSUE #1096: Print filename if a warning occurred
Matthew writes:
```
When I ran the driver, I received the following warning message:
 
Unnamed `col_types` should have the same length as `col_names`. Using smaller of the two.
 
Is there any way to easily identify which chunk (or better yet which input file) 
is causing this warning?
```
It would be useful if `load_csv_files()` checked for this (probably via `tryCatch()`) and printed out file information.

bpbond:
(Whoever is available to review. Thanks.)

Closes #1096


bpbond:
FYI @mbins

R/utils.R
```diff
@@ -68,7 +68,16 @@ load_csv_files <- function(filenames, optionals, quiet = FALSE, ...) {
     header <- find_header(fqfn)
     col_types <- extract_header_info(header, label = "Column types:", fqfn, required = TRUE)
 
-    readr::read_csv(fqfn, comment = COMMENT_CHAR, col_types = col_types, ...) %>%
+    # Attempt the file read
+    op <- options(warn = 2)
```
pralitp: What does the `warn = 2` here?  Maybe a comment detailing the reason would be helpful

bpbond: 👍

PR #1137, opened 2019-08-27
Gcam dev 082619

pralitp:
GCAM development since GCAM 5.1.3 including but not limited to:
* GCAM-USA updates
* Regional Ag Markets
* Resource - Reserves


pralitp:
@bpbond, on the chance you have any insights.  Travis tests are failing the "xml_conv_utils" tests with the one hint:
```
Warning message:
system call failed: Cannot allocate memory 
```

Intuition and checking locally indicates we shouldn't be using much memory at this point.  From printing the output from my test `gc()` call locally:
```
used  (Mb) gc trigger   (Mb)  max used   (Mb)
Ncells  1507335  80.6    3886542  207.6   3886542  207.6
Vcells 85036570 648.8  339808110 2592.6 530491852 4047.4
```

I had previously thought that we were using too much memory because the GCAM CI was failing due to the above memory error when calling `driver()`.  But the failure here indicates that probably isn't really the case:  On Travis when we run the driver it is with out writing XMLs (in #1073 because of similar issues we are seeing here) but we still get the memory error even though the `gcamdata` memory should be long gone.

I'm not sure what angle to pursue next, is `gc()` not readily handing back memory to the system?

PR #1150, opened 2020-01-21
Yisheng-L223.electricity_CHINA

sysdavid98:


LZRGRAY:
Hi! I am working on a provincial-level model for China but was unable to use the chunks to generate input files. Which chunk was used to create the L223 files you called in this chunk here? Also wondering in which files I should enter base year electricity data or should I create new ones?

"if(command == driver.DECLARE_INPUTS) {
    return(c("L223.Supplysector_elec_GRIDR",
             "L223.SubsectorShrwtFllt_elec_GRIDR""

PR #1156, opened 2020-02-26
Backport Gcam dev 022620

pralitp:
This includes:
* Water supply chunks
* calibrating historical DDGS
* Land data bug fixes
* near term policy updates
* remove `_` from crop names
* cleaning up importFrom statements
* tidy verse 1.2.0 fixes


pralitp:
Note, we will need to update gcamdata.compdata too so we expect the old/new tests to fail until then.

bpbond:
>Files changed 364

Anything/place I should focus on particularly @pralitp ?

pralitp:
Umm, nothing in particular.  It has all been reviewed at least one time already.  I did some general look over to make sure the merge out from the GCAM repo went smooth (no duplicated commits or the like) as that always seems like some black magic to me.

Maybe also the latest tidy verse fixes might be worth looking at if you haven't already as that was kind of the motivation for doing this merge now.  They are in `module-helpers.R`, `zchunk_L101.nonghg_en_USA_S_T_Y.R`, and `zchunk_LA100.IEA_downscale_ctry.R`

I'll load up the comparison data (from output from he GCAM repo which I'd like to verify doesn't change just because it is now gcamdata standalone repo) and point this PR to that until all looks good.

PR #1158, opened 2020-03-26
Data table functions

kanishkan91:
Adding 2 functions with documentation,

1. **fast_group_by**- A faster alternative to the traditional dplyr alternative. It makes use of data.table. It groups data, performs a function, ungroups. Essentially performs a group_by, mutate and ungroup. It can be used within dplyr pipes. Speed increases exponentially with the increase in the volume of underlying data.

2.  **data_table_bind**- A faster alternative to bind_rows that takes advantage of data.table's data processing capabilities. Returns a tibble after binding all input datasets.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -494,3 +494,68 @@ screen_forbidden <- function(fn) {
   }
   rslt
 }
+
+#' fast_group_by
+#'
+#' A version of group_by that uses data.table instead of dplyr
+#'
+#' This group_by function that uses data.table offers a much higher speed
+#' especially when working with high volume datasets. This function can also be
+#' called within dplyr pipes. data.table will also inherently ensure consistency
+#' between LHS and RHS
+#'
+#'
+#' @param df The tibble on which the group_by is to be performed
+#' @param by A vector with the criteria for the group_by.
+#' @param colname A string with the column name on which the grouping is to be performed
+#' @param func A string with the function to be performed. Default is set to "sum"
+#' @return A tibble with the aggregated data.
+#' @importFrom data.table as.data.table
+#' @importFrom tibble as_tibble
+#' @author kbn 24 Mar 2020
+fast_group_by<- function(df,by,colname="value",func= "sum"){
+
+
+    #Convert relevant column to numeric
+    df[,colname]<- as.numeric(df[,colname])
+
+    #Store as data.table
+    df <- as.data.table(df)
+
+    #Complete operations
+    df<- df[, (colname) := (get(func)(get(colname))), by]
+
+    #Save back to tibble
+    df<- as_tibble(df)
+
+    return(df)
+}
+
+#' data_table_bind
+#'
+#' A binding function that uses data.table. This can be used as a replacement for rbind or bind_rows.
+#'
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -494,3 +494,81 @@ screen_forbidden <- function(fn) {
...
+#' @return A tibble with the aggregated data.
+#' @importFrom data.table as.data.table
+#' @importFrom tibble as_tibble
+#' @importFrom dplyr %>%
+#' @author kbn 24 Mar 2020
+#' @export
+fast_group_by<- function(df,by,colname="value",func= "sum"){
+
+
+    #Convert relevant column to numeric
```
bpbond: For consistency with rest of codebase, please add a space after all these `#`s

R/utils.R
```diff
@@ -494,3 +494,81 @@ screen_forbidden <- function(fn) {
...
+    df[,colname]<- as.numeric(df[,colname])
+
+    #Store as data.table
+    df <- as.data.table(df)
+
+    #Complete operations
+    df<- df[, (colname) := (get(func)(get(colname))), by]
+
+    #Save back to tibble
+    df<- as_tibble(df)
```
bpbond: I would just make line 537 the last one of the function: `as_tibble(df)`

R/utils.R
```diff
@@ -494,3 +494,81 @@ screen_forbidden <- function(fn) {
...
+data_table_bind<-function(...){
+
+    #Create a list for binding
+    list_for_bind =list(...)
+
+    #bind into one dataframe using rbindlist
+    df <- rbindlist(list_for_bind,use.names=TRUE)
+
+    #Return as tibble
+    df<-as_tibble(df)
```
bpbond: Ditto

R/utils.R
```diff
@@ -494,3 +494,81 @@ screen_forbidden <- function(fn) {
...
+#' @return A tibble with combined data.
+#' @author kbn 24 Mar 2020
+#' @export
+data_table_bind<-function(...){
+
+    #Create a list for binding
+    list_for_bind =list(...)
+
+    #bind into one dataframe using rbindlist
+    df <- rbindlist(list_for_bind,use.names=TRUE)
```
bpbond:
```suggestion
    df <- rbindlist(list_for_bind, use.names = TRUE)
```

R/utils.R
```diff
@@ -494,3 +494,81 @@ screen_forbidden <- function(fn) {
...
+#' @param by A vector of strings with the criteria for the group_by.
+#' @param colname A string with the column name on which the grouping is to be performed
+#' @param func A string with the function to be performed. Default is set to "sum"
+#' @return A tibble with the aggregated data.
+#' @importFrom data.table as.data.table
+#' @importFrom tibble as_tibble
+#' @importFrom dplyr %>%
+#' @author kbn 24 Mar 2020
+#' @export
+fast_group_by<- function(df,by,colname="value",func= "sum"){
```
bpbond:
```suggestion
fast_group_by <- function(df, by, colname = "value", func = "sum"){
```

PR #1162, opened 2020-08-28
Alternative driver function using R package drake

### LINKED ISSUE #976: Rearchitect the driver to use drake package?
https://github.com/ropensci/drake

This is getting rave reviews in some quarters - "An R-focused pipeline toolkit for reproducibility and high-performance computing"

enlochner:
Proposing the addition of a new function, driver_drake(), that utilizes the R package `drake` to build the data system.

When building the data system, `driver_drake()` skips steps that are already up-to-date, saving time. The central function of drake is `drake::make()`, which builds the data system. `drake's` `make` is more sophisticated than the standard GNU `make` based systems because it checks the content has substantively changed instead of only if a file has been modified, and similarly only runs subsequent steps that are actually affected by the latest changes since the previous `make()`. Using `driver_drake()` significantly speeds up the process for making changes after the initial data system build.

To implement driver_drake, we had to change the way csv files were read in, in `find_csv_file()` in `utils.R` because of the Windows MAX_PATH file path limit when referencing files using the absolute path.  The new method normalizes the file path to the working directory. 

You can use the typical arguments to `driver_drake()` such as `stop_before` or `return_data_names` to return outputs from cache after the initial run. We also have provided a utility method `load_from_cache` for returning outputs in the same format as data returned from `driver(stop_after = "module_emissions_L121.nonco2_awb_R_S_T_Y")`.  We recommend using this utility rather than directly using `drake::readd` directly.

Parallel computing is supported by `drake` and can be implemented in `driver_drake()`. 

We've included more information about `driver_drake()` in the vignette. 

Fixes #976


R/utils.R
```diff
@@ -235,6 +242,43 @@ find_csv_file <- function(filename, optional, quiet = FALSE) {
   }
 }
 
+#' Normalize file path to the working directory
```
bpbond, line 245: Since you've added rmarkdown anyway, see `markdown::relative_to()`–could this be used instead?

R/utils.R
```diff
@@ -494,3 +538,103 @@ screen_forbidden <- function(fn) {
...
+#' Clean up potential target names so they do not include any
+#' characters which drake does not allow.  At the moment this
+#' just means replacing slashes, which we include in FILE or XML
+#' chunk inputs / outputs, with a '.'
+#'
+#' @param chunk_names A character vector of chunk names to clean
+#' @return A cleaned up \code{chunk_names} which can be used as
+#' targets in drake.
+#' @importFrom assertthat assert_that
+sanitize_drake_target_name <- function(chunk_names) {
```
bpbond, line 555: Can you just use `make.names()` instead of defining this?

R/utils.R
```diff
@@ -494,3 +538,103 @@ screen_forbidden <- function(fn) {
...
+load_from_cache <- function(return_data_names, ...) {
+  # gracefully handle an empty request
+  if(length(return_data_names) == 0) {
+    return(c())
+  }
+
+  # convert to drake target names
+  sanitized_names <- sanitize_drake_target_name(return_data_names)
+
+  # we need to filter the ... arguments to only those that rake::readd take
```
bpbond, line 582: "those that **d**rake::readd take**s**"

R/utils.R
```diff
@@ -494,3 +538,103 @@ screen_forbidden <- function(fn) {
...
+  # reset names to the gcamdata names as users would otherwise expect
+  names(ret_data) <- return_data_names
+
+  invisible(ret_data)
+}
+
+#' create_datamap_from_cache
+#'
+#' Re-creates the GCAM data map by pulling the data out of cache and processing
+#' it's metadata.  We need to know the plan to be able to identify which data to
```
bpbond, line 603: "its"

tests/testthat/test_driver_drake.R
```diff
@@ -0,0 +1,129 @@
...
+       integer_columns <- sapply(x, class) == "integer"
+       x[integer_columns] <- lapply(x[integer_columns], as.numeric)
+
+       numeric_columns <- sapply(x, class) == "numeric"
+       x[numeric_columns] <- round(x[numeric_columns], digits)
+
+       # expect_equivalent no longer accepts rows in different order but we
+       # do want to allow this so we will sort all columns before testing
+       # arrange_columns <- select(x, dplyr::everything())
+       # arrange(x, arrange_columns)
```
bpbond, line 57: If not using, delete these comments?

tests/testthat/test_driver_drake.R
```diff
@@ -0,0 +1,129 @@
...
+
+  #run driver_drake
+  driver_drake(stop_after = "module_water_L145.water_demand_municipal", xmldir = xml_dir, cache = test_cache)
+
+  #get build history and save time of last build of downstream target
+  hist1 <- drake_history(cache = test_cache)
+  row <- tail(which(hist1$target == "L145.municipal_water_R_W_Yh_km3"), n=1)
+  time_mw1 <- hist1[[row, 3]]
+
+  #of other target
```
bpbond, line 102: Here and throughout, gcamdata standard is to have a space after `#` ... this may have changed in recent years though

tests/testthat/test_util.R
```diff
@@ -67,3 +67,15 @@ if(require(mockr, quietly = TRUE, warn.conflicts = FALSE)) {
...
 }
+
+test_that("relative file path works", {
+  #testing to see if relative file path from get_relative_to_workdir resolve to same file as the absolute path
+
+  #get paths for file A_regions.csv in emissions folder
+  full_path <- system.file("extdata/emissions/A_regions.csv", package = "gcamdata")
+  relative_path <- get_relative_to_workdir(full_path)
+
+  #test if paths are equal
```
bpbond, line 78: Not needed

vignettes/driverdrake_vignette.Rmd
```diff
@@ -0,0 +1,252 @@
...
+If you get the following error while trying to load `clustermq`, make sure you have the `zeromq` library loaded ( `module load zeromq/4.1.4`).  
+
+```
+library(clustermq)
+Error: package or namespace load failed for 'clustermq' in dyn.load(file, DLLpath = DLLpath, ...):
+ unable to load shared object '/qfs/projects/ops/rh6/R/3.5.1/lib64/R/library/clustermq/libs/clustermq.so':
+  libzmq.so.5: cannot open shared object file: No such file or directory
+
+```
+On PIC, we had good performance with `clustermq`, type `multicore`, and `jobs = 48`.  The initial build took 5 minutes and 40 seconds as opposed to just over 30 minutes for the `driver_drake()` build without parallelism. For reference, the build with `driver()` on PIC took  22 minutes and 41 seconds.
```
bpbond, line 183: Wow, those build times are LONG. Holy crap, I did not realize.

vignettes/driverdrake_vignette.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+# we can also combine this with other gcamdata utilities to
+# load all input or outputs of a chunk as well
+data <- load_from_cache(outputs_of("module_emissions_L121.nonco2_awb_R_S_T_Y"))
+data
+```
+
+
+## The drake "plan"
+To utilize drakes features `driver_drake()` must generate a drake `plan` to supply to`drake::make()`, which builds the data system. The `plan` is a data frame with columns “target” and “command”. Each row is a step in the workflow and the target is the return value of the corresponding command. `drake` understands the dependency relationships between targets and commands in the plan, regardless of the order they are written. The `make()` function runs the targets in the correct order and stores the results in a hidden cache.  In `gcamdata` the targets are either inputs/outputs or chunk names.  The plan get be obtained by calling `driver_drake(return_plan_only = TRUE)` and may be useful or debugging or in use in additional drake features described next.
```
bpbond, line 218: "drake**'**s features**,**"

bpbond: "The plan can be obtained"?

"or when using additional"?

bpbond:
Woooo @enlochner nice work 👏 ! This is really cool. See above for some suggestions and questions.

tests/testthat/test_driver_drake.R
enlochner: Since we want to check equality regardless of order, I changed this test to check if `all.equal()` is true, ignoring row and column order because I was getting this error when using `arrange`  `Error: Argument 1 is of unsupported type data.frame`.  This `round_df()` function is used over in `test_oldnew.R` as well, should we make a similar change there or do you have any other suggestions?

R/utils.R
enlochner: Currently rmarkdown is just a suggestion. This is a good idea - do you think it's worth adding another dependency?

bpbond: For now, `drake` is just a suggestion too, so it seems OK to me. It's easy to imagine that we'll want to use `rmarkdown` to generate other reports as well. Thoughts @pralitp ?

tests/testthat/test_driver_drake.R
bpbond: Hmm, I don't know. I just meant that it's rarely a good idea to leave commented-out code lying around.

R/utils.R
pralitp: I think the difference is that because it is being called from `find_csv_file` adding the `rmarkdown`  may actually break some users set up who just want to use plain `driver()`.  And sometimes updating packages isn't something to to take lightly.  That is my only concern.

bpbond: Fair point. Maybe log this in an issue to come back to later, but leave alone for now.

enlochner:
Thanks for your feedback @bpbond ! I think I've addressed most of the requested changes. Let me know what you think/if there is anything else to be looked at.

bpbond (approved):
Glad it was useful, and again, very nice work.

Issue #1102, opened 2019-05-17
Feature Request - 


ssmithClimate:

bpbond:
We'll get right on that. 😆

ssmithClimate:
(I have a bad habit of accidentally hitting return to early...)

bpbond:
Re (II):

Under the hood, in `find_csv_file()` the code uses [system.file()](https://www.rdocumentation.org/packages/base/versions/3.6.0/topics/system.file) to find a chunk-requested specified file; this can handle requests for folders as well, so no problem there.

We'd need to refactor `load_csv_files()` a bit (specifically lines 66-80) to check (`dir.exists()`) whether the object being requested is a file or folder, and if the latter, get all csv files within it, adding to the object metadata at each step.

We're assuming that all files in a folder will have the same structure and so can be sensibly row-bound?

pralitp:
A quick response:  Yes we have been aware of the need for some new capabilities to better facilitate generating / running sensitives: #1080 

Generally, I don't think we want user sensitivities to be generated inside of gcamdata (unless it is capability that needs to be added to the Core although even then, just creating every permutation is increasingly unfeasible.) because it is harder to track what are just the Core assumptions vs your own.  Instead we would want users to "shim" (#12) their new data into the processing from outside of the package.  And ideally we could generate some way of identifying the alternative scenarios and how it was generated.  Perhaps going as far as having `gcamdata` produce the GCAM configuration file itself.

ssmithClimate:
Requiring that files have the same structure is reasonable. If we have cases where file structure might be slightly different there could just be different folders (for example, we have some emission controls that are specified relative to GDP/capita, and others that are specified by year. Those could be two different drop folders named appropriately - pollutant-em-control-gdp,  pollutant-em-control-year).

ssmithClimate:
Note that in the applications I mentioned above these would be core model inputs.

bpbond:
I responded above to this from a very narrow technical viewpoint, just thinking about our previous email exchange. But the more I read back through this post, the more complicated and not very possible it seems. Aside, possibly, from the narrow use case of a folder of structurally identical files that are periodically updated by dropping in a new file, the capabilities described above would _greatly_ complicate the `gcamdata` package, and run counter to many of its fundamental design principles (e.g. explicitly specified inputs are the norm).

@pralitp 's note about the 'shim' capability seems spot on to me; I also agree that we need better flexibility in a number of areas.

[...]

Issue #1061, opened 2018-10-17
Ensuring reproducibility when users make data changes


bpbond:
Interesting conversation with @rplzzz and @rjplevin who pointed out we really want to give users an easy-to-use, reproducible, trackable mechanism to generate _changes_ in the GCAM input data. If folks just start messing with input files to generate new outputs, things will get ugly fast. (This has a parallel on the GCAM side, when folks make changes to `configuration_ref.xml` and then can't reproduce problems that crop up.)

This might be as simple as a comment at the top of each XML noting any changes made (generated via some approved 'shim' mechanism that allows users to change the system; see #12 ).

It could also adopt a more [pygcam](https://github.com/jgcri/pygcam) approach, with the data system generating a set of sandboxed changed files only.

rplzzz:
The problem wasn't that users were making changes to `configuration_ref.xml`; the problem is that they _weren't_ changing the input configuration file.  Instead they would change the inputs to the (old) data system and rerun it, changing the files _referred to_ in `configuration_ref.xml` from under it.  They would then rerun GCAM with an input that _looked_ like the reference configuration, but which was actually describing a changed configuration.  This made it very hard to be certain of what was actually run.  Sometimes they even forgot to change the scenario name, adding to the confusion.

I don't think that merely adding a comment at the top of each XML file would help in any meaningful way.  GCAM ignores XML comments, and users operating in this mode tend to overwrite the XML files each time they generate a new scenario.  The real takeaway here is that configuring a GCAM run is a difficult and error-prone process, and therefore we need to provide users with powerful tools for making certain that all the steps have been performed correctly.

What I would like to see is for us to treat the input data as untouchable in normal (i.e., not development) use.  In this scheme, if a user wanted to use the data system to generate a scenario, we would provide a mode in which the user supplied:
1. A directory with changed input files.  
2. A new output directory (a "sandbox") for the changed GCAM XMLs
2. A comment (hopefully informative) to be added to any changed input files.
3. A name for the new scenario.

[...]

PR #1170, opened 2021-03-23
Add the ability to shim user code into the data system

### LINKED ISSUE #12: Ability to 'shim' into driver
Users will want to test out new things. Per @pralitp suggestion - we could give them the ability to modify data frames as the driver runs. For example, they can say "whenever X gets created, call this user-supplied function to modify it before anything else happens".

The more general point is we want to make it easy, or at least provide robust tools, to work with the data system.

pralitp:
First cut at adding the ability to shim user code into the data system workflow which gives users a chance to generate alternative scenarios by modifying data from their own scripts outside of the data system.

Closes #12.  And is potentially related to #1061 and #1102


pralitp:
There may be some considerations on how to best do it.  It works as it stands right now but some worts are required to ensure all it all works transparently and error checking still works.

The following is an example script to get an idea on how one might use it:

```R
devtools::load_all("input/gcamdata")
# Users can write a "user modification" chunk that can be "pluged in" to the data system
# where they will be given the oportunity to modify any "objects" used/created in gcamdata
# they choose before it gets used elsewhere in the data system.
usermod_elec_eff <- function(command, ...) {
  # For a "user modifications" chunk you must define a new command: driver.DECLARE_MODIFY
  # which identifies which "objects" you would like to modify.
  # Note there is no case for the command driver.DECLARE_OUTPUTS since user chunks are
  # required to return the objects the user wanted to modify
  if(command == driver.DECLARE_MODIFY) {
    return(c(FILE = "energy/A23.globaltech_eff",
             "L202.CarbonCoef"))
  } else if(command == driver.DECLARE_INPUTS) {
    # In addition to the objects users want to modify we can also ask for any other
    # inputs we need to do our operations but won't be modified
    return(c(FILE = "common/GCAM_region_names"))
    # The driver will call MAKE on this chunk once it has all of the modify data and
    # other inputs.  They will be made available through "all_data" and we get them
    # the same as in any other chunk.
    all_data <- list(...)[[1]]
    A23.globaltech_eff <- get_data(all_data, "energy/A23.globaltech_eff")
    L202.CarbonCoef <- get_data(all_data, "L202.CarbonCoef")

    # Users could also read in additional files that exist outside of the data system
    # They should do that manually instead of through the driver.DECLARE_INPUTS so as to
    # avoid mixing user's custom files with Core files
    #A23.globaltech_eff.mine <- read_csv("/path/to/my/custom/A23.globaltech_eff_with_random_changes.csv")

    # Make some changes..
    A23.globaltech_eff[1,"improvement.rate"] <- 0.025
    L202.CarbonCoef[1, "PrimaryFuelCO2Coef"] <- 10.0

    # Now we will have to use a specialized return function to return our modified
    # versions of these objects.
    # NOTE: when modifying FILE inputs the object name includes the path to the file,
    # i.e. "energy/A23.globaltech_eff" not "A23.globaltech_eff" thus we will have to
    # explicitly "name" it to make sure it is set correctly.  Other objects can be
    # listed out just like for `return_data`
    return_modified("energy/A23.globaltech_eff" = A23.globaltech_eff, L202.CarbonCoef)
  } else {
    stop("Unknown command")
  }
}

# We added the `user_modifications` option to driver and driver_drake (which is where
# this is really useful, as it will know how to only re-run the chunks that are affected)
# and it will take care of seamlessly "splicing" in the user modifications chunk into the workflow
driver_drake(user_modifications = c("usermod_elec_eff"))
# Now you can see that, for instance, not only has the electricity.xml file been updated with
# the proper efficiency but also the other XMLs that are affected which you may not have thought
# about, like all_energy_emissions.xml, ssp2_emissions_factors.xml, or water_elec_liquids_limits.xml
# One caveat here is this does not take care of managing XMLs for you.  So if you are not careful your
# Core XMLs could be directly replaced with the modified version.  One simple solution is to just set the `xmldir` argument to `driver` / `driver_drake` to some alternative path to keep your modified set of XMLs separate from the Core set.
```

R/driver.R
```diff
@@ -66,7 +66,7 @@ check_chunk_outputs <- function(chunk, chunk_data, chunk_inputs, promised_output
     pc <- attr(chunk_data[[obj]], ATTR_PRECURSORS)
     pc_all <- c(pc_all, pc)
     empty_precursors <- empty_precursors & is.null(pc)
-    matches <- pc %in% c(chunk_inputs, promised_outputs)
+    matches <- pc %in% c(gsub(data.USER_MOD_POSTFIX, '', chunk_inputs), promised_outputs)
```
bpbond, line 69: An explanatory comment might be useful here

R/driver.R
```diff
@@ -195,6 +196,60 @@ driver <- function(all_data = empty_data(),
...
 
+  # check if any user chunks are set in which case we need to adjust the
+  # chunklist/chunkinputs/chunkoutputs to shim the user chunk in place
+  if(!is.null(user_modifications)) {
+    # a user modification chunk uses a special command: driver.DECLARE_MODIFY
+    # to indicate which ds "objects" it wants to modify
+    # that chunk will then require those objects as inputs AND produce those
+    # objects as output
+    # these chunks will also be allowed to specify regular driver.DECLARE_INPUTS
+    # which will be passed as input but NOT output back out
```
bpbond, line 207: Nice design 👏

R/driver.R
```diff
@@ -195,6 +196,60 @@ driver <- function(all_data = empty_data(),
...
+
+    # in order to shim the user modification in we adjust all chunk_inputs
+    # for any object to be modified to input the object with the constant
```
bpbond, line 210: A bit unclear

R/driver.R
```diff
@@ -195,6 +196,60 @@ driver <- function(all_data = empty_data(),
...
+
+    # first get a list of all objects that are to be modified
+    lapply(user_modifications, chunk_inputs, driver.DECLARE_MODIFY) %>%
+      bind_rows() %>%
+      # generate the mod data name by appending data.USER_MOD_POSTFIX
+      mutate(data_mod = paste0(input, data.USER_MOD_POSTFIX)) ->
+      modify_table
+
+    # adjust chunkinputs so chunks that require as input any object that is
+    # to be modified will now input the data.USER_MOD_POSTFIX appended name
```
bpbond, line 223: Ah, this is much clearer—got it

R/driver.R
```diff
@@ -195,6 +196,60 @@ driver <- function(all_data = empty_data(),
...
+             from_file = if_else(is.na(data_mod), from_file, FALSE)) %>%
+      select(-data_mod) %>%
+      # add on the input requirements for the user mod chunk which are the
+      # original object names as well as any other driver.DECLARE_INPUTS they
+      # require
+      bind_rows(select(modify_table, -data_mod),
+                lapply(user_modifications, chunk_inputs, driver.DECLARE_INPUTS)) ->
+      chunkinputs
+
+    # add in the outputs from the user mod chunks which are the modify object names
```
bpbond, line 236: "modified"?

R/driver.R
```diff
@@ -195,6 +196,60 @@ driver <- function(all_data = empty_data(),
...
+    # appended with data.USER_MOD_POSTFIX
+    lapply(user_modifications, chunk_outputs, driver.DECLARE_MODIFY) %>%
+      bind_rows() %>%
+      mutate(output = paste0(output, data.USER_MOD_POSTFIX)) %>%
+      bind_rows(chunkoutputs) ->
+      chunkoutputs
+
+    # now we just need to add the user mod chunks to the chunklist
+    bind_rows(chunklist,
+              tibble(name=user_modifications,
```
bpbond, line 246: For consistency, use spaces around `=` in these rows

R/utils.R
```diff
@@ -393,7 +393,7 @@ find_chunks <- function(pattern = "^module_[a-zA-Z\\.]*_.*$", include_disabled =
 #' 'file_file' (whether object is read from a file), and 'optional' (whether
 #' the object is optional or not).
 #' @export
-chunk_inputs <- function(chunks = find_chunks()$name) {
+chunk_inputs <- function(chunks = find_chunks()$name, call_flag = driver.DECLARE_INPUTS) {
```
bpbond, line 396: Was Roxygen documentation added for this new parameter?

bpbond (approved):
Nice design! I like this—pretty elegant for a first cut. A few questions below but all pretty minor; the approach seems sound.

russellhz:
I was playing around with this pull request last week. As it is written now, it works great to create one new set of files. But if you run a for-loop to change a value multiple times and output to separate xml folders, it generates all xmls and takes a very long time. Copying the xmls first didn't seem to help. Perhaps you could combine a for-loop with a `stop_after` call to just change one xml file, but that defeats the benefits of driver_drake. 

In my dream world, I would run something like this:
```R
usermod_elec_eff <- function(command, ...) {
  if(command == driver.DECLARE_MODIFY) {
    return(c(FILE = "energy/A23.globaltech_eff"))
  } else if(command == driver.DECLARE_INPUTS) {
    return(c())
    all_data <- list(...)[[1]]
    A23.globaltech_eff <- get_data(all_data, "energy/A23.globaltech_eff")

    # Change random value to TECH.IMPROVE.RATE
    A23.globaltech_eff[1,"improvement.rate"] <- TECH.IMPROVE.RATE
  } else {
    stop("Unknown command")
  }
}

# Sequence of improvement rates to loop through
improvement_rates <- seq(0.1, 0.9, 0.1)
for (i in 1:length(improvement_rates){
  TECH.IMPROVE.RATE <- improvement_rates[i]
  data.USER_MOD_POSTFIX <- paste0("__", i)
  driver_drake(user_modifications = c("usermod_elec_eff"))
}
```

And it would produce only the xml files that need to be changed, all with a "__X.xml" at the end in one folder. This seems like it would be pretty easy to code in `driver()` (minus the only xmls we need part), since we can trace the objects that we change and then change input/output names as needed. But I have no idea how to do this in `driver_drake()`, which is still a bit of a black box to me. 

@pralitp @crvernon

russellhz:
Note that the failure is from the old-new test due to compdata not being updated yet

PR #1174, opened 2021-06-24
Adding renv package to gcamdata

hahsan1:
This PR adds the renv package to gcamdata with the goal to reduce R package and version issues when building the data system.

One of the new files that is included is renv.lock, which is a list of the R package names and versions. This branch has been successfully tested on various platforms (Windows, Mac, PIC) for running the data system. 

The steps to test this branch are:
1. Clone gcamdata repo and checkout gcamdata-renv
2. Open gcamdata.Rproj (or load R on pic)
3. Install renv package (if not already installed) – to install run: install.packages( "https://cran.r-project.org/src/contrib/Archive/renv/renv_0.9.2.tar.gz", repos = NULL, type = "source" )
4. Run the following in sequence:
- renv::init( bare = TRUE) # initiates the local renv library and may prompt you to restart the R session
- renv::restore() # installs the packages
- devtools::load_all() # loads gcamdata
- driver_drake() # builds the data system

@crvernon and @pralitp : I have been told to keep you in the loop with this branch. Also, I'm not sure how we would document the steps for setting up renv – is there a wiki?


russellhz:
I updated the renv package version list from Hamza, added in github actions, and merged in main since this was pretty far behind. To pass the checks, I also:
- renamed the function prebuilt_data() to extract_prebuilt_data() so that we don't have two markdown files with the same name
- added a line in test_oldnew.R to convert NaN to NA for one file in the comparison data 
- Ellie had previously removed some %notin% functions and addressed some small csv file issues

I was also thinking this should be merged into the development branch instead of main

Once this is merged, I need to put these directions in the wiki (just a note to self)

Issue #1091, opened 2019-02-28
Hydroelectric generation not written to XML in hindcast mode


pkyle:
Right now the methods for processing hydroelectricity generation do not write the output for years that are within the HISTORICAL_YEARS but after the MODEL_BASE_YEARS. The result is that when running in hindcast mode, for example with a final calibration year of 1990, the model time periods between 1990 and 2015 are assigned zero hydroelectricity. This causes a significant jump in the generation by other sources during this time period, which falls back down in 2015.
The cause seems to be the way that the data are processed; generation during model base years is handled in one table (`L223.StubTechFixOut_elec`), and future years are handled in a separate table (`L223.StubTechFixOut_hydro`). The latter table is built from `L118.out_EJ_R_elec_hydro_Yfut`, which explicitly includes only FUTURE_YEARS. As a simple fix, we could set `L118.out_EJ_R_elec_hydro_Yfut` to include both HISTORICAL_YEARS and FUTURE_YEARS, or to include any HISTORICAL_YEARS that are within the MODEL_FUTURE_YEARS. The changes would be made in `module_energy_LA118.hydro` and `module_energy_L223.electricity`.

pkyle:
Oops! That discrepancy was not intentional. Just re: the scales of what's at stake here, globally, in 2015 hydropower output was 14.6 EJ, tidal power was 0.004 EJ, and "other sources" was 0.11 EJ. From the IEA documentation: "Other sources includes production not included elsewhere such as fuel cells." So, that category should probably be excluded from hydro, by revising the mapping table `IEA_product_fuel`. Can just leave the assignment blank, similar to "Heat output from non-specified combustible fuels". 
As for tidal power, it seems like that ought to be included in the power sector, and hydro is the logical technology to assign. South Korea does operate a 254 MW tidal power plant, and there are a few others around the world. So I guess I'd recommend two changes: (1) revise the mapping file to drop "Other sources" (don't map it to anything), and (2) Revise `module_energy_LA118.hydro` to take in the mapping file `IEA_product_fuel`, and join that to the energy balances, filter to `fuel == "elec_hydro"`, then group/sum to combine hydropower with tidal power.

PR #1189, opened 2022-03-11
Bug/hydro historical

russellhz:
Addressing #1091.

I'm first just dealing with the change to years in `L223.StubTechFixOut_hydro` (and an incorrect precursor for `L223.StubTechFixOut_elec`). Assuming this passes the checks, I will add in the fix to make hydropower defined consistently, which will also require a change to the compdata (and therefore fail the oldnew test).


russellhz:
Checks passed at 8652ab4bdfb9b371107fea87a8e4a2148aede0e0, then should fail with new commits

PR #1195, opened 2022-04-21
add constants to drake

russellhz:
add constants to drake to ensure that any changes to constants.R will be captured.

This works by cleaning the entire cache if constants has been changed (and it is not the first time drake is being used).  A message is written out to inform the users


russellhz:
note that failures are just due to failing old-new, which is from previous PRs

Issue #1068, opened 2018-12-01
Using semi_join versus filter


bpbond:
It looks like `semi_join(x, y, by = "z")` is 25-30% faster than using `filter(x, z %in% y$z`.

If this is right, see how often the latter is used in code and change.

See also #1065

russellhz:
This turns out to be true, except for when there are multiple filters applied at once, in which case it is often faster to have one call to `filter()` rather than use `filter() %>% semi_join()`. 

There were somewhere around a hundred places where it was faster to replace with `semi_join()` or `anti_join()`. I have changed these in #1210 . I also added this to the speed tips but am not planning to enforce it via tests because of the exception.

PR #1210, opened 2022-08-09
Replace in with semijoin

russellhz:
Based on #1068

`semi_join(x, y, by = "z")` is faster than using `filter(x, z %in% y$z)`, except for when there are multiple filters applied at once, in which case it is often faster to have one call to `filter()` rather than use `filter() %>% semi_join()`. 

There were somewhere around a hundred places where it was faster to replace with `semi_join()` or `anti_join()`. I also added this to the speed tips but am not planning to enforce it via tests because of the exception.


bpbond:
Huh, that's interesting!

PR #1211, opened 2022-08-12
Speed up filters

russellhz:
Originally based on #1068 

However, for clarity we will generally not be replacing `filter(x, z %in% y$z)` with `semi_join(x, y, by = "z")`.

This PR includes:

1. Replacing `filter(x, !(z %in% y$z))` with `anti_join(x, y, by = "z")`,
2. Significant speed improvement to `zchunk_L252.MACC.R` 
3. Removing unnecessary filtering in `zchunk_L224.heat.R`
4. General clean up, such as removing consecutive `filter()` calls.


pralitp (approved):
This all looks good to me.  Thanks for tackling this!

The only comment/clarification: was it your intent to _forbid_ the use of `filter(! a %in% b)` or merely suggest using `anti_join` as an alternative?  If the later, what is the strategy when people realistically forget the suggestion?  Perhaps creating a utility search for these (and ideally other performance sinks) similar to`screen_forbidden` would be useful even if we don't enforce it via tests.  Of course easier said than done so I'll leave it up to you.

russellhz:
@pralitp I didn't want to forbid it because there are some circumstances where it's not faster, or would be a pain to avoid. For example, this line: `filter(iso %in% FAO_PRODSTAT_DOWNSCALED$iso & !iso %in% LDS_ctry_SHARES$iso)`. 

Or this line `filter(!(region %in% A_regions$region[A_regions$tradbio_region == 0] & resource == "traditional biomass"))`

I have written regex pattern that catches these instances, but it's hard to avoid catching instances like this as well: `mutate(Non.CO2 = if_else(Non.CO2 == "SO2" & !region %in% states_subregions$state, SO2_name, Non.CO2)) %>%`. 

We can check for the word `filter`  in the line, but that would skip any instances where it would be on the second line. Which might be ok because usually in that case, there be multiple conditions to the `filter()` command, and therefore it's probably faster to not use `anti_join()`

All of this to say, that I wrote a new `screen_undesired()` function that we can use to check this without including it in the tests.

Also, generally, the most important slow-downs I've found is unnecessary grouped operations, which I don't think we'll be able to screen for.

PR #1223, opened 2022-10-26
add chunk_details_description

russellhz:
@hassaniazi 
New function to get the details and description for a single chunk. 

To get the full data system info, users can run:

```
chunk_names <- find_chunks('^module_[a-zA-Z\\.]*_L.*$')
lapply(chunk_names$name, chunk_details_description) %>% bind_rows()
```

Still need to address missing details/description for the following chunk:
image


hassaniazi:
I think it's a great idea to have it as single utility function, which gives users the ability to extract introductory information of a chunk. However, `info("data_object")` can also do something similar. Although yes, not for chunks, but for individual chunks one can easily find details and description in the chunk header by simply scrolling up.
The motivation behind this is to give an overarching summary of data system at one place, which can be handy for starters, or during debugging, or even setting scenarios. So, I think it could be equally great to have it run on full data system and return a file `gcamdata_chunks_info` in `data/` folder

